PR #2, opened 2020-03-31
Wflow sbm

verseve:
Add wflow_sbm (initial)
- initializes wflow_sbm vertical model (SBM struct with states and model parameters)
- reads netcdf files from wflow test model for Moselle catchment (staticmaps.nc and lai_clim.nc)


visr:
Great, I think functionally this is good to go.

For style if possible I'd like to follow https://github.com/invenia/BlueStyle and the base Julia style guide.

The base docs say this about the naming conventions: https://docs.julialang.org/en/latest/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/-1

So the function `StateVariables` should probably be lowercase, how about `statenames`, which goes nicely with the base function `fieldnames`?

The `SBM` struct members should ideally also be lowercase, to avoid that people will assume that they are types or modules. Unless they are single symbols perhaps. If there is an established name used in literature we can also use that, for example `θᵣ` (typed in the REPL or Juno as \theta<tab>\_r<tab>) instead of `thetaR`. Or do you want to avoid unicode characters altogether? I think it may be nice if it reads like a paper or textbook.

PR #4, opened 2020-04-16
support TOML configuration files

visr:
This is just a stub, but good to get some feedback on. I decided to wrap 
the Dict returned by the TOML parser into a Config struct. This way, we 
can easily define functions that operate on the configuration. 
Specifically, I already implemented Base.getproperty, to make it easy to 
get values with dot notation:

```
config.input.time instead of config["input"]["time"]
```

Since TOML is used in the Pkg manager, it is included with it. Officially it is an internal API, but will likely be moved to an official standard library in the future. I noted some examples of TOML syntax that is not yet supported in the example file.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,9 +4,11 @@ authors = ["Deltares"]
 version = "0.1.0"
 
 [deps]
+CFTime = "179af706-886a-5703-950a-314cd64e0468"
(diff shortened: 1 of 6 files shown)


verseve:
In the wflow design doc we decided to use the TOML file format for the configuration file(s) of wflow model(s). This implementation looks good, I like the dot notation, and it seems to the TOML syntax is good enough for our purpose.

PR #6, opened 2020-05-19
switch from array of structs to struct of arrays layout

visr:
Specifically, using the TypedTables package.

This also includes a bit of work in progress forcing code, I think you can ignore that for now. Just wanted to get some eyes on the switch first, before continuing.

I added BenchmarkTools and Juno as dependencies, but these should be removed again later on, or added as test only dependencies. But having them here is just for convenience for now.


verseve:
I think the switch to TypedTables is a good idea. It allows for more flexibility compared to array of structs (e.g. optional columns depending on initial settings (for example use of lai climatology or including of snow modelling)). Also the switch does not affect the run time of the wflow_sbm model.

PR #21, opened 2020-06-25
Reservoirs

verseve:
Reservoir (simple) functionality has been added. 
Large differences with wflow Python:
1) pits (5) are not burned into the local drain direction (ldd) for reservoir locations, this is handled in the code
2) reservoirs are now included within the kinematic wave for river flow (and not seperately as in wflow Python).


visr:
Closing in favor of #22 which also includes these commits.

PR #24, opened 2020-07-02
expand NetCDF input and output support

visr:
This adds an `NCWriter` struct that lets users write specified output maps to NetCDF after every timestep. These currently need to be present in the vertical model table. Support for maps such as runoff from the lateral part still needs to be added but can be done in a separate PR, since this is big enough as is.

It also adds support for data with yearly cyclic patterns, and support to specify which parameters you want to update, and from which NetCDF:

```toml
[forcing_parameters]
precipitation = "P"
temperature = "TEMP"
potevap = "PET"

[cyclic_parameters]
lai = "LAI"
```


src/vertical_process.jl
```diff
@@ -36,7 +36,9 @@ function rainfall_interception_gash(
     # Hack for stemflow
 
     pt = 0.1 * canopygapfraction
-    p_sat = max(0.0, (-cmax / e_r) * log(1.0 - (e_r / (1.0 - canopygapfraction - pt))))
+    # avoid a negative log
+    logarg = max(1.0 - (e_r / (1.0 - canopygapfraction - pt)), 1e-10)
+    p_sat = max(0.0, (-cmax / e_r) * log(logarg))
```
visr, line 41: This was needed to avoid a negative log, which only works for complex numbers.

verseve:
Looks good! Just added some small changes related to config TOML settings (for example starttime, precipitation variable).

PR #27, opened 2020-07-13
major movements

visr:
No functional changes. But quite some code movements. So I thought I'd better push this soon to prevent too many conflicts. See the individual commit messages for what they do.


verseve:
Yes, I agree these code movements look good.

PR #37, opened 2020-09-21
remove internal parameter map

### LINKED ISSUE #36: Extract variables from model defined in TOML file
I suggest we revisit the functionality how we extract variables defined in the TOML file from the model to netCDF and CSV. Now, it is for example not possible to write overlandflow (q), since it will pick up the river discharge in the code:

"Warning: output parameter q not unique, currently set to river"

We could perhaps do something with: "river.q" and "land.q", and resolve this in the code. This approach may also work for writing parameters to CSV.

visr:
This updates the config.toml such that the names and locations directly correspond to the structs and their location in the model struct, removing a level of indirection, and making it easier to extend the code without keeping an internal directory (paramap).

This should allow but does not yet include refactoring and simplification of the initialization code.

Fixes #36


verseve:
Looks good! I made some small changes to the states (see commit and comments), please check if you agree with the changes.

PR #60, opened 2020-12-07
support rivers with no upstream nodes

visr:
Otherwise this gets a "cannot reduce empty collection" error as found by @markhegnauer.


verseve (approved):
Yes, these code changes solve the issue, looks good.

PR #72, opened 2020-12-17
run JuliaFormatter on all code

visr:
This uses the default style. Did this: `using JuliaFormatter; format(".")`

If you want to run this from VS Code, you can use this extension (VS Code Julia currently still has a different default formatter): https://github.com/singularitti/vscode-julia-formatter/

Already went through the diff and I think it's better in almost all cases. Only don't like this so much: https://github.com/Deltares/Wflow.jl/compare/format?expand=1#diff-b6a501ee6fbf026b1fd082d57f38964b960813f58679f4b54485649557bc26e9R102-R107
But perhaps we should just split those lines anyway.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -5,10 +5,7 @@ pages = [
     "Home" => "index.md",
     "Quick start" => "quick-start.md",
     "Model structure" => "structure.md",
-    "Models" => [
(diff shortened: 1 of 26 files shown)


verseve (approved):
Formatting looks good, only changed one small thing (!!! note was missing a return in gwf.md)

PR #74, opened 2020-12-18
drop a few dependencies

visr:
BenchmarkTools was only needed for testing, so it becomes a test dependency only.
The second commit removes the CSV and DataFrames dependencies.
This speeds up loading a precompiled Wflow from 4.3 to 3.2 seconds on my machine, julia 1.6.


verseve (approved):
I agree with the removal of these dependencies, and indeed we are (for now) not using many features of CSV and DataFrames.

PR #77, opened 2021-01-04
add progress logging

visr:


verseve:
Would be good to also have the option (or standard?) to write log events to a file. As part of this PR or do we want to include this in another PR (wflow_cli?)?

PR #84, opened 2021-01-08
grab bag of some late improvements

visr:
Including renaming the main function from `run_simulation` to `run`.

I still have a few more things I want to add to the docs and test, but this can already be merged. Will probably start registration tomorrow around noon.


verseve (approved):
Some very nice last improvements before the first release!

PR #95, opened 2021-01-15
Fix drains option

verseve:


visr (changes requested):
Would be good to add a regression test for this. I tried quickly to verify that it works by adding a test but it still fails somewhere:

https://github.com/Deltares/Wflow.jl/pull/95/checks?check_run_id=1709389774#step:6:104

visr (approved):

PR #106, opened 2021-02-04
doy should not exceeed 365

verseve:
small fix, doy is used to lookup the column in a HQ table, Q has as maximum of 365.


laurenebouaziz:
this fixes the issue, model runs for 1980-12-31

PR #114, opened 2021-03-19
Check params

verseve:
Review and changes of struct parameters to get a bit more consistency. An example: changing the Manning's n parameter when the model was already initialized did not result in any changes, because the update of the alpha parameter was only part of the initialization and not the update of the model. Changing parameters after initialization remains tricky, and requires detailed knowledge of the model.


visr:
Looks mostly good, great to see the reduction in code.
Looking at `altitude`, is it correct that for `sbm_gwf` altitude is now under `input` in the toml, whereas for other models it is under `input.vertical`? What is the reason for this difference? This change here:

https://github.com/Deltares/Wflow.jl/pull/114/commits/4028bd18357edf0109f09ad16a1481858dc7f97d#diff-5091fa59158833690f71ac46441bdc5de6c5a40b1fd099edebbe8ab21b8054dd

is technically a breaking change to the config file input. If we want that, we should think a bit more about how to approach breaking changes.

PR #119, opened 2021-03-31
Sediment update

hboisgon:
Updated sediment.jl:

- corrected a bug in sediment deposition in the river
- new deposition process for coarse sediment in the reservoirs with a new parameter restrapefficiency
- new variables in the sediment structs so that more variables can be sevaed as outputs

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+- New deposition process for coarse sediment in the reservoirs with a new parameter 
+  ``restrapefficiency`` in the sediment model.
+- New variables added to the LandSediment struct and RiverSediment struct in order to save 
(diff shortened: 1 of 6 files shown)


visr:
Overall, the changes here look good to me.

The extra reservoir trap efficiency parameter is currently not optional, which means that old sediment TOML files will not work with newer releases. If that is what you want, we should tag the new release as 0.3.0 to signify it is a breaking release. Or do you have a good default value that you can set in `ncread` with the `defaults` option?

Could you also add the changes to the changelog file: https://github.com/Deltares/Wflow.jl/blob/master/docs/src/changelog.md? Does the new parameter also need to be added to the docs, here for instance? https://deltares.github.io/Wflow.jl/stable/lateral/sediment/#Lake-and-reservoir-modelling

hboisgon:
Good suggestions!
I added a default value to restrapefficiency (1.0) and updated the documentation and the changelog.
The changelog was not in my branch yet so I had to merge master first.

Should be good to go now

PR #121, opened 2021-04-07
update SBM model in parts (bmi)

visr:
To couple an external groundwater model to the SBM model it is convenient to update the SBM model in parts. The update function of BMI now exposes a function to 1) update the model until recharge, and 2) update the model after update of groundwater model.

### DIFF
--- a/src/flow.jl
+++ b/src/flow.jl
@@ -218,3 +218,11 @@ function update(ssf::LateralSSF, network, frac_toriver)
         )
     end
 end
+
+@get_units @with_kw struct GroundwaterExchange{T}
+    Δt::T | "s"                         # model time step [s]
+    exfiltwater::Vector{T}              # Exfiltration [mm]  (groundwater above surface level, saturated excess conditions)
+    zi::Vector{T}                       # Pseudo-water table depth [mm] (top of the saturated zone)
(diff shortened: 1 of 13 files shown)


src/flow.jl
```diff
@@ -218,3 +218,11 @@ function update(ssf::LateralSSF, network, frac_toriver)
...
     end
 end
+
+@get_units @with_kw struct GroundwaterExchange{T}
+    Δt::T | "s"                         # model time step [s]
+    exfiltwater::Vector{T}              # Exfiltration [mm]  (groundwater above surface level, saturated excess conditions)
+    zi::Vector{T}                       # Pseudo-water table depth [mm] (top of the saturated zone)
+    to_river::Vector{T} | "m3 s-1"      # Part of subsurface flow [mm³ Δt⁻¹] that flows to the river
+    ssf::Vector{T} | "mm3 Δt-1"         # Subsurface flow [mm³ Δt⁻¹]
+end
```
visr, line 228: The units are the same in the comment but not the same in the annotation, m vs mm and s vs delta-t.

verseve: Yes, the unit "m3 s-1" for  ``to_river`` was not correct, this should be the same as we use for the subsurface flow component (internal), so mm and delta-t.

PR #122, opened 2021-04-15
support Float32 structs

visr:
This makes the code and test changes needed to be able to run with Float32 number types in the structs.

Right now we hardcode `const Float = Float64`, but if you change this to `Float32` the tests also pass. To make that work I needed to increase the test tolerance a bit, most often by comparing to a Float32, which causes the approx test to use a larger relative tolerance. This can be merged as is, since it already touches quite some code. But after this we want to be able to set this from the configuration file, as well as updating more code, especially Float64 literals, to prevent functions converting to Float64 in between.


verseve (approved):
Looks good!

PR #126, opened 2021-05-04
Sbm threading

verseve:


visr:
Some timing on the extra latency introduced by the two new dependencies (with many indirect deps) here, compared to master (v0.3.1).

`@time using Wflow`; from 2.3s to 7.6s
`@time Wflow.run("sbm_simple.toml")`; from 17s to 35s

Running the (very short) sbm_simple model with the `wflow_cli` binary, the latency increase is less strong, `Measure-Command {wflow_cli sbm_simple.toml}` goes from 15s to 17s. With somewhat larger model runs this extra latency is quickly offset by faster runtimes though.

PR #128, opened 2021-05-04
use a single internal dimension order with increasing axes

visr:
Previously we accepted x and y dimensions in either order, and adapted to that internally. However, some issues arose with other variants present in and between the netCDF inputs that made it clear that we need to support more variants. Some benchmarking showed that moving away from in place netCDF loading hardly affected performance. So to be both more flexible, and reduce some internal code complexity, we treat the data at entry to always follow the dimension order (x, y, layer, time), and to have increasing axes for each dimension.

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -72,7 +72,7 @@ Base.iterate(config::Config, state) = iterate(Dict(config), state)
 "Extract NetCDF variable name `ncname` from `var` (type `String` or `Config`). If `var` has 
 type `Config`, either `scale` and `offset` are expected (with `ncname`) or a `value` (uniform
 value), these are stored as part of `NamedTuple` `modifier`"
-function ncvar_name_modifier(var)
+function ncvar_name_modifier(var; verbose=true)
     ncname = nothing
     modifier = (scale = 1.0, offset = 0.0, value = nothing)
     if isa(var, Config)
@@ -83,7 +83,9 @@ function ncvar_name_modifier(var)
             scale = param(var, "scale", 1.0)
             offset = param(var, "offset", 0.0)
             modifier = (scale = scale, offset = offset, value = nothing)
-            @info "NetCDF parameter $ncname is modified with scale $scale and offset $offset"
+            if verbose
+                @info "NetCDF parameter $ncname is modified with scale $scale and offset $offset"
+            end
         elseif haskey(var, "value")
             modifier = (scale = 1.0, offset = 0.0, value = param(var, "value"))
         end
@@ -94,9 +96,9 @@ function ncvar_name_modifier(var)
 end
 
 "Extract a NetCDF variable at a given time"
-function get_at!(
-    buffer,
-    var::NCDatasets.CFVariable,
+function get_at(
+    ds::NCDataset,
+    varname::AbstractString,
     times::AbstractVector{<:TimeType},
     t::TimeType,
 )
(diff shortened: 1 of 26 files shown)


src/sbm_gwf_model.jl
```diff
@@ -108,6 +97,18 @@ function initialize_sbm_gwf_model(config::Config)
     βₗ = ncread(nc, param(config, "input.lateral.land.slope"); sel = inds, type = Float)
     clamp!(βₗ, 0.00001, Inf)
     ldd_2d = ncread(nc, param(config, "input.ldd"); allow_missing = true)
+
+    # the ldd provided has the Y direction flipped, fix this first
+    ldd_fix_map = Dict{Int8, Int8}(
```
verseve, line 102: This part seems a model case specific solution for the test model of sbm_gwf_model? I guess the ldd values are not correct in our test model, my suggestion is to solve this in the test model, and not in the code.

visr: Ah yes, good catch. Somehow I was under the impression that I was adding this to the test code. This shouldn't be here.

src/io.jl
```diff
@@ -191,22 +159,33 @@ function update_forcing!(model)
...
+                    data[sel_lake] .= 0
                     param_lake[par][i] = avg
                 end
             end
         end
 
         param_vector = param(model, par)
-        param_vector .= buffer[sel]
+        data_sel = data[sel]
+        if par == (:lateral, :land, :h_riv) || par == (:lateral, :land, :q_riv)
```
verseve, line 178: I don't think we need this check if `h_riv` and `q_riv` are not used as part of the `OverlandFlowSediment` struct. For this to work we probably also need to modify the sediment_config.toml file of the sediment test model.

visr: They are not used, but it will error if we attempt to add those missing cells to the `param_vector::Vector{Float64}` in the line below. I was planning to remove them from the `OverlandFlowSediment` and load the forcing into the `RiverSediment` struct directly. When that is ready, we can remove this check.

You're right that if we do this, we probably need to move

```toml
[input.lateral.land]
h_riv = "h"
q_riv = "q"
```

to `[input.lateral.river]`. That would be a breaking change, so perhaps best to still include that in this release?

verseve: My preference is to resolve this now, we're going anyway from version 0.2.0 to 0.3.0. I think we also need to change tehn the `forcing` part in the toml file to:

```toml
forcing = [
  "vertical.h_land",
  "vertical.interception",
  "vertical.precipitation",
  "vertical.q_land",
  "lateral.river.h_riv",
  "lateral.river.q_riv",
]
```

visr: Ok, let's do that, still in this PR, or a follow up? I realised yesterday that we could probably get rid of `forcing` in the TOML altogether, and determine it ourselves, since we fail if a parameters has multiple timesteps but is used statically, and a forcing parameter needs to have that time dimension.

verseve: My preference is to resolve it in this PR. For the `forcing` part although not required from a code perspective, from a user perspective it may be better that this is explicit (similar to the `cyclic` part). What do you think?

visr: Indeed it is nice to be able to look at a TOML file and see which data is forcing data, and which is cyclic. We can make this a separate issue, there is no rush for this.

I can work on h_riv / q_riv early next week. If you have time before, feel free :)

verseve: Resolved this in commit https://github.com/Deltares/Wflow.jl/pull/128/commits/343b68fed1fa20927e91827a90266e2ab459c1a2, do you agree with these changes?

PR #129, opened 2021-05-05
Kinwave vars

verseve:


hboisgon (approved):
Definition of the inwater and volume variables are correct and the resulting water balance error for surface water (land+river) are small enough.

PR #140, opened 2021-06-09
fix and test restart

visr:
Still needs to add tests to compare that runs with a restart are equal to ones without, reading the output of the runs from test/run.jl, and a changelog entry.


verseve (approved):
Looks good, only changed two comments about used state.

Issue #135, opened 2021-06-03
specific comments on documentation


laurenebouaziz:
For the first wflow stand-up, I went through the documentation and noted a few things that could be improved (if there is time!), they are listed below:

In Config and TOML
In model section, it was not clear to me what this is doing:
min_streamorder = 3                 # minimum stream order to delineate subbasins, default is 4
but I am reading later that this is for the multi-threading – maybe good to add: # minimum stream order to delineate subbasins, default is 4 (for parallel computing purposes)

in https://deltares.github.io/Wflow.jl/dev/model/sbm/ :
what are T, N, M, R, L ?
vertical => struct SBM{T,N,M}
lateral.subsurface => struct LateralSSF{T}
lateral.land => struct SurfaceFlow{T,R,L}
lateral.river => struct SurfaceFlow{T,R,L}
lateral.river.lake => struct NaturalLake{T} # optional
lateral.river.reservoir => struct SimpleReservoir{T} # optional

In the description of sbm processes (https://deltares.github.io/Wflow.jl/dev/vertical/sbm/): 
-	whenever the variables used in the model are described, it would be good to also add the unit, e.g.  
This amount of water (excesswater and exfiltwater) becomes overland flow (saturation excess overland flow).
And for the subsequent parts aswell.

-	In transpiration and evaporation section is 
sbm.soilwatercapacity[i]
the maximum amount of water in the upper soil layer (0-100mm)?

-	typo in rootindepth ?

-	The transpiration section is sometimes difficult to understand, perhaps some graphs (Feddes reduction, wetroots) could help here as well? 
o	how is availcap calculated?
o	What happens when whole_ust_available is set to False?
o	Can we add the equations for Wflow.acttransp_unsat_sbm ? Now you need to look into the code to see how transpiration is calculated from the unsat zone.
o	Also the equations for capillary rise are not all stated.

-	In HBV:
o	HQ=KUZ(1+α)HQ=KHQUZHQ
difficult to see what is a variable and what is a multiplication here --- KHQUZHQ

-	In groundwater flow modeling:
o	List of required and possible outputs with units?
o	More description of the concept

Issue #132, opened 2021-05-17
add an overview of model variables to documentation


ctenvelden:
It would be good to include an overview of model variables (names + explanation) in the documentation.

visr:
> The documentation provides a mapping from model struct to the different component

This part is probably mostly interesting for developers. Users don't necessarily need to know how we represent things in structs.

For users it would be nice to see if we can generate such a table:

parameter | description | unit | default
-- | -- | -- | --
cfmax | Degree-day factor   for snow | mm ᵒC-1 Δt-1 | 3.75653

Ideally without duplicating this in code and docs, but generating this from the code or a TOML for instance.

Issue #131, opened 2021-05-17
link to HydroMT for model building


visr:
https://github.com/Deltares/hydromt
https://github.com/Deltares/hydromt_wflow

Use example based on Northern Italy basin, which has public data

PR #153, opened 2021-07-20
Docs comments

verseve:
This PR is linked to the following issue https://github.com/Deltares/Wflow.jl/issues/135#issue-910437742.

Edit by @visr: the new docs can be previewed here: https://deltares.github.io/Wflow.jl/previews/PR153/

Edit by @verseve: also linked the following issue https://github.com/Deltares/Wflow.jl/issues/132 and https://github.com/Deltares/Wflow.jl/issues/131.


laurenebouaziz (commented):
Willem, thanks for all these additions! 
Just a few minor things:

in infiltration, is the unit correct, if all of these variables are indeed fluxes, then we should adapt the unit?:
avail_forinfilt [mm] --> [mm t^{-1}]
infiltsoilpath [mm] --> [mm t^{-1}]
infiltexcess [mm] --> [mm t^{-1}]
excesswater  [mm] --> [mm t^{-1}]


in: capillary rise explanation (same here, I would assume these variables are fluxes and depend on the model time step?):
The actual capillary rise actcapflux [mm]  --> [mm t^{-1}]
maxcapflux [mm] --> [mm t^{-1}]

in leakage:
maxleakage [mm/day] --> [mm t^{-1}]


in sbm.jl:
Actual infiltration (compacted and the non-compacted areas) [mm] --> [mm Δt⁻¹] 
also here it is a flux which depend on the model time step?


what is the variable i here:
sbm.soilwatercapacity[i] ?


In Structures -> model, the explanation "These structs are defined as a parametric composite type..." is still technical. 
Perhaps the reference below can be added as well? 
https://syl1.gitbook.io/julia-language-a-concise-tutorial/language-core/custom-types

Thanks!!

verseve:
Hi @laurenebouaziz,
Thanks for the feedback, yes about the units we did follow how this was already specified in the Python code (a long time ago). Generally, input parameters that represent fluxes/rates have 'correct' units, not always explicitly defined in the code, but here is an example from wflow_hbv (https://github.com/openstreams/wflow/blob/master/wflow/wflow_hbv.py#L308). These are then converted to the model time step, and if we follow the same example the unit is then specified as [mm], see also https://github.com/openstreams/wflow/blob/master/wflow/wflow_hbv.py#L1218. So, fluxes are not specified explicitly, except for the ones that differ from the model time step, e.g. [m3/s]. So, there are basically two options:

1. Follow the wflow Python code, and thus for all units that are part of a struct we do not specify time if it is the same as the model time step. This also needs some further correction, because it is now mixed (unit specification in Wflow.jl code).
2. Be explicit about the unit, also for the case when time unit is equal to model time step. This means we also have to change the units for precipitation, evapotranspiration etc. I guess this is also the reason why Jaap went for the first option.

What are your thoughts on this @visr and @laurenebouaziz?

laurenebouaziz:
In the documentation, I think it would make sense to indicate all fluxes with a unit of [L/T] and indicate states with a unit of [L].
In wflow.jl, for many fluxes it is correctly indicated, see for exemple: https://github.com/Deltares/Wflow.jl/blob/master/src/sbm.jl lines 30 to 35.

verseve:
Ok, this is fine with me, it is more explicit. I think the units should be the same in the Documentation and code., so both need to be changed then.

visr:
Yeah I like the explicit version better as well, leaves less room for confusion.

verseve:
@hboisgon : we have changed the units of fluxes in the code and documentation. Could you also please make these changes for the Sediment part? Thanks!

hboisgon (approved):
The new pages in the docs look really nice especially the tables explaining the different struct variables, including units as well as a link to the templates TOML for each model.
The links to hydromt and the examples for building/updating a model are also good.

docs/src/vertical/sbm.md
```diff
@@ -46,37 +45,38 @@ soilinfreduction = true
 The remaining storage capacity of the unsaturated store is determined.  The infiltrating
 water is split in two parts, the part that falls on compacted areas and the part that falls
 on non-compacted areas. The maximum amount of water that can infiltrate in these areas is
-calculated by taking the minimum of the maximum infiltration rate (`infiltcapsoil` for
-non-compacted areas and `infiltcappath` for compacted areas) and the water on these areas.
-The water that can actual infiltrate is calculated by taking the minimum of the total
+calculated by taking the minimum of the maximum infiltration rate (`infiltcapsoil` [mm
+t``^{-1}``] for non-compacted areas and `infiltcappath` [mm t``^{-1}``] for compacted areas)
+and the amount of water available for infiltration `avail_forinfilt` [mm]. The water that
+can actual infiltrate `infiltsoilpath` [mm] is calculated by taking the minimum of the total
```
laurenebouaziz: should it be ?
[mm t``^{-1}``]

docs/src/vertical/sbm.md
```diff
@@ -46,37 +45,38 @@ soilinfreduction = true
...
 maximum infiltration rate (compacted and non-compacted areas) and the remaining storage
 capacity.
 
 Infiltration excess occurs when the infiltration capacity is smaller then the throughfall
-and stemflow rate. This amount of water (`infiltexcess`) becomes overland flow (infiltration
-excess overland flow). Saturation excess occurs when the (upper) soil becomes saturated and
-water cannot infiltrate anymore. This amount of water (`excesswater` and `exfiltwater`)
+and stemflow rate. This amount of water (`infiltexcess` [mm]) becomes overland flow
```
laurenebouaziz: also in this paragraph [mm t``^{-1}``] ?

docs/src/vertical/sbm.md
```diff
@@ -210,22 +307,68 @@ zone layer(s).
 Wflow.acttransp_unsat_sbm(rootingdepth, ustorelayerdepth, sumlayer, restpotevap, sum_actevapustore, c, usl, θₛ, θᵣ, hb, ust::Bool = false)
 ```
 
-Capillary rise is determined using the following approach: first ``K_{sat}`` is determined
-at the water table ``z_{i}``; next a potential capillary rise is determined from the minimum
-of the ``K_{sat}``, the actual transpiration taken from the ``U`` store, the available water
-in the ``S`` store and the deficit of the ``U`` store. Finally the potential rise is scaled
-using the distance between the roots and the water table using:
+The actual capillary rise `actcapflux` [mm] is determined using the following approach:
```
laurenebouaziz: actcapflux [mm t``^{-1}``] ?

docs/src/vertical/sbm.md
```diff
@@ -111,31 +111,61 @@ original Topog\_SBM vertical transfer formulation, by specifying in the TOML fil
...
 
 ## Transpiration and soil evaporation
 The potential evaporation left over after interception and open water evaporation (rivers
 and water bodies) is split in potential soil evaporation and potential transpiration based
 on the canopy gap fraction (assumed to be identical to the amount of bare soil).
 
-For the case of one single soil layer, soil evaporation is scaled according to:
+For the case of one single soil layer (the SBM soil column is not split-up into different
+layers), soil evaporation [mm] is scaled according to, as shown by the following code block
+(`i` refers to the Vector position):
```
laurenebouaziz, line 164: what do you mean exactly by the vector position?

verseve: it is the index of the Vector.

docs/src/vertical/sbm.md
```diff
@@ -210,22 +313,71 @@ zone layer(s).
...
+```
+
+In case of multiple unsaturated layers (`n_usl` ``>`` 1), the calculation of the actual
+capillary rise starts at the lowest unsaturated layer while keeping track of the remaining
+capillary rise `netcapflux` [mm t``^{-1}``].
 
 ## Leakage
-If the `maxleakage` parameter is set > 0, water is lost from the saturated zone and runs out
-of the model.
+If the `maxleakage` [mm/day] input model parameter is set > 0, water is lost from the
```
laurenebouaziz, line 379: [mm t``^{-1}``] instead of mm/day for leakage?

verseve: Input model parameter that represents a flux has always time unit of day (base timestep of Wflow).

laurenebouaziz (approved):
looks really good Willem, very nice additions!
Just two very minor comments added.

docs/src/vertical/sbm.md
laurenebouaziz: ok, you mean the index in space over the grid?

docs/src/vertical/sbm.md
laurenebouaziz: ah good point thanks!

docs/src/vertical/sbm.md
verseve: yes, internally we work with 1D arrays (based on active cells), so it is the index of the 1D array. feel free to change it, if it is not clear!

visr (approved):
Great to have these descriptions of the model parameters, and the other improvements to the docs.

PR #155, opened 2021-07-20
fix drains

verseve:
Link the drain flux from GroundwaterFlow to the overland flow component (sbm_gwf_model.jl). This flux was only linked to the river kinematic wave. After some discussion with @laurenebouaziz and @Huite we decided to link the drain flux only to the overland flow component. There is also a check if drains are located where overland does not occur (because of the river width. the overland flow width may be set to zero), and if this is the case these drain locations are removed.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   relation to precipitation and evapotranspiration fluxes. This was set to the fixed Wflow
   `basetimestep` of 86400 s, and should be set to the actual model time step from the TOML
   configuration file.
+- Add `flux` from `Drainage` (`GroundwaterFlow`) in the `sbm_gwf_model` to the overland flow
+  component instead of the river component of the kinematic wave.
 
 ## v0.3.1 - 2021-05-19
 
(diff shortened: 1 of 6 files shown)


verseve:
@visr : changed the PR to draft to also include a fix for constant_head, issue that @laurenebouaziz did find.

verseve:
@laurenebouaziz : would be great if you could review the constant_head fix!

laurenebouaziz:
thanks Willem, it does not get stuck on this anymore, but there is a next error...

ERROR: LoadError: type NamedTuple has no field ssf
in
    writer = prepare_writer(
        config,
        reader,
        modelmap,
        indices_reverse,
        x_nc,
        y_nc,
        nc,
        maxlayers = sbm.maxlayers,
    )

I also pushed the change of the initialize lake in this branch, so I can remove the other branch I created

verseve:
I think the ERROR is caused by settings in your TOML file. Probably best to take the sbm_gwf_config.toml as an example from the test model.

laurenebouaziz:
thanks Willem, I didn't know it was there :) !

visr (approved):
These changes look good to me.

laurenebouaziz (approved):
Just tested the drain module for the Rhine, it works and reduces the peaks considerably.

PR #156, opened 2021-07-23
fix outflow calc when negative value in log

laurenebouaziz:
@verseve could you check if these changes seem ok?
I will also let you know after the weekend how the test run goes.


verseve (approved):
Yes, the proposed change will fix the issue, thanks!

verseve:
@laurenebouaziz : did these changes solve your issue?

PR #157, opened 2021-08-09
support multi file forcing as well

visr:
Then, if you have a large multi file forcing netCDF that you don't want to combine first, you can change

```toml
[input]
path_forcing = "data/large-forcing-file.nc"
```

to something like

```toml
[input]
path_forcing = "data/monthly-forcing-file-*.nc"
```

For if you have files like
```
data/monthly-forcing-file-2021-01.nc
data/monthly-forcing-file-2021-02.nc
```

This adds a small dependency, https://github.com/vtjnash/Glob.jl/, to do the file finding.

## TODO
- [ ] wait for and require https://github.com/Alexander-Barth/NCDatasets.jl/pull/137
- [ ] add changelog entry
- [ ] add documentation
- [ ] support `*` in the path as well as the filename


hboisgon:
Nice upcoming dev @visr ! I was wondering, would it support a time split only or also a split with variables (eg precip and temp in different nc files)?

visr:
I'm not sure actually, in [the docs](https://alexander-barth.github.io/NCDatasets.jl/latest/dataset/#NCDatasets.NCDataset) it states:

> Variables are aggregated over the first unlimited dimension or over the dimension aggdim if specified. The append mode is only implemented when deferopen is false.

>All variables containing the dimension aggdim are aggregated. The variable who do not contain the dimension aggdim are assumed constant.

I don't think so, we'd have to test it. Would be nice if it just worked. Does xarray do this by default? And do you have test data perhaps?

PR #159, opened 2021-08-19
fix multithreading for multiple basins

visr:
This still needs a test. Currently the test models have single basins.


visr:
Thanks, this test catches the issue properly.

PR #163, opened 2021-09-14
Option to fix alpha value in the kinematic wave

hboisgon:
When alpha is updated in the kinematic wave with the new h value (wetted perimeter), this creates water balance errors.
To avoid these errors and have a more closed water balance the alpha value is now fixed. A bankfull height map can be provided for the river part to calculate the fixed alpha value (based on half bankfull height). For the land part the wetted perimeter is based on the flow width only.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- Option `update_alpha`. When true (default) the ``\alpha`` parameter of the kinematic wave
+  is updated because of changes in water height (this can result in large water balance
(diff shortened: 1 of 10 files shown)


DirkEilander:
@hboisgon Was just checking this in relation hydroMT + pyflwdir developments. I noticed you are reading a h_bankfull_land map from the staticmaps nc file. For rivers we can create this map, but for land a constant value or even zero is the best we can do I think.

verseve:
I agree @DirkEilander , will include this in the review.

verseve (approved):
Thanks for adding this @hboisgon ! I added the fix also to the other models using the kinematic wave. Also, the kinematic wave iterations and update_alpha options were added to the struct SurfaceFlow (no need to provide these to the update function then), and I did update the docs.

@visr : do you agree with adding these options to the SurfaceFlow struct?

PR #165, opened 2021-10-04
River local inertial

verseve:
Option to use the local inertial model for river flow as part of the sbm_model.jl.


docs/src/quick-start.md
```diff
@@ -178,13 +178,15 @@ inspiration.
...
+can run on multiple threads. The optional local inertial model for river flow as part of the
+wflow\_sbm model [SBM + Local inertial river](@ref) model can also run on multiple threads.
+The threading functionality for the kinematic wave may also be useful for models that make
+(partly) use of this routing approach as the [HBV model](@ref) and the wflow\_sbm model 
+[SBM + Groundwater flow](@ref). The multi-threading functionality in Wflow is considered
 experimental, see also the following
 [issue](https://github.com/Deltares/Wflow.jl/issues/139), where an error was not thrown
 running code multi-threaded. Because of this we advise to start with running a Wflow model
-single-treaded (for example during the testing phase of setting up an new Wflow model). 
+single-treaded (for example during the testing phase of setting up an new Wflow model).
```
visr, line 189: missing an h in threaded

src/Wflow.jl
```diff
@@ -80,28 +80,35 @@ include("io.jl")
...
-struct Model{N,L,V,R,W}
+struct Model{N,L,V,R,W,T}
     config::Config  # all configuration options
     network::N  # connectivity information, directed graph
     lateral::L  # lateral model that holds lateral state, moves along network
     vertical::V  # vertical model that holds vertical state, independent of each other
     clock::Clock  # to keep track of simulation time
     reader::R  # provides the model with dynamic input
-    writer::W  # writes model output
+    writer::W  # writes model outputend
```
visr, line 90: remove end

src/Wflow.jl
```diff
@@ -154,7 +161,7 @@ function run(model::Model; close_files = true)
 
     # in the case of sbm_gwf it's currently a bit hard to use dispatch
     model_type = config.model.type::String
-    update_func = model_type == "sbm_gwf" ? update_sbm_gwf : update
+    #update_func = model_type == "sbm_gwf" ? update_sbm_gwf : update
```
visr, line 164: can be removed

src/flow.jl
```diff
@@ -59,6 +187,8 @@ function update(
     # use fixed alpha value based on 0.5 * h_bankfull
     @. sf.α = sf.alpha_term * pow(sf.width + sf.h_bankfull, sf.alpha_pow)
 
+    @. sf.qlat .= sf.inwater ./ sf.dl
```
visr, line 190: If you use `@.` you don't need to use . operators for broadcasting as well.

src/flow.jl
```diff
@@ -283,3 +415,423 @@ end
...
+    for i = 1:_ne
+        zb_max[i] = max(zb[nodes_at_link.src[i]], zb[nodes_at_link.dst[i]])
+        width_at_link[i] = min(width[nodes_at_link.dst[i]], width[nodes_at_link.src[i]])
+        length_at_link[i] = 0.5 * (dl[nodes_at_link.dst[i]] + dl[nodes_at_link.src[i]])
+        mannings_n[i] =
+            (
+                n_river[nodes_at_link.dst[i]] * dl[nodes_at_link.dst[i]] +
+                n_river[nodes_at_link.src[i]] * dl[nodes_at_link.src[i]]
+            ) / (dl[nodes_at_link.dst[i]] + dl[nodes_at_link.src[i]])
+    end
```
visr, line 538: Would be a bit easier to read if at the top of the loop you assign these to a variable:
```
nodes_at_link.src[i]
nodes_at_link.dst[i]
```

src/flow.jl
```diff
@@ -283,3 +415,423 @@ end
...
+
+    # set depth h of reservoir and lake location to bankfull depth
+    h = fill(0.0, n + length(index_pit))
+    if !isnothing(reservoir)
+        inds_reservoir = findall(x -> x > 0, reservoir_index)
+        h[inds_reservoir] = bankfull_depth[inds_reservoir]
+    end
+    if !isnothing(lake)
+        inds_lake = findall(x -> x > 0, lake_index)
```
visr, line 547: Nit, but FYI, this also works: `findall(>(0), lake_index)`, because `>(0)` creates a 1 argument function.

src/flow.jl
```diff
@@ -283,3 +415,423 @@ end
...
+        Δt = stable_timestep(sw)
+        if t + Δt > sw.Δt
+            Δt = sw.Δt - t
+        end
+        shallowwater_river_update(sw, network, Δt, inflow_wb, doy, update_h)
+        t = t + Δt
+    end
+    sw.q_av ./= sw.Δt
+    sw.h_av ./= sw.Δt
+end
```
visr, line 694: This is not yet done consistently, but a `return sw` would be good here, instead of implicitly returning `sw.h_av`. (Or if you prefer `return nothing`).

src/flow.jl
```diff
@@ -283,3 +415,423 @@ end
...
+
+"""
+    stable_timestep(sw::ShallowWaterRiver)
+
+Compute a stable timestep size for the local inertial approach for river flow, based on
+Bates et al. (2010).
+
+Δt = α * (Δx / sqrt(g max(h))
+"""
+
```
visr, line 705: remove blank line

src/flow.jl
```diff
@@ -283,3 +415,423 @@ end
...
+Set `inwater` of the river component for a `Model` with vertical `SBM` concept.
+`ssf_toriver` is the subsurface flow to the river.
+"""
+function set_river_inwater(model::Model{N,L,V,R,W,T}, ssf_toriver) where {N,L,V<:SBM,R,W,T}
+    @unpack lateral, vertical, network = model
+    inds = network.index_river
+
+    @. lateral.river.inwater = (
+        ssf_toriver .+ lateral.land.to_river[inds] +
+        # net_runoff_river
```
visr, line 728: Can be removed?

src/sbm_model.jl
```diff
@@ -1,3 +1,5 @@
+abstract type sbm end
```
visr, line 1: Should start with a capital. How about `AbstractSBM`? Since SBM exists.

src/sbm_model.jl
```diff
@@ -221,49 +176,52 @@ function initialize_sbm_model(config::Config)
...
+            ldd = ldd_riv,
+            dl = riverlength,
+            width = riverwidth,
+            reservoir_index = resindex,
+            reservoir = reservoirs,
+            lake_index = lakeindex,
+            lake = lakes,
+            Δt = Δt,
+        )
+    end
```
visr, line 211: else error

src/sbm_model.jl
```diff
@@ -309,16 +267,26 @@ function initialize_sbm_model(config::Config)
...
+        )
+    elseif river_routing == "local-inertial"
+        river = (
+            graph = graph_riv,
+            nodes_at_link = nodes_at_link,
+            links_at_node = adjacent_links_at_node(graph, nodes_at_link),
+            indices = inds_riv,
+            reverse_indices = rev_inds_riv,
+        )
+    end
```
visr, line 289: else error

src/utils.jl
```diff
@@ -423,3 +436,38 @@ tosecond(x::Hour) = Float64(Dates.value(Second(x)))
 tosecond(x::Minute) = Float64(Dates.value(Second(x)))
 tosecond(x::T) where {T<:DatePeriod} = Float64(Dates.value(Second(x)))
 tosecond(x::T) where {T<:TimePeriod} = x / convert(T, Second(1))
+
+"""
+adjacent_nodes_at_link(graph)
```
visr, line 441: indent for code formatting, also directly below

visr (changes requested):
Great work. Left some minor comments.

docs/src/changelog.md
```diff
@@ -29,9 +31,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   wave equation.
 - Fixed values for forcing parameters are supported, see also [Fixed forcing values](@ref).
 
+### Added
+- Option to use the local inertial model for river flow as part of the [SBM + Kinematic
+  wave](@ref). See also [SBM + Local inertial river](@ref).
```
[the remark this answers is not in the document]
verseve: Good point, I will leave out the SBM + Kinematic wave part here.

docs/src/lateral/local-inertial.md
```diff
@@ -0,0 +1,83 @@
...
+  Nudurupati, S. S., and Istanbulluoglu, E., 2017, The Landlab v1.0 OverlandFlow component:
+  a Python tool for computing shallow-water flow across watersheds, Geosci. Model Dev., 10,
+  1645–1663, <https://doi.org/10.5194/gmd-10-1645-2017>. 
++ Bates, P. D., M. S. Horritt, and T. J. Fewtrell, 2010, A simple inertial formulation of
+  the shallow water equations for efficient two-dimensional flood inundation modelling, J.
+  Hydrol., 387, 33–45, doi:10.1016/ j.jhydrol.2010.03.027.
++ Coulthard, T. J., Neal, J. C., Bates, P. D., Ramirez, J., de Almeida, G. A. M., and
+  Hancock, G. R., 2013, Integrating the LISFLOOD-FP 2- D hydrodynamic model with the CAESAR
+  model: implications for modelling landscape evolution, Earth Surf. Proc. Land., 38,
+  1897–1906, doi:10.1002/esp.3478.
```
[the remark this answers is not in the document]
verseve: Yes, will add these!

docs/src/lateral/local-inertial.md
```diff
@@ -0,0 +1,83 @@
...
+Simplified reservoirs and lakes models can be included as part of the local inertial model, 
+see also [Reservoirs and Lakes](@ref).
+
+## Inflow
+External water (supply/abstraction) `inflow` [m``^3`` s``^{-1}``]  can be added to the local
+inertial model for river flow, as a cyclic parameter or as part of forcing (see also [Input
+section](@ref)).
+
+## Multi-Threading
+The local inertial model for river flow can be executed in parallel using multiple threads.
```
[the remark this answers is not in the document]
verseve: No, the solution depends on the flow at the previous time step, so this is much easier.

docs/src/lateral/local-inertial.md
```diff
@@ -0,0 +1,83 @@
...
+river_routing = "local-inertial" # default is kinematic-wave
+inertial_flow_alpha = 0.5      # alpha coefficient for model stability (default = 0.7)
+froude_limit = true            # default is true, limit flow to subcritical-critical according to Froude number
+h_thresh = 0.1                 # water depth [m] threshold for calculating flow between cells (default = 1e-03)
+riverlength_bc = 1000          # river length [m] for boundary points (default = 1e05)
+```
+
+The momentum equation is most stable for low slope enviroments, and to keep the simulation
+stable for (partly) steep environments the `froude_limit` option can be set to true. This
+setting limits flow to subcritical-critical conditions based on the Froude number ($\le 1$),
```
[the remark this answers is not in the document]
verseve: Will describe this in more detail, it is indeed not very clear now.

src/sbm_model.jl
verseve: This is actually not used, will remove this.

src/flow.jl
verseve: Will add `return nothing`.

src/flow.jl
verseve: removed blank line.

src/flow.jl
verseve: Ah, ok, did change it to `>(0)`.

src/flow.jl
verseve: I agree, did assign these to variables `src_node` and `dst_node` at the top of the loop.

docs/src/quick-start.md
verseve: changed.

src/Wflow.jl
verseve: removed end.

src/Wflow.jl
verseve: yes, did remove this line.

src/flow.jl
verseve: did remove the . operators.

src/flow.jl
verseve: Not sure what you mean here, why could this part be removed?

src/sbm_model.jl
verseve: Good point, added else error.

src/utils.jl
verseve: applied indent for code formatting.

src/sbm_model.jl
verseve: Good point, added else error.

docs/src/lateral/local-inertial.md
```diff
@@ -0,0 +1,83 @@
...
+similar to Coulthard et al. (2013) in the CAESAR-LISFLOOD model and Adams et al. (2017) in
+the Landlab v1.0 OverlandFlow component.
+
+For the downstream boundary condition (ghost point) a fixed water surface elevation of 0 m
+is used. The river length [m] of the boundary cell can be set through the TOML file with
+`riverlength_bc`, the default value is 100 km.
```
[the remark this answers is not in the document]
verseve: I think the default fixed water surface elevation of 0 m is fine (~sea level boundary). The `riverlength_bc` is a bit more tricky to generalize, left this default value out for now.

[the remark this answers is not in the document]
visr: Ok thanks. Searching for `riverlength_bc` in the code, I see you updated the actual default, but a few comments/docs still need to be updated as well:
```
# level). river length at boundary point is by default 1.0e5 m (riverlength_bc).
riverlength_bc = 1000 # river length [m] for boundary points (default = 1e05)
`riverlength_bc`, the default value is 100 km.
```

A default of 0 for the elevation seems reasonable for many applications, though I don't know if we have a highland model, or one that goes to the Dead Sea (-430m), how much that default value will affect the results. Defaults should generally be safe, otherwise they shouldn't be a default. Can't we copy the riverlength from upstream as well, and extrapolate the elevation?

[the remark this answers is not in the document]
verseve: I think copying the riverlength from upstream as well, and extrapolate the elevation is more risky, since we don't know the (dynamic) river depth at the upstream cell. This can lead easily (especially with smaller grid cells and thus river lengths) to an overestimation of the water slope. I also discussed this with @DirkEilander earlier and we concluded then that this is a simple appropriate solution for now, similar to for example the implementation in CaMa-Flood.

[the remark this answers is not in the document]
verseve: Thanks for catching! I did update the docs.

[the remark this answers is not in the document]
visr: Ok good! Glad to hear you considered the options.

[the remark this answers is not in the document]
DirkEilander: An alternative could be to set the water surface elevation of the ghost node based on the river bed elevation of the outlet cell. In that case we could use a fixed large default value for `riverlengt_bc` (say 10 or 100 km) such that it will not influence the upstream water levels independent of the elevation of the outlet. This bc basically simulates a zero water depth boundary condition at a set distance which is arguable more generic in light of the extreme cases that @visr pointed out.  I believe this is currently also used as one of the boundary condition types you can set in SFINCS.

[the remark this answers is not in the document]
verseve: Thanks for the comments @DirkEilander, yes this is indeed a more generic approach, will update the code based on this.

docs/src/lateral/local-inertial.md
```diff
@@ -45,7 +45,7 @@ river_routing = "local-inertial" # default is kinematic-wave
 inertial_flow_alpha = 0.5      # alpha coefficient for model stability (default = 0.7)
 froude_limit = true            # default is true, limit flow to subcritical-critical according to Froude number
 h_thresh = 0.1                 # water depth [m] threshold for calculating flow between cells (default = 1e-03)
-riverlength_bc = 100000.0      # river length [m] for boundary points
+riverlength_bc = 1000.0        # river length [m] for boundary points (default = 1e04)
```
DirkEilander, line 48: This change looks good. Small detail: In the text you mention 10 km, but here it says 1 km. Also in the comments of src/flw.jl#L511 it says 1 km.

verseve: Thanks for catching the typo in flow.jl comments, fixed this now in flow.jl. For the docs: a `riverlength_bc` of a 1000.0 m is specified in the TOML, while the default is 10 km (1e04 m as comment in line 48).

PR #166, opened 2021-10-04
improve error message for reservoirs and lakes that are pits

visr:


verseve:
Actually this is expected, we do not support lakes and reservoirs without downstream nodes. I made a small change to the code (throw an (easy to read) error when this happens) and to the changelog (instead of a fix, marked it an addition). @visr : could you please check if you agree with the error messages, and feel free to change?

verseve (approved):
See my earlier comment on the changes I did commit.

Issue #141, opened 2021-06-10
inflow kinematic wave river


verseve:
To use external time series as point inflow for the kinematic wave of the river domain. For example when measured timeseries at the outflow location of a reservoir are available.

ctenvelden:
I think, since measured timeseries will often come from external sources, this will often be in csv format. So I would say point inflow is most logical. But maybe other cases will have 2d inflows?

PR #167, opened 2021-10-08
Add simple external inflow to the kinematic wave

hboisgon:
As discussed in #141


hboisgon:
Cyclic inputs for other structs than vertical are also now supported (for example cyclic inflow to the river)

PR #173, opened 2021-11-09
Fix subdaily sbm gwf

verseve:
- Changed the scaling of capillary rise, now based on a common approach found in literature.
- Fixed conversion of boundaries of `GroundwaterFlow` [m3/d] to m3/s in `sbm_gwf` model concept (conversion was based on model time step and should be based on basetimestep (86400 s).


verseve:
Suggestion of @JoostBuitink to include the empirical coefficient n (fixed at 2.0), in the scaling of the potential capillary rise, as model parameter with a default value of 2.0, now included in the commit b0d309c

JoostBuitink (approved):

PR #174, opened 2021-11-16
Fix sbm_gwf exfiltwater

verseve:


docs/src/changelog.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+- Fixed calculation of `exfiltwater` as part of the `sbm_gwf` model type. This was based
+  directly on groundwater head above the surface level, without multiplying by the
+  `specific_yield`, resulting in an overestimation of `exfiltwater`.
+
```
Huite, line 14: This changelog doesn't fully explain it yet, and I'm sure someone will wonder in the future what is meant here exactly. To add a little bit:

```
- Fixed calculation of `exfiltwater` as part of the `sbm_gwf` model type. This
  was based directly on groundwater head above the surface level, without
  multiplying by the `specific_yield`, resulting in an overestimation of
  `exfiltwater`. This is required since the groundwater model estimates the
  head based by dividing the volume by the specific yield or storativity of the
  aquifer. So, should the groundwater table rise above surface level, the head
  above surface level does not represent a water column one to one. (This also
  means the groundwater model (slightly) overestimates heads when the head
  rises above the surface level. However, this water is immediately removed,
  and the head will be set to surface level.)

```

Maybe the changelog is not the right place, but since it took me atleast a few minutes of thinking things over, I think it's good to write it down somewhere...

src/sbm_gwf_model.jl
```diff
@@ -511,10 +511,13 @@ function update_sbm_gwf(model)
...
-    # determine excess head [m] (exfiltwater) in groundwater domain (head > surface) and reset head
+    # determine excess water depth [m] (exfiltwater) in groundwater domain (head > surface)
+    # and reset head
     exfiltwater =
-        lateral.subsurface.flow.aquifer.head .-
-        min.(lateral.subsurface.flow.aquifer.head, lateral.subsurface.flow.aquifer.top)
+        (
+            lateral.subsurface.flow.aquifer.head .-
+            min.(lateral.subsurface.flow.aquifer.head, lateral.subsurface.flow.aquifer.top)
+        ) .* lateral.subsurface.flow.aquifer.specific_yield
```
Huite, line 520: It might be worthwhile to replace `lateral.subsurface.flow.aquifer.specific_yield` by `storativity(lateral.subsurface.flow.aquifer)`, since the `storativity` function will pick the right term also for a confined aquifer. 

For the record, this is the storativity function defined in aquifer.jl:
```
storativity(A::UnconfinedAquifer) = A.specific_yield
storativity(A::ConfinedAquifer) = A.storativity
```

The sbm_gwf model now always uses the unconfined aquifer and so specific yield is currently always the right one, but should somebody e.g. experiment with an sbm_gwf with confined, this will ensure it still works. And there's no real downsides.

docs/src/changelog.md
verseve: Yes, good point, I did update the changelog in commit a49fc6d.

src/sbm_gwf_model.jl
verseve: Yes, I agree, see commit a49fc6d.

PR #177, opened 2021-11-30
abs path forcing

verseve:
fixed use of abs path for the forcing file (was broken)

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -18,8 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Model type `hbv`: the surface width for overland flow was not corrected with the river
   width.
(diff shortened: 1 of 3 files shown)


visr (approved):
Thanks for fixing this. I still added a test, and updated the fix. Also updated the changelog entry that this is only an issue with v0.5.1.

`normpath`, like `joinpath`, can take n amount of parts. I was not aware however that unlike `joinpath`, `normpath` does not have a method that operates on a vector of paths.

verseve:
Hey @visr , thanks for the quick review. Not sure if your implementation works with multiple forcing files (*)? I think I also started with your solution, however I think it did not work with multiple forcing files.

verseve:
With the changes I get this when using multiple forcing files (`forcing-sbm-groundwater-part*.nc`)
```julia
julia> dynamic_paths = glob(glob_path, glob_dir)
String[]
```

PR #182, opened 2021-12-13
Local inertial solution 1-D/2-D

verseve:
Routing for model type `sbm` has been extended with a local inertial solution for 2-D for overland/ floodplain flow. It is linked to the existing local inertial solution for 1-D river flow (subgrid channel).  The 2-D solution makes use of a D4 (four direction) routing scheme, while the 1-D river flow solution uses a D8 (eight direction) routing scheme.  For 2-D overland/ floodplain on a staggered grid the numerical scheme proposed by Almeida et al. (2012) is adopted.


docs/src/lateral/local-inertial.md
```diff
@@ -74,23 +74,75 @@ fixed at 0.0 m.
...
+
+```math
+Q_{i-1/2}^{n+1} = \frac{\left[ \theta Q_{i-1/2}^{n} +\frac{(1-\theta)}{2}(Q_{(i-3/2)}^{n} + \\
+  Q_{(i+1/2)}^{n})\right]- g h_f \frac{\Delta t}{\Delta x} (y^n_i - y^n_{i-1}) \Delta y}{1+g\Delta t \\
+   n^2 |Q_{i-1/2}^{n}|/(h_f^{7/3} \Delta y)}
+```
+
+where subscripts ``i`` and ``n`` refer to space and time indices, respectively. Subscript
+``i-1/2`` is to the link between node ``i`` and ``i-1``, subscript ``i+1/2`` is the link
+between node ``i`` and node ``i+1``, and subscript ``i-3/2`` is the link between node ``i``
```
JoostBuitink, line 91: The difference between subscript ``i-3/2`` and subscript ``i-1/2`` is not clear from this text. In the code it is ``qu`` and ``qd``, perhaps include this here as well?

src/sbm_model.jl
```diff
@@ -364,6 +407,27 @@ function initialize_sbm_model(config::Config)
             end
         end
         lateral.land.volume .= lateral.land.h .* lateral.land.width .* lateral.land.dl
```
JoostBuitink, line 409: This line can be removed right? Is is now nested under the if statement when kinematic wave is selected (line 410&411)

verseve: Yes, this can indeed be removed. I just noticed it when I was rebasing this branch. It is solved in commit https://github.com/Deltares/Wflow.jl/pull/182/commits/57ba9fd9b8c3a90460b7c1394176ef8d462cd019.

docs/src/lateral/local-inertial.md
verseve: Thanks for spotting this. The link `i-3/2` was not described correctly (should be between node `i-1` and `i-2`). Also added an image of the numerical scheme.

src/flow.jl
```diff
@@ -720,24 +722,385 @@ function update(
...
+
+    sw.qx0 .= sw.qx
+    sw.qy0 .= sw.qy
+
+    # update qx
+    @threads for i = 1:sw.n
+        yu = indices.yu[i]
+        yd = indices.yd[i]
+        xu = indices.xu[i]
+        xd = indices.xd[i]
```
visr, line 979: It isn't quite clear to me how we iterate over space within a timestep for overland flow, and what the dependencies are like. Why is at-threads safe here? It would also be good to know if this at-threads helps us at all for the average model, otherwise we may as well remove it.

verseve: There are no dependencies within a timestep, so the numerical solution is very suitable for threading (we do the same for river flow).

visr (approved):
Looks good to me! I added some commits, please check them before merging.

Possibly there are more performance gains still on the table, but we can always improve on that later. Would be nice to have some animations of local inertial 2D spreading along the land, maybe as embedded videos in the documentation? Not for this PR though.

verseve:
Thanks for the review @visr! I agree, there are probably more performance gains, just came across this interesting paper (lisflood-fp):
https://www.sciencedirect.com/science/article/pii/S1364815217307478

Issue #139, opened 2021-06-08
check error logging on multi threading


visr:
There was an issue with (wrong) negative f inputs that understandably caused a DomainError in the `log` function here:
https://github.com/Deltares/Wflow.jl/blob/af04ae4b8c9d2b65624a466a86f3b834b404be37/src/horizontal_process.jl#L85

However when running this code multithreaded, this error was not thrown. We should check why that is. #138 will help here as well.

laurenebouaziz:
In hydromt, we recently fixed the calculation of the f parameter, which should now not produce negative values anymore (constrain M function was previously not applied to f).

visr:
Yes, thanks, Willem showed me that you fixed it. Still created this issue to understand the interaction with error handling and multithreading a bit better.

PR #183, opened 2021-12-14
Log

verseve:
Adding log file `log.txt` as output including extension of log messages in the code, as part of `function run(tomlpath)`.
Log level `Info` contains:
- Wflow version
- Mapping of external parameters to input parameters (initialization)
- General (important) model settings
- Run information
- Output information

For a Delft-FEWS run a single line logging file is produced.


visr (approved):
I went through it in detail and made a bunch of changes still, so please have another look if you agree with all those. I'm happy with this as it is. I also made corresponding changes to `wflow_cli` and tested that it is working as desired with this branch.

One thing to keep in mind is that the log message for the terminal is interpreted as Markdown. If your terminal supports colors this will look quite nice (this is Windows Terminal):
image

If there is just one or two variables it can be fine to interpolate them into the log message (`@info "Number of cells is $n."`), but if there are more relevant variables, we probably should add them as keyword argument (`@debug "Starting timestep." time i`). If variable names are not clear you can also change them for the logging: `@debug "Starting timestep." time timestep=i`.

If there is an error, this also ends up with a stacktrace in the log file, though the stack trace is not included in the one line FEWS logging format (the error is). In the future we could also consider just writing the PI diagnostics XML format for FEWS, since we special case it anyway.

verseve:
Thanks for the detailed review and additional commits @visr. It looks good, only made some small changes to a couple of log messages ea143b2.

visr (changes requested):
I don't think we can merge this yet. @markhegnauer tested a model that gave an InexactError on master, and in this branch instead of throwing an error it just appears to hang. It could be related to threading as well, since if I interrupt the process, I get

```
fatal: error thrown and no exception handler available.
InterruptException()
sigatomic_end at .\c.jl:437 [inlined]
task_done_hook at .\task.jl:492
jl_apply at /cygdrive/c/buildbot/worker/package_win64/build/src\julia.h:1703 [inlined]
jl_finish_task at /cygdrive/c/buildbot/worker/package_win64/build/src\task.c:209
start_task at /cygdrive/c/buildbot/worker/package_win64/build/src\task.c:845
```

verseve:
> I don't think we can merge this yet. @markhegnauer tested a model that gave an InexactError on master, and in this branch instead of throwing an error it just appears to hang. It could be related to threading as well, since if I interrupt the process, I get
> 

This could be related to issue https://github.com/Deltares/Wflow.jl/issues/139?

visr:
Yes I think it's quite likely related. This PR does seem to make it worse, perhaps due to the way errors are handled here. The error handling is tested here:

https://github.com/Deltares/Wflow.jl/blob/ea143b2f58e0c77291cf1de3f08a0c47c0b5c2fc/test/io.jl#L366-L370

But perhaps we should make a similar test that will crash while running the simulation.

verseve:
> I don't think we can merge this yet. @markhegnauer tested a model that gave an InexactError on master, and in this branch instead of throwing an error it just appears to hang. 

A wrong initialization of theta_r (was mapped to theta_s) caused the model with this branch to run very slow, and thus seemed to hang (compared to the master branch). This is fixed in https://github.com/Deltares/Wflow.jl/pull/183/commits/98063a2e2d9e0e08be267c73d97fc22c95dc1f24  and also a test is included in https://github.com/Deltares/Wflow.jl/pull/183/commits/bb54ba5a0f1304e6bb17e3134da6de576df3c664 for error handling during the update of the model (forcing an InexactError).

visr (approved):

visr:
Thanks for that fix and the extra test. Good to know that it was a very slow model and not a hang or suppressed error. Do you have any idea why it seemed to slow down so much more on this branch versus master?

verseve:
> Do you have any idea why it seemed to slow down so much more on this branch versus master?

Yes, the wrong mapping between theta_r (internal) and theta_s (external) was introduced in this branch because of a typo in the code (a typo that is not part of the master branch). Did not investigate further where this change had the most impact on the run time of the model, most probably this would be lateral subsurface flow and overland flow.

visr:
Oh great, I assumed the typo was also on master. That's a relief.

janverkade:
Many thanks, everybody, for implementing the logging feature. Very helpful indeed.
Hope it's okay to share suggestions with you ;)
Log.txt file starts every line with "timestamp =". I think that's a superfluous line. Could I request that that's taken out? Many thanks!
[log.txt](https://github.com/Deltares/Wflow.jl/files/7789636/log.txt)

visr:
Done in 3e655defd90e147275aeee0e1a1531b30a87b6be, thanks for the feedback.

PR #184, opened 2021-12-23
Updated documentation 

JoostBuitink:
Update the structure, and added a more detailed user guide

Edit: preview docs: https://deltares.github.io/Wflow.jl/previews/PR184/


docs/make.jl
```diff
@@ -47,8 +54,9 @@ makedocs(;
         prettyurls = get(ENV, "CI", "false") == "true",
         canonical = "https://deltares.github.io/Wflow.jl",
         assets = String[],
+		collapselevel = 2,
     ),
     pages = pages,
 )
 
-deploydocs(; repo = "github.com/Deltares/Wflow.jl", push_preview = true)
+#deploydocs(; repo = "github.com/Deltares/Wflow.jl", push_preview = true)
```
visr, line 62: Why is this commented out? Would be nice to see a preview. It will only deploy to the official docs when it is on the master branch.

JoostBuitink: An artifact from my (local) testing, should be fixed now!

docs/src/user_guide/step1_requirements.md
```diff
@@ -0,0 +1,35 @@
...
+wflow\_sbm model different model configurations are possible. The following model
+configurations are supported in wflow:
+
+ - wflow\_sbm:
+    - SBM + kinematic wave for subsurface and surface flow
+    - SBM + kinematic wave for subsurface and overland flow + local inertial river
+    - SBM + kinematic wave for subsurface flow + local inertial river (1D) and land (2D)
+    - SBM + groundwater flow + kinematic wave for surface flow
+ - wflow\_hbv: HBV + kinematic wave for surface routing
+ - wflow\_sediment as post processing of wflow\_sbm or wflow\_hbv output
```
verseve, line 31: Not sure if linking wflow_sediment to wflow_hbv is already possible, I think wflow_sbm is always used. @hboisgon: can we mention here wflow_hbv?

docs/src/user_guide/step3_input_data.md
```diff
@@ -0,0 +1,56 @@
...
+
+Meteorological data is provided as a single NetCDF file, with several variables containing
+the forcing data for precipitation, temperature and potential evaporation. The code snippet
+below shows the contents of the example file (downloaded [here](@ref wflow_sbm_data)), and
+opened with `xarray` in Python. As can be seen, each variable consists of a
+three-dimensional dataset (`lon`, `lat`, and `time`), and each timestep consists of a
+two-dimensional map with values at each gridcell. Only values within the basin are required.
+
+```
+<xarray.Dataset>
```
verseve, line 21: I think it would be better to show here output from Julia instead of Python, after all, this is a Julia package using NCDatasets to read and write NetCDF files.

docs/src/user_guide/step3_input_data.md
```diff
@@ -0,0 +1,56 @@
...
+Data variables:
+    P        (time, lat, lon) float32 ...
+    TEMP     (time, lat, lon) float32 ...
+    PET      (time, lat, lon) float32 ...
+```
+
+## Static data
+
+
+### List of essential static data
```
verseve, line 36: I did change this list because it also contained some non-essential static data that is not required for a Wflow run, but can provide additional useful info (for example streamorder). Also the required maps (not optional) depend on the model type. Perhaps we should mention here only static data that is required for one wflow model configuration, for example wflow_sbm + kinematic wave? In that case the wflow_dem is also not required.

docs/src/user_guide/sample_data.md
```diff
@@ -0,0 +1,82 @@
+# [Tutorial models](@id sample_data)
```
verseve, line 1: I think this section needs a bit more work/description. It is now just a list of the different test models without further explanation. Also there is overlap with the section (https://deltares.github.io/Wflow.jl/previews/PR184/user_guide/model-setup/#Building-a-model-from-scratch). Do we want to include a real tutorial on models here, or just provide some sample data from our test bank (with some more description)?

verseve (changes requested):
Looks good @JoostBuitink ! See my review comments and some additional commits.

docs/make.jl
```diff
@@ -2,40 +2,48 @@ using Wflow
...
+    "Model documentation" => [
+        "model_docs/intro.md",
+        "model_docs/model_configurations.md",
+        "Vertical concepts" => [
+            "model_docs/vertical/sbm.md",
+            "model_docs/vertical/hbv.md",
+            "model_docs/vertical/sediment.md",
+            "model_docs/shared_concepts.md",
+        ],
+        "Lateral concepts" => [
```
verseve, line 35: Local inertial is missing here.

docs/src/model_docs/shared_concepts.md
```diff
@@ -1,8 +1,6 @@
-# Vertical processes
+# Shared model components
```
verseve, line 1: I think `processes` is better here, these are not really components of a Model.

verseve (changes requested):

docs/src/user_guide/step1_requirements.md
hboisgon: Never did it but technically possible I think. As long as you can get the forcing (water levels and discharge) and the staticmaps contains the slope/ldd and landuse, the rest is alright.

docs/src/model_docs/shared_concepts.md
JoostBuitink: That's true, changed the header to `Shared processes`. I noticed that we currently describe the rainfall interception processes here as well, while they are in fact only used by the `sbm` concept, right?

docs/src/user_guide/sample_data.md
JoostBuitink: Given that the other steps in the user guide are already a small tutorial, perhaps we can rename this section to (something along the lines of) `Example models`? A user can explore these input files combined with the TOML to understand the requirements for the different model types.

docs/src/user_guide/step3_input_data.md
JoostBuitink: Changed the NetCDF information to Julia's `ncinfo` output

docs/src/intro/publications.md
```diff
@@ -115,7 +138,7 @@ indicators with combination of process-based models and data assimilation techni
 Ecological Indicators 66, 428–439. <https://doi:10.1016/j.ecolind.2016.02.016>.
 
 
-## Theses
+### Theses & Internship reports
```
hboisgon, line 141: Can we also add Laurène thesis there :) ?

docs/src/model_docs/vertical/sediment.md
```diff
@@ -13,9 +16,9 @@ structure for the soil loss and lateral structure for the transport in overland
 
 Overview of the different processes for a land cell in wflow\_sediment.
 
-![sediment_inland](../images/soilloss-scheme.png)
+![sediment_inland](../../images/soilloss-scheme.png)
 
-## Soil Erosion
+### Soil Erosion
```
hboisgon, line 21: The level here is not nice as compared to SBM and HBV when you open the page you only see description and reference but not soil erosion and delivry to the river. Could we change the indent level? And delete Description title then.

docs/src/user_guide/additional_options.md
```diff
@@ -0,0 +1,250 @@
+# Additional wflow options
+
+## [Starting the model with "warm" states](@id reinit)
```
hboisgon, line 3: the state for reservoirs and lakes can cause errors easily if you forget to add/remove when switching on and off reservoirs/lakes modelling. I would move the reservoirs/lakes states in the specific reservoirs and lakes section in that same file

docs/src/user_guide/additional_options.md
```diff
@@ -0,0 +1,250 @@
...
+b = "lake_b"
+e = "lake_e"
+locs = "wflow_lakelocs"
+outflowfunc = "lake_outflowfunc"
+storfunc  = "lake_storfunc"
+threshold  = "lake_threshold"
+waterlevel = "lake_waterlevel"
+```
+
+## [Using multithreading] (@id multi_threading)
```
hboisgon, line 110: Can we here put the real CLI examples? That would avoid to search for them in the different pages links. (-t 4 for 4 threads I think ?)

docs/src/user_guide/install.md
```diff
@@ -0,0 +1,122 @@
...
+- If you want to stay up-to-date with the latest version, explore and modify the model code,
+  and write your own Julia scripts around the wflow package, we recommend installing wflow
+  as a [Julia](https://julialang.org/) package.
+- If you don't need extra features, but just want to run simulations, a complied executable
+  version is available. This consists of a single executable, `wflow_cli`, allowing you to
+  run the model via the command line.
+
+Below, we describe how to install both versions of wflow. 
+
+## Installing as Julia package
```
hboisgon, line 14: We also use to have info on how to install Julia itself and some intro how it works in the old quick guide. As Julia is still new to a lot of people, I think maybe useful to keep that info.

docs/src/model_docs/shared_concepts.md
verseve: Yes, indeed, so better to move/merge the rainfall interception part with the SBM section.

docs/src/user_guide/step4_running.md
```diff
@@ -1,4 +1,26 @@
-# Command Line Interface
+# [Step 4: Running a simulation](@id run_wflow)
+
+## Using Julia
```
hboisgon, line 3: With Julia install you can also use the command line interface, you don’t necessarily need to write a script: julia -e "using Wflow; Wflow.run()" "path/to/config.toml"

hboisgon (commented):
Super nice new organisation of the docs @JoostBuitink ! And very nice new pictures for the SBM concepts :)
Here just a few comments mainly on install and usage.

docs/src/user_guide/sample_data.md
verseve: Yes, sounds good, including a bit more description (see also model-setup.md) and remove this section from model-setup.md.

docs/src/intro/publications.md
JoostBuitink: Good suggestion - it is now added to the list!

docs/src/model_docs/vertical/sediment.md
JoostBuitink: Headers now have a correct level.

docs/src/user_guide/additional_options.md
JoostBuitink: Good spot, I have added a state section to the reservoir and lakes sections

docs/src/user_guide/additional_options.md
JoostBuitink: I have added an example for calling Julia from the command line (the `-t` flag does not work for the `wflow_cli` option)

docs/src/user_guide/install.md
JoostBuitink: I felt like the section on installing an using Julia was a bit redundant to keep in the documentation for a package of Julia, so I now added a link to a quick start guide from julialang.org

docs/src/user_guide/step4_running.md
JoostBuitink: That's correct, overlooked this option. It is now added added as an option as well!

docs/src/model_docs/shared_concepts.md
JoostBuitink: The interception section is now moved to (and merged with) the SBM section.

docs/src/user_guide/install.md
hboisgon: Agreed but from experience with hydromt-wflow (python) I get a lot of questions on installation if we just say the classic "conda install -c conda-forge hydromt-wflow"...
So we in the end added some lines also on first download miniconda and create an empty environment.
Saves you some trouble with brand new julia users :)

Your additions are good but I would add the previous lines with the install julia link as well. The rest can go I agree.
"First download and install the [current stable release of Julia](https://julialang.org/downloads/#current_stable_release). Please see [platform specific instructions](https://julialang.org/downloads/platform/) for further installation instructions and if you have trouble installing Julia."

docs/src/user_guide/sample_data.md
JoostBuitink: Moved the text from model-setup.md to sample_data.md, and added a bit more information.

docs/src/user_guide/step3_input_data.md
JoostBuitink: Fixed, and added a link to the example models

docs/src/user_guide/install.md
JoostBuitink: That is a valid point. The links to the Julia install instructions are now added to the docs.

docs/src/user_guide/step3_input_data.md
verseve: Changed this to `NCDatasets` since we use this package in Wflow.

verseve (approved):

PR #186, opened 2021-12-24
disallow NaN in ncread if allow_missing=false

visr:
If the fill value of the netCDF is NaN, NCDatasets will make it a missing. But if that is not the case and we don't want missings, I cannot think of any case where we would want NaN.

This will work for instance to catch NaN values if they are present in the river width with the current call to ncread:

```julia
ncread(nc, param(config, "input.lateral.river.width"); type = Float, fill = 0)
```

### DIFF
--- a/src/hbv_model.jl
+++ b/src/hbv_model.jl
@@ -399,6 +399,8 @@ function initialize_hbv_model(config::Config)
 
     riverlength = riverlength_2d[inds_riv]
     riverwidth = riverwidth_2d[inds_riv]
+    minimum(riverlength) > 0 || error("river length must be positive on river cells")
+    minimum(riverwidth) > 0 || error("river width must be positive on river cells")
 
     ldd_riv = ldd_2d[inds_riv]
     if do_pits
--- a/src/sbm_gwf_model.jl
+++ b/src/sbm_gwf_model.jl
@@ -122,6 +122,8 @@ function initialize_sbm_gwf_model(config::Config)
     # river flow (kinematic wave)
     riverlength = riverlength_2d[inds_riv]
     riverwidth = riverwidth_2d[inds_riv]
+    minimum(riverlength) > 0 || error("river length must be positive on river cells")
+    minimum(riverwidth) > 0 || error("river width must be positive on river cells")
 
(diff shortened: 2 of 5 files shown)


verseve (changes requested):
I guess this PR is related to the NaN values that were caused by the resolved issue https://github.com/Deltares/Wflow.jl/issues/187? 

Although, on the one hand I think it is good to check for NaN values, on the other hand I think it is very rare that this occurs (or does not occur at all). For example, the maps in staticmaps.nc (from HydroMT) all have a `_FillValue`, so NaN is not explicitly defined in staticmaps.nc. For the river part a check for zero values in:
https://github.com/Deltares/Wflow.jl/blob/89e7307be95aa05939f0c38996d8a666bba5c7e1/src/sbm_model.jl#L38
https://github.com/Deltares/Wflow.jl/blob/89e7307be95aa05939f0c38996d8a666bba5c7e1/src/sbm_model.jl#L41

would make more sense I think. These kind of inconsistencies in the input data will result in NaN values when running the model (e.g. division by zero).

So, my suggestion is to leave this part out of the code, and to include for example checks for zero values in river width and length.

visr:
It was triggered by but is not really related to #187. You're right that for HydroMT staticmaps this is not needed, but I see it as an extra defense against missing/NaN data entering and the calculation and causing issues further in the code. Also netCDF files with a (non-NaN) fill value can have NaN values inside, which would currently just slip past the initialization code.

Your suggestion on the river lengths is added now in c42f78f62b7a536e3768af7900837e11053d4535.

verseve (approved):
Thanks for the addition in https://github.com/Deltares/Wflow.jl/commit/c42f78f62b7a536e3768af7900837e11053d4535!

Yes, let's keep it then as an extra defense mechanism against missing/NaN data entering during initialization. It can indeed happen when input files are edited or generated without HydroMT. 

What about updating the changelog?

visr:
I'd say these kind of changes are not really important/relevant enough for users. Before and after this would throw an error. Perhaps at most we could add a general "Improved error messages" line.

verseve:
Sounds good, I agree, an update of the changelog is not necessary.

Issue #181, opened 2021-12-07
Cyclic parameter: should also work for static parameter


verseve:
Cyclic parameter now only works if the model parameter is not initialized (with `ncread`) by the model. If the model parameter is also initialized as part of the initialize function, the following error occurs:

```julia
ERROR: ArgumentError: NetCDF dimension time has length 12.
Only extra dimensions of length 1 are supported.
```

Issue #191, opened 2022-01-05
fixed timestep river kinematic wave


verseve:
If the model timestep is smaller than the fixed timestep for kinematic wave river (`kw_river_tstep`) an InExactError is thrown. The number of iterations for river flow:

https://github.com/Deltares/Wflow.jl/blob/0190b5be0cb99869dbb177f56624d705c4619fac/src/flow.jl#L166

should be calculated in the same way as for overland flow:
https://github.com/Deltares/Wflow.jl/blob/0190b5be0cb99869dbb177f56624d705c4619fac/src/flow.jl#L82

PR #192, opened 2022-01-10
Fix issues #181 and #191

verseve:
Fixing issues https://github.com/Deltares/Wflow.jl/issues/191 and https://github.com/Deltares/Wflow.jl/issues/181.

Issue https://github.com/Deltares/Wflow.jl/issues/181 is fixed by checking if the NetCDF variable `var` has dimension `time` in the `ncread` function. If this is the case, it is a cyclic parameter that is handled by the `update_cyclic!` function.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -28,6 +28,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed how number of iterations `its` for kinematic wave river flow are calculated during
   initialization when using a fixed sub-timestep (specified in the TOML file). For a model
   timestep smaller than the fixed sub-timestep an InexactError was thrown.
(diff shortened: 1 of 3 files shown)


visr:
The #191 fix looks good.

For the #181 fix, I see you check if time is in the dimnames. The `read_dims` function is made such that size 1 dimensions are allowed, so this would break size 1 time variables. Hence the error in #181 says "Only extra dimensions of length 1 are supported."

How does `return nothing` work for `ncread`? I'd think that this would crash later during initialization. One alternative approach would be to just read in the first or current timestep, which would then be updated by `update_cyclic!`. Perhaps we can make use of `get_at` to do this.

verseve:
Yes, good point about the [#181](https://github.com/Deltares/Wflow.jl/issues/181) fix. The approach is now to include dim `time=1` in `dim_sel` of the `ncread` function.

PR #195, opened 2022-01-23
fix a few issues identified by JET

visr:
I ran `report_package` from JET. Most identified issues could be ignored, but it did find a few actual mistakes, which is very helpful.

https://aviatesk.github.io/JET.jl/dev/jetanalysis/#JET.report_package

Would be nice to do the optimization analysis in another PR: https://aviatesk.github.io/JET.jl/dev/optanalysis/.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   width.
 - Fixed use of absolute path for `path_forcing` in TOML file, which gave an error in Wflow
   v0.5.1.
+- Fixed a crash when using glaciers.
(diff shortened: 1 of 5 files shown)


verseve (approved):
See commit https://github.com/Deltares/Wflow.jl/pull/195/commits/f886ff7aa5ad7914d1bfba05c9e570b6fc6b8bc6.

visr:
Thanks. The delta t change was indeed wrong, that was my interpretation, not JET's. I assumed the `basetimestep` was a float, but that is also of type Second. Dividing the two in `glacier_hbv` then gives a float, so that will be fine.

```
julia> using Dates
julia> Second(3)/Second(4)
0.75
```

PR #198, opened 2022-01-26
rename h_bankfull to bankfull_depth

### LINKED ISSUE #197: consistent naming of "bankfull_depth" parameter
The same parameter is named "h_bankfull" for kin wave routing. I suggest to change "h_bankfull" to "bankfull_depth".

DirkEilander:
fixes #197


DirkEilander:
I did not get the tests to run on my machine.. Could the failing tests have to do with a wflow config toml file that need to be updated somewhere?

verseve:
It now uses the actual `riverdepth` map for the kinematic wave, before this change the default value was used. This is causing some small differences.

verseve:
Could you also please update the changelog (this is breaking)? And request a review?

verseve (approved):
Thanks for solving this issue!

PR #199, opened 2022-02-02
don't break if h_bankfull is still used

visr:
By accepting it as an alias, and giving a warning about the name change.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -12,8 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   width `dw` as model type `sbm`.
 - Renamed `h_bankfull` parameter to `bankfull_depth` for consistency between kinematic-wave
   and local-inertial river routing. When using the old name under the
-  `[input.lateral.river]` section of the TOML file, it will still work but suggest to update
-  the name.
+  `[input.lateral.river]` section of the TOML file, it will work but it is suggested to
+  update the name.
 
 ### Added
 - Additional log messages and log file as output, see also [Logging](@ref).
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -10,9 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Model types `sbm_gwf` and `hbv` use the same approach for the calculation of the drain
   width `dw` as model type `sbm`.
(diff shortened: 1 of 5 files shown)


docs/src/changelog.md
```diff
@@ -10,9 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Model types `sbm_gwf` and `hbv` use the same approach for the calculation of the drain
   width `dw` as model type `sbm`.
-- Renamed `h_bankfull` parameter to `bankfull_depth` for consistency between 
-  kinematic-wave and local-inertial river routing. Note that this may require 
-  updating the `[input.lateral.river]` section of the TOML file!
+- Renamed `h_bankfull` parameter to `bankfull_depth` for consistency between kinematic-wave
+  and local-inertial river routing. When using the old name under the
+  `[input.lateral.river]` section of the TOML file, it will still work but suggest to update
```
verseve, line 15: Suggest a small change: ..it will work but it is suggested to update....

src/flow.jl
```diff
@@ -121,24 +121,31 @@ function initialize_surfaceflow_river(
 
     n_river = ncread(
         nc,
-        config.input,
-        "lateral.river.n";
+        config.input.lateral.river,
```
verseve, line 124: By removing the `lateral.river` part, this part now does not end up in the logging. In `ncread` this String is the `parameter` argument, which is also used for info log messages in `ncread`. Now, it is for example difficult to differentiate between `n` for `land` and `river` domain. If I understand correctly, using `lateral.river.bankfull_depth` (below) is not possible in combination with an alias. But, in case an alias is not used I prefer to use the full info here. Or perhaps you've another idea how we can deal with this @visr ?

verseve (changes requested):
Thanks, I have two comments.

src/flow.jl
visr: Can you check ad058f807141ed12f8c4876186a26b729d378fc6? Should be good to go now. I don't think adding a test is needed, though I did confirm that the old name works and gives an error now.

verseve: Looks good, I also did change a coupe of soil parameters to full parameter name where an alias is used, see https://github.com/Deltares/Wflow.jl/pull/199/commits/06edf23e3bbbe17abc649970b0d07ecf7c738c88.

verseve (approved):

PR #204, opened 2022-03-10
Change unit of lateral subsurface flow (ssf)

verseve:
The unit of lateral subsurface flow (ssf) has been changed from m3/Δt to m3/day, where Δt is the model timestep. Other flow variables are already stored independently from Δt, this allows for easier interpretation and to use states independently of Δt.


JoostBuitink (approved):
Looks good - thanks for fixing this issue!

PR #207, opened 2022-03-24
Fix calculation of h_av as part of 2D overland flow

verseve:
The summation of `h` was not correct, resulting in too low values for `h_av`. For river cells of 2D overland flow `h_av` was only updated as part of the river domain (`ShallowWaterRiver`), this value is now also updated as part of the land domain (`ShallowWaterLand`). The reference level for `h` and `h_av` of 2D overland flow with river cells is changed from river bed elevation `zb` to cell elevation `z`.


laurenebouaziz (approved):
tested the changes made for the Meuse model and the model runs well. The changes in streamflow are small with the corrections that were made. Having the elevation of the water level on the floodplains in river cells (h and h_av) is a good improvement. h_av_land is also corrected in non river cells. 
On river cells, we found that h_av_land and h_av_river - riverdepth can deviate between each other when overbank flow only occurs for some of the timesteps in the iteration, which makes sense. 

I think this PR can be merged into main.

PR #208, opened 2022-04-04
Support compression of NetCDF file

JoostBuitink:
Support using the deflatelevel option when writing the netcdf output data. This option is currently only available for the gridded NetCDF data (so not for the states and scalar output), as this file can become very large. 

The option can be specified in the `[output]` section of the TOML file, via the `compressionlevel` keyword. A value of 0 (default) indicates no compression, and values between 1 and 9 indicate increasing levels of compression, but at the cost of increased computation times. 

Some tests for the Moselle and the Rhine showed that file size reduction is already substantial when using `compressionlevel = 1` (file sizes are reduced by a factor of 4-5), while the total computation time has increased by 5-25%, depending on the basin. Using `compressionlevel = 9`, files are only slightly smaller, but runtimes increased by roughly 40%.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -8,9 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- Added support for NetCDF compression, through the option `compressionlevel` in the
-  `[output]` section in the TOML file. The setting defaults to 0 (no compression), but all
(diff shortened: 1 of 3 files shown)


verseve (approved):
Looks good! I made some small changes to the docs (see https://github.com/Deltares/Wflow.jl/pull/208/commits/e5ff8b60cd4616115bccb0e715de0953e9772ed6)

Issue #209, opened 2022-04-20
Error when creating log in new directory


JoostBuitink:
A SystemError `No such file or directory` is thrown when Wflow tries to create a log.txt file that is positioned in a not yet existing directory. This happens either when specifying a `dir_output` in the TOML, or a directory in the `path_log`. This directory should be created when it not yet exists, like is currently done when writing the output NetCDF and/or CSV files.

PR #210, opened 2022-04-20
Create dir before creating log file

JoostBuitink:
Fix the `No such file or directory` error when Wflow tries to write a log file in a (not yet) existing directory (see https://github.com/Deltares/Wflow.jl/issues/209). Now, the directory is created before trying to create the log file.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+- Fixed an error with the log file, when writing to a folder that does not (yet) exists.
+  Now, the folder is created prior to writing the log file.
+
 ## v0.6.0 - 2022-04-14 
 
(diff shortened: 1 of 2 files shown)


visr:
Thanks! we missed this, we do this for other output files as well.

Side note, seeing #209, do you think there is an added benefit of creating issues in cases like this, where you'll also directly work on the fix? I'd say issues are only useful to keep track of things that must be fixed later, or to discuss what is the problem or how to solve it.

visr (approved):

JoostBuitink:
I was doubting whether it was necessary to create an issue, but considered it might be useful to both flag that the issue has been found, and to keep track of the fixed issues for future releases (while writing this; I realize I should update the changelog as well). However, especially since the fix for this issue is rather straightforward and harmless, I agree it might be a bit "overkill".

visr:
Yeah we can also use labels on PRs to indicate that they fix a bug. In the auto generated release notes, https://github.com/Deltares/Wflow.jl/releases/tag/v0.6.0, both issues and PRs are listed.

It's good to have this in the changelog, but it would have also been fine to leave out, since it's only meant for notable changes (for some definition of notable).

JoostBuitink:
Thanks for the explanation - good to know. Will keep this in mind for future bugfixes!

PR #211, opened 2022-04-26
fix for NCDatasets 0.12.3

visr:
Fix an error in Wflow.read_dims introduced by a new struct that is used in NCDatasets 0.12.3: https://github.com/Alexander-Barth/NCDatasets.jl/commit/fe1af7fed1067b7f0e7a0c22e0ad2859d54ff568

It would probably be wise to release this quickly since otherwise the last Wflow release is broken with this NCDatasets patch release.


JoostBuitink (approved):
Thanks for the quick fix!

Issue #162, opened 2021-09-13
support forcing without exact model grid definition


verseve:
Many wflow applications use forcing (precipitation, potential evapotranspiration and temperature) at a (much) coarser resolution than the actual wflow model grid. Supporting regridding of forcing during a wflow run (in memory) could be a nice addition to wflow for these kind applications.

hcwinsemius:
I would rephrase "nice addition" to something imperatively important where it concerns very large runs :-)

visr:
I suppose that if we set up the cell to cell mapping once at the beginning and reuse it, it doesn't have to add too much runtime overhead. Though it's good to refine a bit what is needed.

- What kind of regridding methods are needed? Just nearest, or also others?
- Does it need to be able to handle different projections? If so, we'd probably need a GDAL / PROJ dependency. If not, we can probably stick to something like https://github.com/JuliaMath/Interpolations.jl
- Is this only for forcing, or do we want this for the staticmaps as well?

verseve:
I agree, some additional info about requirements would be good. @hcwinsemius or @markhegnauer could you please add (see also questions from @visr)? For now, I would say this is only for forcing, I don't see much added value to implement this for staticmaps.

markhegnauer:
Agreed, for now only for forcing needed. A simple interpolation (nearest) would already be very useful. In the future, reprojection would be a nice-to-have.

verseve:
And what about using a lapse rate to correct temperature with the DEM?

Huite:
Just to butt in, uninvited (thank to Github notifications): I'm rather confident that in-memory runtime regridding is a poisoned chalice. 

I feel confident because in-memory runtime regridding has been a staple of iMODFLOW models for as long as I've used them, and I've honestly always hated it. The primary reason is that when debugging your model, it's no longer possible to inspect your model input directly and it gives me a distinct feeling of lacking control -- maybe this is just a personal defect. 

Secondly, as Martijn mentions, it can also quickly spiral out of control in terms of complexity, especially if you consider your staticmaps. In this case, the appropriate regridding method depends on the forcing or parameter in question (with many possible valid schemes: nearest, mean, median, max, min, sum, first-order conserving, etc., etc.; all area-weighted or not). Automatic reprojection is probably ten times worse due to the generally poor understanding of projections and coordinate systems by users.

[...]

PR #214, opened 2022-08-02
Forcing factors

laurenebouaziz:
Added factors for precip and potential evaporation and additive for temperature to be able to run climate scenarios with the delta change method easily in Wflow (without duplicating the forcing file). 
@visr, we tested the changes and they seem to work fine, but could you perform a review of the changes? thanks!

### DIFF
--- a/src/sbm.jl
+++ b/src/sbm.jl
@@ -78,6 +78,12 @@
     temperature::Vector{T} | "°C"
     # Potential evapotranspiration [mm Δt⁻¹]
     potential_evaporation::Vector{T}
+    # Multiplication factor [-] for epot
+    epot_factor::Vector{T} | "-"
+    # Multiplication factor [-] for precip
+    precip_factor::Vector{T} | "-"
+    # Additive factor [-] for temp
+    temp_add::Vector{T} | "-"
     # Potential transpiration, open water, river and soil evaporation (after subtracting interception from potential_evaporation)
     pottrans_soil::Vector{T}
     # Transpiration [mm Δt⁻¹]
@@ -510,6 +516,30 @@ function initialize_sbm(nc, config, riverfrac, inds)
         defaults = 1.0,
         type = Float,
     )
+    epot_factor = ncread(
+        nc,
+        config,
+        "vertical.epot_factor";
+        sel = inds,
+        defaults = 1.0,
+        type = Float,
+    )
+    precip_factor = ncread(
+        nc,
+        config,
+        "vertical.precip_factor";
+        sel = inds,
+        defaults = 1.0,
+        type = Float,
+    )
+    temp_add = ncread(
+        nc,
+        config,
+        "vertical.temp_add";
+        sel = inds,
+        defaults = 0.0,
+        type = Float,
+    )
 
     cmax, e_r, canopygapfraction, sl, swood, kext = initialize_canopy(nc, config, inds)
 
@@ -571,6 +601,9 @@ function initialize_sbm(nc, config, riverfrac, inds)
         cap_hmax = cap_hmax,
         cap_n = cap_n,
         et_reftopot = et_reftopot,
+        epot_factor = epot_factor,
+        precip_factor = precip_factor,
+        temp_add = temp_add,
         c = svectorscopy(c, Val{maxlayers}()),
         stemflow = fill(mv, n),
         throughfall = fill(mv, n),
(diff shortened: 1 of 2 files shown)


visr:
Isn't this already supported? https://deltares.github.io/Wflow.jl/dev/user_guide/step2_settings_file/#Modify-parameters

```toml
[input.vertical.precipitation]
netcdf.variable.name = "P"
scale = 1.5
offset = 0.5
```

laurenebouaziz:
good question, for climate change we are using monthly grids of delta's for each forcing variable. so with the current PR we can add them as cyclic parameter: 
cyclic = [ "vertical.leaf_area_index","vertical.precip_factor","vertical.epot_factor","vertical.temp_add",]
epot_factor = "delta_pr_CNRM-CM6-1_ssp585_far"
where epot_factor in the netcdf has dimension (12) time, y, x. 
I don't think this can be done with the current implementation of modify parameter right?

visr:
Ah indeed, that cannot be done using the modify parameters feature. Nice that you've been able to make it work this way. Do you think this will be used regularly? I'm wondering if this feature is worth including in the main branch. If we decide to include it, it would also need some documentation and tests.

This could also be used as an example that we work out in the docs of a custom run loop, which I mentioned also in https://github.com/Deltares/Wflow.jl/issues/162#issuecomment-921710968.  In here:
https://github.com/Deltares/Wflow.jl/blob/977a4d76b7c6f302204c029fbb57e21e94183933/src/Wflow.jl#L208-L212
If we add a bit of code between `load_dynamic_input` and `update(model)`, there we can update any variable in any way we want, depending on the current time.

laurenebouaziz:
actually I think this is a feature that we may want to use in other projects in which we run climate scenarios, so I think it would be good to include in the main. It will have the advantage that it can then also be ran with the (next) executable. With the custom run loop, it means that we run from Julia right?
And I agree with you that we should then add some documentation and tests. Let me know what you think and I can have a look at documentation and tests.

src/sbm.jl
```diff
@@ -78,6 +78,12 @@
     temperature::Vector{T} | "°C"
     # Potential evapotranspiration [mm Δt⁻¹]
     potential_evaporation::Vector{T}
+    # Multiplication factor [-] for epot
+    epot_factor::Vector{T} | "-"
```
visr, line 82: This is somewhat double on top of the `et_reftopot` factor, though they do serve different purposes, so I guess it's fine to have both.

src/sbm.jl
```diff
@@ -78,6 +78,12 @@
...
+    # Multiplication factor [-] for precip
+    precip_factor::Vector{T} | "-"
+    # Additive factor [-] for temp
```
visr, line 85: Additive factor -> offset. And units are °C.

src/sbm.jl
```diff
@@ -78,6 +78,12 @@
...
+    temp_add::Vector{T} | "-"
```
visr, line 86: Regarding the naming of all three; I'd prefer to be consistent with the existing code as much as possible, so could you use the `<existing name>_<operation>` naming, where operation is scale or offset, and existing name is `potential_evaporation`, etc.

The same goes for the TOML and variable names below.

visr (changes requested):
I left a few minor comments, but if tests and documentation are added I'm ok with adding these.

visr:
> With the custom run loop, it means that we run from Julia right?

In principle yes. Though if executables are needed it is also possible create a branch on this repository with the custom changes to `Wflow.run`, that we wouldn't merge, but we can easily create executables from a branch.

src/sbm.jl
laurenebouaziz: yes, also noticed this, but agree with you that these have different purposes, so I would also leave it like this

src/sbm.jl
laurenebouaziz: good point, changed in 1a9c2c43cfc531832f5c56f4be854853992b03f4

src/sbm.jl
laurenebouaziz: agree, changed in 1a9c2c43cfc531832f5c56f4be854853992b03f4

laurenebouaziz:
thanks for the review! I will try to also add documentation and unit test shortly.

laurenebouaziz:
> > With the custom run loop, it means that we run from Julia right?
> 
> In principle yes. Though if executables are needed it is also possible create a branch on this repository with the custom changes to `Wflow.run`, that we wouldn't merge, but we can easily create executables from a branch.

thanks, good to know!

visr:
Note that there are also 3 BMI tests that need to get updated for the new number of variables; https://github.com/Deltares/Wflow.jl/runs/7631174183?check_suite_focus=true#step:6:142

verseve:
> If we add a bit of code between `load_dynamic_input` and `update(model)`, there we can update any variable in any way we want, depending on the current time.

I think this is a good idea for the proposed functionality. Although the implementation is probably more simple than what was done in the AXA project, I think it makes sense to pull this out the main Wflow functionality (also in case more complex changes are required at some point). Let's discuss further during one of the stand-ups @laurenebouaziz , @visr and @JoostBuitink ?

verseve:
During the wflow stand-up today we did conclude that it is preferred to have this kind of functionality outside of the Wflow computational engine. @laurenebouaziz will save (fork) this branch (as project backup), and after that this branch can be removed. 

@visr: what would be a good location to store these custom run functions, also in relation to building an exe (TeamCity)?

visr:
> what would be a good location to store these custom run functions, also in relation to building an exe (TeamCity)?

For building an exe the easiest is just to modify the main run function in a branch on this repository that is not for merging. We could add this to the docs with some examples of useful customizations.

PR #215, opened 2022-08-16
remove _FillValue from coordinate variables

visr:
There were no actual missing values, but setting the attribute itself is also not allowed.

Quoting https://cfconventions.org/cf-conventions/cf-conventions.html#attribute-appendix

> `_FillValue`: Allowed for auxiliary coordinate variables but not allowed for coordinate variables.


JoostBuitink (approved):
Thanks for fixing this! I was not aware of this "rule" in the CF conventions, but luckily it is a straightforward fix!

PR #217, opened 2022-08-22
Glacier parameter g_sifrac (conversion)

verseve:


hboisgon (approved):
All good! One comment, you updated the units in the docs/params_vertical but maybe also good to update the unit in the structures as well (ie in sbm.jl, hbv.jl, flextopo.jl) ?

verseve:
> One comment, you updated the units in the docs/params_vertical but maybe also good to update the unit in the structures as well (ie in sbm.jl, hbv.jl, flextopo.jl) ?

Yes, good suggestion! I did update this in commit https://github.com/Deltares/Wflow.jl/pull/217/commits/2c7580ecca5a3881037579b8448412e23d8bed64.

Issue #221, opened 2022-09-27
Improve error when _FillValue in time dimension


JoostBuitink:
Throw an error when _FillValue is present in the time dimension, as Wflow is unable to run with this attribute present. Currently, an error is thrown in the `update_forcing!` function, at the following line https://github.com/Deltares/Wflow.jl/blob/0e0b50bbc69ca84ea910ea8637a19f8a21e18d1d/src/io.jl#L270

Current workaround (with Python, xarray): set `ds.time.encoding["_FillValue"] = None` prior to writing the netcdf file. See also https://github.com/Deltares/hydromt_wflow/issues/60

visr:
So it currently throws an error when the time dimension has _FillValue, even if there are no actual missing times? Or only if there are missing times?

Probably we can also just support this rather than throwing an error. We'd need to handle missings explicitly in https://github.com/Deltares/Wflow.jl/blob/0e0b50bbc69ca84ea910ea8637a19f8a21e18d1d/src/io.jl#L162-L174.

We'd probably have to widen the type signature for `times`, and use `skipmissing`:

```
julia> findfirst(>=(3), skipmissing([missing, 2, 3, missing]))
3
```

verseve:
> So it currently throws an error when the time dimension has _FillValue, even if there are no actual missing times? Or only if there are missing times?

In this case an error was thrown with a _FillValue in the time dimension, without missing times. Missing data is not allowed in coordinate variables according to NetCDF metadata conventions, so I think it is preferred to throw an error instead of supporting this in Wflow.

visr:
Ok, yes throwing an error is definitely defensible.

Related, right now we do accept _FillValue without actual missing in the spatial coordinate dimensions, and hydromt_wflow writes it with _FillValue currently: https://github.com/Deltares/hydromt_wflow/issues/60#issuecomment-1216508822

PR #229, opened 2022-11-09
Allow for _FillValue in time dimension

JoostBuitink:
Fixed an issue where the `_FillValue` attribute in the time dimension of the forcing file causes an error, as the `get_at` function was unable to dispatch due to unexpected types. This addition checks (if the `_FillValue`  is present) whether there are actual missing values in the time dimension. If this is the case, an error is thrown, otherwise the dates are converted to the correct type for the `get_at` function.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
+- Removed error when `_FillValue` is present in the time dimension of the forcing NetCDF file.
+  The simulation is allowed to continue with the attribute present, given that there are no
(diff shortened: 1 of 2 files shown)


visr (approved):
Looks good to me! I assume you tested this locally. Might be good to add that is is not allowed according to CF. But this is nicer/more forgiving to users. If missings in practice often occur (e.g. from Delft-FEWS) we could probably even support it as discussed in https://github.com/Deltares/Wflow.jl/issues/221#issuecomment-1270695767, though if needed that can always be done later.

JoostBuitink:
Thanks for the quick review! I indeed tested this locally. I have added a warning message when `_FillValue` is present in the time dimension, to give feedback to the user. Also moved the check for actual missing values to outside of the `update_forcing!` function.

PR #233, opened 2022-11-18
optional 1d floodplain schematization for local inertial river flow

verseve:


JoostBuitink (changes requested):
Nice work! I have added some comments, which are mainly related to dealing with the extrapolation of floodplain data

src/flow.jl
```diff
@@ -1102,6 +1142,278 @@ function update(sw::ShallowWaterLand{T}, swr::ShallowWaterRiver{T}, network, Δt
...
+        a = sw.hf[i] * sw.width[j]
+    else
+        flood_depth = sw.hf[i] - sw.bankfull_depth[j]
+        a_channel = sw.width[j] * sw.bankfull_depth[j]
+        i1, i2 = interpolation_indices(flood_depth, sw.floodplain.depth)
+
+        Δh = flood_depth - sw.floodplain.depth[i1]
+        slope = sw.floodplain.slope[j][i2]
+        # rectangular floodplain segment (also for extrapolation (i1==i2))
+        if slope == T(Inf) || i1 == i2
```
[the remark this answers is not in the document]
verseve: Good point, I agree and have changed this.

src/flow.jl
```diff
@@ -1102,6 +1142,278 @@ function update(sw::ShallowWaterLand{T}, swr::ShallowWaterRiver{T}, network, Δt
...
+function interpolate_roughness_nodes(sw::ShallowWaterRiver, i::Int, i_src::Int, i_dst::Int)
+
+    if sw.hf[i] <= sw.bankfull_depth[i_src]
+        n_src = sw.mannings_n[i_src]
+    else
+        flood_depth = sw.hf[i] - sw.bankfull_depth[i_src]
+        i1, i2 = interpolation_indices(flood_depth, sw.floodplain.depth)
+        j = max(nearest_neighbor_index(flood_depth, sw.floodplain.depth), 2)
+
+        Δh = i1 == i2 ? 0.0 : flood_depth - sw.floodplain.depth[i1]
```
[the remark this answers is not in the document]
verseve: Changed (see first comment).

src/flow.jl
```diff
@@ -1102,6 +1142,278 @@ function update(sw::ShallowWaterLand{T}, swr::ShallowWaterRiver{T}, network, Δt
...
+            ) / (sw.width[i_src] + fw)
+    end
+    if sw.hf[i] <= sw.bankfull_depth[i_dst]
+        n_dst = sw.mannings_n[i_dst]
+    else
+        flood_depth = sw.hf[i] - sw.bankfull_depth[i_dst]
+        i1, i2 = interpolation_indices(flood_depth, sw.floodplain.depth)
+        j = max(nearest_neighbor_index(flood_depth, sw.floodplain.depth), 2)
+
+        Δh = i1 == i2 ? 0.0 : flood_depth - sw.floodplain.depth[i1]
```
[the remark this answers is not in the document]
verseve: Changed (see first comment).

src/flow.jl
```diff
@@ -1102,6 +1142,278 @@ function update(sw::ShallowWaterLand{T}, swr::ShallowWaterRiver{T}, network, Δt
...
+        p[2:end, i] = cumsum(p[2:end, i])
+        a[:, i] = cumsum(a[:, i])
+    end
+
+    if incorrect_vol > 0
+        @warn string(
+            "The provided volume of $incorrect_vol rectangular floodplain schematization",
+            " segments is not correct and has been adapted.",
+        )
+    end
```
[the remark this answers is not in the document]
verseve: I agree and have added the extra info to the warning message.

src/sbm_model.jl
```diff
@@ -25,6 +25,7 @@ function initialize_sbm_model(config::Config)
     kw_land_tstep = get(config.model, "kw_land_tstep", 0)
     kinwave_it = get(config.model, "kin_wave_iteration", false)::Bool
     routing_options = ("kinematic-wave", "local-inertial")
+    floodplain_1d = get(config.model, "floodplain_1d", false)::Bool
```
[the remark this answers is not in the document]
verseve: I have added this option to the initialization part of the local inertial river routing (`function initialize_shallowwater_river`).

src/flow.jl
```diff
@@ -1382,12 +1384,16 @@ function initialize_floodplain_1d(nc, config, inds, riverwidth, riverlength, ind
         p[2:end, i] = 2.0 * p_unit[2:end, i] .* h
         p[2:end, i] = cumsum(p[2:end, i])
         a[:, i] = cumsum(a[:, i])
+
+        riv_cells += min(incorrect_vol, 1)
```
JoostBuitink, line 1393: This counter is not correctly keeping track of the river cells. A counter (or flag) should be reset at the start (but inside) the for loop, to correctly count the number of affected river pixels

src/flow.jl
```diff
@@ -1102,6 +1145,287 @@ function update(sw::ShallowWaterLand{T}, swr::ShallowWaterRiver{T}, network, Δt
...
+    i_dst::Int,
+    p_src::Real,
+    p_dst::Real,
+)
+
+    if sw.hf[i] <= sw.bankfull_depth[i_src]
+        n_src = sw.mannings_n[i_src]
+    else
+        flood_depth = sw.hf[i] - sw.bankfull_depth[i_src]
+        j = max(nearest_neighbor_index(flood_depth, sw.floodplain.depth), 2)
```
JoostBuitink, line 1211: Why is this value maximized at 2? It is used to select the correct manning value, which should be defined at the same intervals as the `sw.floodplain.depth` values right?

verseve: This is because index 1 is at flood depth 0.0 (convenient for interpolation), without a manning's n value for the floodplain.

docs/src/model_docs/params_lateral.md
```diff
@@ -198,6 +200,49 @@ bankfull_depth = "RiverDepth"
...
+The Table below shows the parameters (fields) of struct `FloodPlain` (part of struct
+`ShallowWaterRiver`), including a description of these parameters, the unit, and default
+value if applicable. The parameters in bold represent model parameters that can be set
+through static input data (netCDF), and can be listed in the TOML configuration file under
+`[input.lateral.river.floodplain]`, to map the internal model parameter to the external
+netCDF variable.
+
+|  parameter  | description  	  | unit  | default |
+|:--------------- | ------------------| ----- | -------- |
+| **`profile`** (`n`)    |  Floodplain profile `FloodPlainProfile` | | |
```
JoostBuitink, line 215: This `n` should not be here right?

src/flow.jl
```diff
@@ -440,8 +440,10 @@ end
     q::Vector{T} | "m3 s-1"                 # river discharge (subgrid channel)
     q0::Vector{T} | "m3 s-1"                # river discharge (subgrid channel) at previous time step
     q_av::Vector{T} | "m3 s-1"              # average river discharge [m³ s⁻¹]
+    q_total_av::Vector{T} | "m3 s-1"        # average total (river and floodplain) discharge [m³ s⁻¹]
```
JoostBuitink, line 443: While I agree that it is nice to have the option to keep the river discharge and combined (river+floodplain) discharge separate, I am thinking whether it is a good idea to have the combined output under a different variable name. If a user switches between using and not using 1d floodplains, this means that the output variables need to be changed as well. Is it therefore not more user-friendly to have the combined output under `q_av` and only the river discharge under something like `q_river_av`?

Regardless of the choice, it is good to have this explained in the documentation as well (it is currently missing from the parameter tables as well if I am right).

JoostBuitink (approved):
Thanks for the great work @verseve! I have two (minor) comments, but other than that, the code as well as the results for the Rhine look good. If @laurenebouaziz agrees, I think we are good for merging!

docs/src/model_docs/params_lateral.md
verseve: Yes, well spotted, I have remove it.

src/flow.jl
```diff
@@ -1102,6 +1235,236 @@ function update(sw::ShallowWaterLand{T}, swr::ShallowWaterRiver{T}, network, Δt
...
+        for j = 1:(n_depths-1)
+            # assume rectangular shape of flood depth segment
+            width[j+1, i] = diff_volume[j] / (h[j] * riverlength[i])
+            # check provided flood volume (floodplain width should be constant or increasing
+            # as a function of flood depth)
+            if width[j+1, i] < width[j, i]
+                incorrect_vol += 1
+                riv_cell = 1
+                error_vol =
+                    error_vol + ((width[j, i] - width[j+1, i]) * h[j] * riverlength[i])
```
DirkEilander, line 1392: After fixing this in https://github.com/Deltares/hydromt_wflow/pull/123 I'm still getting a warning here. The diff in width is however very small and I suspect it's a rounding error. These rounding errors might be important for the calculation, but warning message like the one below might also confuse users. 

> The provided volume of 734 rectangular floodplain schematization segments for 132 river cells (6.63 % of total river cells) is not correct and has been increased with 0.0 % of provided volume.

I suggest to only raise a warning if the error is relative large, for instance by allowing for a 1cm diff in width:

```suggestion
                # raise warning only if diff is larger than rounding error of 1 cm
                if (width[j, i] - width[j+1, i]) > 0.01
                    incorrect_vol += 1
                    riv_cell = 1
                    error_vol =
                        error_vol + ((width[j, i] - width[j+1, i]) * h[j] * riverlength[i])
                end
```

src/flow.jl
verseve: Very good point, I agree and made the required changes. For only river discharge the parameter `q_channel_av` is used. I also added a bit more info (parameters and states) to the documentation.

src/flow.jl
verseve: Good suggestion! Changed this in commit https://github.com/Deltares/Wflow.jl/pull/233/commits/0a4ffa88e0123fd4c48947f967cc2c377cb1e233.

verseve: With this suggested change for the Rhine model % of total river cells not correct dropped from 41.82 to 0.46, while the increase in total volume (correction) stayed the same (0.18 % of provided volume). For the Rhine model there is also a mismatch between river width used in hydromt_wflow and in Wflow I think (`wflow_riverwidth_sobek_global`), explaining the required correction. In addition @JoostBuitink did some flood profile testing, and also concluded that mismatches were mostly caused by rounding errors.

laurenebouaziz:
With @JoostBuitink, we compared the results of the 1d floodplain with the 1d2d for the Meuse and the Rhine and came to the following conclusions:

- peak flows are lower and more attenuated in the 1d2d local inertial routing compared to the 1d floodplain routing. Results of the 1d floodplain routing in terms of maximum annual flows are often closer to local inertial 1d (no floodplains) or kinematic wave runs. These differences can likely be explained by: 


- the smaller flood volume in the 1d floodplain schematization compared to the 1d2d schematization. (a test run with floodplain volume * 3 shows much more attenuation along the Meuse river, especially in the French part of the Meuse where we actually would expect some effect from floodplain routing -- maybe we underestimate the current flood volume in this part of the river in the current set-up...)
- in the 1d2d schematization water travels along the D4 network, which likely results in more delay of the floodplain flow compared to the 1d schematization. 
- in the 1d floodplain schematization, part of the floodplain flow is "on top" of the river which has the same manning as the river, leading to less damping than would have occurred in the 1d2d schematization. 

So I think the results we get are in line with what can be expected, and therefore I also think we are good to merge.

Also agree to change the q_total_av to q_av to avoid confusions between different routing options. 

thanks @verseve for all the work!

PR #240, opened 2022-12-14
`NCDatasets` and multifile reading

verseve:
- Issue with reading multifile NetCDF dataset with `NCDatasets` (introduced with v0.12.4) solved in v0.12.11.
- Store the `time` dimension of forcing NetCDF file as part of struct `NCReader` instead of calling `dataset["time"'][:]` each time step when loading forcing data.
- Removed this check: https://github.com/Deltares/Wflow.jl/blob/d1dcbfefa4f7937cceaf6c22feae3e1e0c14dd14/src/io.jl#L248-L252

  it seems not necessary because we already check for (and do not allow) missings when the `NCReader` is initialized.


src/io.jl
```diff
@@ -629,6 +621,7 @@ end
 
 struct NCReader
     dataset::CFDataset
+    dataset_times::Vector{DateTime}
```
visr, line 624: If I remember correctly NCDatasets doesn not always return a `Vector{DateTime}` for `dataset["time"][:]`, but, depending on the calendar used, other time types from CFTime.jl. We'd need to test this with a netCDF with such a calender though. I'd imagine we'd need `struct NCReader{T}` with `dataset_times::Vector{T}`.

verseve: Yes, good point! Indeed, other time types are possible with NCDatasets depending on the calendar used. Suggested changes are in commit https://github.com/Deltares/Wflow.jl/pull/240/commits/9f02023b088ba5bcf92b49d80ed5e9af367c601f.

visr (approved):
This looks good to me.

Ideally we'd have a test where we get forcing values from a netCDF file that doesn't have standard DateTime timestamps. The tests for this currently only cover the Wflow.Clock and parsing the calendar: https://github.com/Deltares/Wflow.jl/blob/a41dc5317747e1ee12119116c9d99d22c95e8428/test/io.jl#L140-L164

Therefore the tests didn't fail without a type parameters.

verseve:
> Ideally we'd have a test where we get forcing values from a netCDF file that doesn't have standard DateTime timestamps. The tests for this currently only cover the Wflow.Clock and parsing the calendar:

I agree, and added a test that reads a (small) forcing NetCDF file with a `noleap` calendar, see also https://github.com/Deltares/Wflow.jl/pull/240/commits/bed8ec4e8baf1eea0ebb0373ff0b5f6a75686176.

PR #241, opened 2023-01-12
Fix time cyclic input and lake rating curve (column index)

verseve:


visr (approved):
Looks good to me. Would be good to add a few unit tests to make sure `julian_day` keeps workin as intended.

```
julia> julian_day(DateTime(2000,2,28))
59

julia> julian_day(DateTime(2000,2,29))
60

julia> julian_day(DateTime(2000,3,1))
60
```

verseve:
> Looks good to me. Would be good to add a few unit tests to make sure `julian_day` keeps workin as intended.

Thanks for the quick review! Good suggestion, I did add a few units tests.

Issue #242, opened 2023-01-20
HQ approach for reservoirs


verseve:
Add SQ (volume based) approach based on thesis work about the Rur catchment in Germany. An HQ and SH approach already exists for the lakes. This is probably also a good moment to merge the Wflow lake and reservoir functionality.

verseve:
Comment from @shartgring :
We first discussed this implementation through modelling the reservoirs as lakes, and adding a new LakeOutflowFunc for the SQ-approach. However, it is also possible by adding the functionality to the reservoirs

ResOutflowFunc is determined by checking if .csv file corresponding to reservoir exists. If it exists, the reservoir will use the SQ-approach. If not, then it is modelled using the old approach. It is a 'quick fix', but I would say it is a good alternative until lakes and reservoirs are merged into 'waterbody'. Advantage of this fix is that it uses parameters such as ResMaxVolume which are generated by hydromt_wflow, and that it will work with previous Wflow models (no breaking).

Code passed tests and 2 new tests are added, for normal flow and for a overflowing reservoir. Results correspond with hand calculations. Attached the csv file used in the test, maybe nice to upload to artifact repo if this will be implemented

[res_sq_1.csv](https://github.com/Deltares/Wflow.jl/files/10849467/res_sq_1.csv)

verseve:
After some discussion we decided to use the existing functionality as part of the Lake module (HQ approach), and add overflow here, to keep the code simple and to avoid (partly) duplication of code. A lake with a rating curve of type 1 (HQ table) is a regulated lake/ dam. Also, for now we keep the reservoir and lake functionality separate, merging is still an option, and can be part of another PR (need to think more about implications for Wflow.jl as well as hydromt_wflow).

PR #244, opened 2023-02-28
Add SQ functionality to reservoirs

shartgring:
See also issue #242

We first discussed this implementation through modelling the reservoirs as lakes, and adding a new LakeOutflowFunc for the SQ-approach. However, it is also possible by adding the functionality to the reservoirs

ResOutflowFunc is determined by checking if .csv file corresponding to reservoir exists. If it exists, the reservoir will use the SQ-approach. If not, then it is modelled using the old approach. It is a 'quick fix', but I would say it is a good alternative until lakes and reservoirs are merged into 'waterbody'. Advantage of this fix is that it uses parameters such as ResMaxVolume which are generated by hydromt_wflow, and that it will work with previous Wflow models (no breaking).

Code passed tests and 2 new tests are added, for normal flow and for a overflowing reservoir. Results correspond with hand calculations. Attached the csv file used in the test, maybe nice to upload to artifact repo if this will be implemented

[res_sq_1.csv](https://github.com/Deltares/Wflow.jl/files/10849467/res_sq_1.csv)

### DIFF
--- a/src/io.jl
+++ b/src/io.jl
@@ -1374,9 +1374,17 @@ function read_hq_csv(path)
     return (H = data[:, 1], Q = data[:, 2:end])
 end
 
+"Read a storage-rating curve from CSV into a NamedTuple of vectors"
+function read_sq_csv(path)
+    data = readdlm(path, ',', Float, skipstart = 1)
+    # Q is a matrix with 365 columns, one for each day in the year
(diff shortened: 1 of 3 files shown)


verseve:
Closed this PR for now, can reopen later when lakes and reservoir functionality has been merged.

PR #245, opened 2023-03-02
define BMI.get_time_units on model instance

visr:
Perhaps this was originally done since it doesn't need any information from the model instance, but since this is the only one, this looks more like a bug to me.


visr:
Pushed another commit after realizing the return value should be "s" to be in line with BMI docs.

PR #253, opened 2023-04-13
Update codecov-action

Hofer-Julian:
https://github.com/codecov/codecov-action#%EF%B8%8F--deprecation-of-v1


verseve:
Thanks for the update.

PR #272, opened 2023-06-09
Add `sum` to reducer options

JoostBuitink:
The option to take the sum over a certain region was not possible. This allows the user to use `sum` as a reducer. This can be useful in cases where (for example) we want to sum the amount of water flowing into the river for each subcatchment/region.


verseve (approved):
Useful addition! One minor comment: would be good to add this option to the [docs](https://deltares.github.io/Wflow.jl/dev/user_guide/step2_settings_file/#Output-CSV-section) and to the tests (at least [unit](https://github.com/Deltares/Wflow.jl/blob/sum_scalar_output/test/io.jl#L232)).

PR #283, opened 2023-07-05
Kinematic wave lateral subsurface flow

verseve:
Use a constant celerity (per time step), based on the average hydraulic conductivity of the saturated store, for the solution of the kinematic wave for lateral subsurface flow. It allows for an easier extenstion to other saturated hydraulic conductivity profiles (with depth), without the requirement to update the water table depth `zi` during Newton-Raphson iterations.


verseve:
@JoostBuitink did check the results for a small basin (piave). Simulated low flows are much more jumpy with this solution. Also after improving ksat calculation in line 85 (could become Inf), and limiting `ssf` (could become larger than available volume when cell dries). Also. the current kinematic wave solution is more appropriate for an exponential declining ksat with depth. For other ksat profiles we can implement another solution and can work on this in another branch/PR.

PR #306, opened 2023-10-23
BMI: bug fixes and updates

### LINKED ISSUE #304: BMI get_value and grid type
### Wflow version checks

### Reproducible Example

Request of grid type seems wrong for some variables. As an example, for `lateral.subsurface.to_river`: request of grid type is incorrect. With BMI.get_var_grid, grid=2 is returned. However, the Vector that is returned with BMI.get_value has the size of grid=4.

### Current behaviour

BMI.get_value deviates from the BasicModelInterface.jl implementation (`dest` argument is missing).
Grid type for some variables seems wrong (see also Reproducible Example).
BMI.get_value should always return a flattened array, this is not the case for variables with an extra dimension (e.g. `layer`), then a Vector of SVectors is returned.

### Desired behaviour

BMI.get_value should have `dest` as argument.
Grid type should be correct for each variable. This has already been resolved in the zmq_server branch (coupling with OpenDA), but good to include in a separate branch that addresses this issue.
BMI.get_value should always return a flattened array (probably need to make use of get_grid_z).

### Additional Context

These BMI issues were reported during an update of eWaterCycle to support Wflow.jl.

verseve:
## Issue addressed
Fixes #304, #307

## Explanation
Issue #304: 
Fixed Basic Model Interface (BMI) functions that deviated from BMI specifications (BasicModelInterface.jl). Model variables that represent 3-dimensional data (stored as 2d Arrays or Vector of `SVectors`) are exposed as 2-dimensional unstructured grids. This minimizes the number of required grid types and users are most probably interested in data per third grid coordinate (for example per soil layer). BMI grid information (type and location) and whether a model variable can be exchanged is now stored as part of Struct metadata. For Wflow components that store model variables on grid `edges` the BMI model grid functions have been extended with `get_grid_edge_count` and `get_grid_edge_nodes`.

Issue #307:
`config.starttime` and `config.endtime` can also be provided as Strings in the TOML file, so conversion to `DateTime` is also required in BMI functions where these variables are used.

## Checklist

## Additional Notes
A total of 7 different grid types are required and from a user perspective fewer grid types would be better. To support fewer grid types, changes at the Wflow `Model` level are required (see also https://github.com/Deltares/Wflow.jl/issues/304#issuecomment-1771121650). This can be part of another PR.

Also added state and time functions (extension of BMI) that can be used for coupling with OpenDA (and other external) software.


src/reservoir_lake.jl
```diff
@@ -1,5 +1,5 @@
-@get_units @with_kw struct SimpleReservoir{T}
-    Δt::T | "s"                                         # Model time step [s]
+@get_units @exchange @grid_type @grid_location @with_kw struct SimpleReservoir{T}
+    Δt::T | "s" | 0                                     # Model time step [s]
```
visr, line 2: Here you use `| 0` and in other places ` | 0 | "none" | "none"`. Is there a reason it isn't consistent? Either way is fine by me. If I understand correctly if exchange is 0, the grid type and var location should never be read anyway.

src/bmi.jl
```diff
@@ -1,19 +1,17 @@
...
-    2 => "unstructured",
-    3 => "unstructured",
-    4 => "unstructured",
-)
-
 # Mapping of grid identifier to a key, to get the active indices of the model domain.
 # See also function active_indices(network, key::Tuple).
-const grids =
-    Dict{Int,String}(0 => "reservoir", 1 => "lake", 2 => "river", 3 => "drain", 4 => "land")
+grids = Dict{Int,Tuple}(
```
visr, line 6:
```suggestion
grids = Dict{Int,Tuple{Symbol}}(
```

What is the reason this is not `const` anymore? The `const` can give performance benefits. From Julia 1.8+ you can also annotate the types of globals, which gives similar performance: `grids::Dict{Int,Tuple{Symbol}} = Dict{Int,Tuple{Symbol}}(...`

src/bmi.jl
```diff
@@ -125,58 +140,82 @@ function BMI.get_output_var_names(model::Model)
 end
 
 function BMI.get_var_grid(model::Model, name::String)
```
visr, line 142: Not for this PR, but I often prefer to add return type annotations for readability, `BMI.get_var_grid(model::Model, name::String)::Int`.

visr (approved):
Left some comments, but overall this looks good to me!

src/bmi.jl
visr: Ok I see. With or without const it is currently a global, and non-const globals are generally slow. So unless it is moved into a function, it's probably best to keep it const.

verseve: Ok, changed in https://github.com/Deltares/Wflow.jl/pull/306/commits/8cb00df93d7781b0f43ff46926f3b3e39921ebc4.

PR #311, opened 2023-11-16
update Project compat entries

visr:
## Issue addressed
Fixes CI

## Explanation
This updates various compatibilities to fix CI. For stdlibs, this updates the compat to the [latest recommendation](https://discourse.julialang.org/t/psa-compat-requirements-in-the-general-registry-are-changing/104958). The reason that CI was failing was because Aqua 0.8 just came out, which now requires compat entries for all test dependencies as well. This also includes Aqua itself, which will stop breakage from breaking Aqua releases.


verseve:
Thanks for this update!

PR #320, opened 2023-12-18
Update publications.md

aweerts:
added several pub;lications, added section Datasets etc

## Issue addressed
Fixes #<issue number>

## Explanation

## Checklist
- [ ] Updated changelog.md if needed

### DIFF
--- a/docs/src/intro/publications.md
+++ b/docs/src/intro/publications.md
@@ -115,8 +115,9 @@ asynchronous filtering for flood forecasting. Hydrol. Earth Syst. Sci., 19, 2911
 <https://doi.org/10.5194/hess-19-2911-2015>.
 
 Ratri,D. N., A.H. Weerts, R. Muharsyah, K. Whan, A. Klein Tank, E. Aldrian, M. H. Hariadi,
-Calibration of ECMWF SEAS5 based streamflow forecast in Seasonal hydrological forecasting for Citarum river basin, West Java, Indonesia,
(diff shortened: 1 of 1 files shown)


docs/src/intro/publications.md
```diff
@@ -164,8 +168,20 @@ Wang, X., Zhang, J., Babovic, V., 2016. Improving real-time forecasting of water
 indicators with combination of process-based models and data assimilation technique.
 Ecological Indicators 66, 428–439. <https://doi:10.1016/j.ecolind.2016.02.016>.
 
+### Datasets
```
verseve, line 171: Thanks for updating the list of publications @aweerts . Nice to also list datasets. I would suggest to only list these if the dataset is not part of a report or paper (to minimize the work to keep track of these datasets and updating the list) , otherwise I think adding the report/paper is fine. Are/will these datasets be part of a publication?
 
@JoostBuitink , @dalmijn, @alimeshgi: what do you think, do you have a preference?

JoostBuitink: I agree to only list a dataset if it is not part of a publication.

Maybe it is nice to move the datasets to the bottom of this page, to put more emphasis on the reports?

aweerts: yes the datasetswill be part of publications (under review)
Hydrological Reference Dataset (Tier 1) for the ESA 4DHydro Open Science Catalogue
Droppers et al. 2024, submitted to Scientific Data

verseve (approved):
Thanks for the update. As the listed datasets will be part of a publication, I have removed these.

PR #325, opened 2024-01-11
Kvfrac leakage

verseve:
## Issue addressed
Issue was not reported.

## Explanation
Include `kvfrac` in the computation of vertical saturated hydraulic conductivity at the bottom of the soil layer, so it is consistent with the unsaturated vertical flow computation.

## Checklist


JoostBuitink (approved):
Nice fix, good to be consistent indeed ;)

As we discussed, in the future, it might be nice to have recharge presented in a more "user-friendly" way (m/dt or m3/d). If you agree, I can make an issue so we wont forget.

verseve:
> As we discussed, in the future, it might be nice to have recharge presented in a more "user-friendly" way (m/dt or m3/d). If you agree, I can make an issue so we wont forget.

Yes, I agree, would be good to make an issue for this!

PR #366, opened 2024-03-04
Add CLI tests

Hofer-Julian:
## Explanation
- Add CLI tests
- Track Manifest files
- Move downloading test data to own file
- Add pixi tasks

## Checklist
- [ ] Updated changelog.md if needed


.gitignore
```diff
@@ -4,6 +4,9 @@
 # VS Code stuff
 .vscode
 
+# Manifest toml from project toml
+**/Manifest.toml
```
Hofer-Julian: Keep in mind that the Manifest needs to be checked in for this branch to work.

Hofer-Julian: You need the Manifest in order to depend on the local Wflow, rather than the published one.
It is also a good idea to check it in, in order to keep builds reproducible and synchronized between machines.

JoostBuitink: I see! But if I look at the current checked in manifests, they do not point to the local Wflow version, but to the released version. Is that correct?

build/wflow_cli/Manifest.toml
```diff
@@ -0,0 +1,641 @@
...
+deps = ["ArrayInterface", "CPUSummary", "HostCPUFeatures", "IfElse", "LayoutPointers", "Libdl", "LinearAlgebra", "SIMDTypes", "Static", "StaticArrayInterface"]
+git-tree-sha1 = "7209df901e6ed7489fe9b7aa3e46fb788e15db85"
+uuid = "3d5dd08c-fd9d-11e8-17fa-ed2836048c2f"
+version = "0.21.65"
+
+[[deps.Wflow]]
+deps = ["BasicModelInterface", "CFTime", "Dates", "DelimitedFiles", "FieldMetadata", "Glob", "Graphs", "IfElse", "LoggingExtras", "LoopVectorization", "NCDatasets", "Parameters", "Polyester", "ProgressLogging", "Random", "StaticArrays", "Statistics", "TOML", "TerminalLoggers", "UnPack"]
+git-tree-sha1 = "29fdf975f5a32b15890182f114b72ea6ba7b21cc"
+repo-rev = "master"
+repo-url = "../.."
```
Hofer-Julian, line 601: This is relative, right?

build/create_binaries/Manifest.toml
```diff
@@ -0,0 +1,679 @@
...
+deps = ["ArrayInterface", "CPUSummary", "HostCPUFeatures", "IfElse", "LayoutPointers", "Libdl", "LinearAlgebra", "SIMDTypes", "Static", "StaticArrayInterface"]
+git-tree-sha1 = "7209df901e6ed7489fe9b7aa3e46fb788e15db85"
+uuid = "3d5dd08c-fd9d-11e8-17fa-ed2836048c2f"
+version = "0.21.65"
+
+[[deps.Wflow]]
+deps = ["BasicModelInterface", "CFTime", "Dates", "DelimitedFiles", "FieldMetadata", "Glob", "Graphs", "IfElse", "LoggingExtras", "LoopVectorization", "NCDatasets", "Parameters", "Polyester", "ProgressLogging", "Random", "StaticArrays", "Statistics", "TOML", "TerminalLoggers", "UnPack"]
+git-tree-sha1 = "29fdf975f5a32b15890182f114b72ea6ba7b21cc"
+repo-rev = "master"
+repo-url = "../.."
```
Hofer-Julian, line 627: Same here

build/wflow_cli/Manifest.toml
JoostBuitink: Ah, I somehow (accidentally) updated my local Manifest files, so I was looking at the wrong versions there. Ignore my comment then ;). This makes indeed sense, sorry for the confusion!

pixi.toml
```diff
@@ -0,0 +1,50 @@
...
+    "Joost Buitink <joost.buitink@deltares.nl",
+    "Willem van Verseveld <willem.vanVerseveld@deltares.nl",
+    "Hofer-Julian <julian.hofer@deltares.nl>",
+]
+channels = ["conda-forge"]
+platforms = ["win-64", "linux-64"]
+
+[tasks]
+# Installation
+install-julia = "juliaup add 1.10.0 && juliaup default 1.10.0"
```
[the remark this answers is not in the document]
visr: Indeed the Manifest should be the same as the pinned version. At first I though specifying a minor release like `1.10` would be sufficient, but I came back from that after first 1.10.1 broke our Linux builds due to a networking change and 1.10.2 broke Python interop on Linux due to a compiler change.

So I suggest you lock to 1.10.0 until 1.10.3 as well, see https://github.com/Deltares/Ribasim/issues/1217.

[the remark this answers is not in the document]
JoostBuitink: Thanks for clarifying this @visr! I also noticed that in the `[compat]` section in our Project.toml files, we have some different julia versions mentioned there. I guess it is best to replace all of those with `julia = 1.10`, or not?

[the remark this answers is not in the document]
visr: Not necessarily, if you do that then newer versions of Wflow cannot install in 1.9 for instance. Often you only drop support when you actually require 1.10 for something.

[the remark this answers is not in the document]
verseve: Ok, but then we might as well revert this back to Julia v1.6?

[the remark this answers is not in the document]
visr: If you want, but not necessarily. Users on unsupported julia versions can still install Wflow, but they will get the last supported version. For Wflow 1.0 it wouldn't be crazy to only support the current stable release.

build/wflow_cli/test/runtests.jl
```diff
@@ -96,3 +80,19 @@ end
         @test filesize(joinpath(outputdir, file)) > 0
     end
 end
+
+@testset "wflow_sbm_timing" begin
+
+    ENV["JULIA_NUM_THREADS"] = 1
```
Hofer-Julian, line 86: Probably better to set the environment locally as shown here:
https://docs.julialang.org/en/v1/manual/running-external-programs/#Cmd-Objects

JoostBuitink: Thanks for this suggestion, this is indeed a better way to do this!

JoostBuitink:
@Hofer-Julian, it appears the the WflowServer tests here use the master branch and not the code from this branch. Do you know why this is happening?

server/test/client.jl
```diff
@@ -33,7 +33,7 @@ end
 
 @testset "model information functions" begin
     @test request((fn = "get_component_name",)) == Dict("component_name" => "sbm")
-    @test request((fn = "get_input_item_count",)) == Dict("input_item_count" => 181)
+    @test request((fn = "get_input_item_count",)) == Dict("input_item_count" => 180)
     @test request((fn = "get_output_item_count",)) == Dict("output_item_count" => 181)
```
visr, line 37: I feel like checking the precise numbers here and in the Wflow BMI tests doesn't help much and needs a lot of maintenance. Should we just check if it is more than a hundred or something like that?

Hofer-Julian:
> @Hofer-Julian, it appears the the WflowServer tests here use the master branch and not the code from this branch. Do you know why this is happening?

I don't see how that could happen

server/test/client.jl
JoostBuitink: Good point! We have decided to keep this test as is, but made the tests below a bit more flexible. Maybe we want to reconsider this in the future, but for now we are fine with the current implementation.

JoostBuitink (approved):

JoostBuitink:
> > @Hofer-Julian, it appears the the WflowServer tests here use the master branch and not the code from this branch. Do you know why this is happening?
> 
> I don't see how that could happen

Just for completeness (as it was already discussed on Teams); I have added a dev install in the pixi task that fixed this issue.

PR #368, opened 2024-03-07
Add floodplain discharge to inflow reservoirs and lakes

### LINKED ISSUE #367: Local inertial routing: floodplain discharge is not added to inflow reservoir or lake
### Wflow version checks

### Reproducible Example

Run a wflow_sbm model with local inertial routing including floodplain routing and a reservoir.

### Current behaviour

Floodplain discharge from the upstream cell of the reservoir is not added to the inflow of the reservoir. The floodplain routing from the reservoir cell to a downstream cell (edge) seems inactive (need to check further), resulting in a loss of floodplain water.

### Desired behaviour

Floodplain discharge should be added to the inflow of a reservoir or lake location.

### Additional Context

_No response_

verseve:
## Issue addressed
Fixes #367 

## Explanation
See also issue #367, fixed the inflow of reservoirs and lakes when floodplain routing as part of local inertial routing is used, the floodplain discharge was not added to the inflow.

## Checklist


hboisgon (approved):
LGTM! Many thanks for fixing this :)

verseve:
> LGTM! Many thanks for fixing this :)

Thanks for reviewing @hboisgon! I have now changed the order of the update of reservoirs and lakes so the inflow is based on current `q` instead of `q0` (discharge at previous time step). This should also improve the water balance, but it seems something is still off. I think this is caused by the different approach we use for local inertial routing compared to kinematic wave routing for inflow to a reservoir or lake. To fix this I need to do some more checking.

hboisgon:
I think unless the reservoir or lake are incorporated directly in the local inertial loop like it is in kinematic wave, there will always be some mistakes right? Maybe not for the reservoir anymore but for all the downstream cells as then the outflow of the reservoir will be given with delays to the downstream cells. 

I am happy with your first implementation, maybe when you replace or revise the LoopVectorization approach you could come back to it?

Anyway let me know if I can help with testing if you find something else in the meantime and again thanks for looking into this!

verseve:
> Anyway let me know if I can help with testing if you find something else in the meantime and again thanks for looking into this!

For the local inertial routing the `to_river`  variable of lateral subsurface flow and overland flow should also be added to the inflow of a water body, while for the kinematic wave these fluxes are added to the river for the water body location (kinematic river flow is computed for the water body cell, while for local inertial routing the water body cell is a boundary condition and the solution is replaced by the reservoir or lake module). I fixed this, and the water balance error is in the same order as river cells without a water body. Also by changing the order of the reservoir or lake update, there is no delay anymore (the reservoir or lake makes use of `q` river inflow with the correct current `dt`).

PR #373, opened 2024-03-18
Remove lower limit of zero for `sbm.runoff`

### LINKED ISSUE #372: Remove lower limit of zero for `sbm.runoff`
### Feature type

### Improvement Description

The lower limit of `runoff` of the `SBM` concept is set at zero. This was implemented in the Python version of wflow_sbm, to avoid a very slow kinematic wave solution for surface flow computation (no distinction between a kinematic wave for overland and river flow). When in the Python version of wflow_sbm the kinematic wave for surface flow was split into a river and land component, the lower limit was removed for river runoff, but not for land runoff. It would be good to also remove this lower limit for land runoff, also because it may result in water balance errors.

Remove lower limit of zero for `sbm.runoff`.

### Additional Context

_No response_

verseve:
## Issue addressed
Fixes #372 

## Explanation
See https://github.com/Deltares/Wflow.jl/issues/372.

## Checklist

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -24,8 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Close netCDF `NCDataset` with state variables in extended BMI function `save_state`.
 
 ### Changed
-- Stop exposing scalar variables through BMI. The `BMI.get_value_ptr` function was
-  not working correctly for scalar model variables (a `view` was applied). Only a few scalar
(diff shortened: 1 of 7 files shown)


JoostBuitink (changes requested):
Based on a discussion with @verseve, we decided to tackle this in a slightly different way, in order to be consistent with `runoff_river` and `net_runoff_river`. In the `runoff` variable, we do not subtract the `ea_openw_l` variable, and the `net_runoff` will become equal to `runoff - ae_openw_l`. This would require a bit more changes, to ensure that the `net_runoff` will be used throughout the code.

PR #381, opened 2024-03-26
`SBM` evaporation and use of fractions

### LINKED ISSUE #374: `SBM` evaporation and fractions (open water, vegetation, soil surface)
### Feature type

### Improvement Description

For evaporation in `SBM` we use potential evaporation with an optional correction factor `et_reftopot`. Then different processes are computed in the following order:  1) interception, 2) open water evaporation (river and land), 3) soil evaporation and 4) transpiration. The remaining potential evaporation is computed for step 1 and 2, and split based on the canopy gap fraction into remaining potential evaporation for step 3 and 4. 

An improvement is to base the potential evaporation on the different fractions (which is now done partly) of open water (`riverfrac` and `waterfrac`), vegetation and bare soil for each process, instead of working with remaining potential evaporation (bookkeeping). We can then also make use of different factors (e.g. a soil factor for bare soil evaporation, crop factor for transpiration), and keep better track of the different fractions.  

_No response_

### Additional Context

_No response_

verseve:
## Issue addressed
Fixes #374

## Explanation
See issue https://github.com/Deltares/Wflow.jl/issues/374

## Checklist

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   routing, the floodplain discharge was not added to the inflow of these locations, 2) the
   `to_river` variable from overland flow and lateral subsurface flow was not added to the
   inflow of these locations.
+- For the computation of Gash interception model parameter `e_r` multiply the precipitation
+  input with the canopy fraction (this was only done for the potential evapotranspiration
+  input).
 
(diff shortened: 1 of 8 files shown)


JoostBuitink (approved):
LGTM, thanks for fixing this! It looks like with these new fixes more water can evaporate from the soil column, rather than having most PET being "removed" by the canopy (in the old configuration). This new implementation seems more realistic!

PR #387, opened 2024-04-05
Fix `BMI.update_until` (`InexactError: Int64`)

### LINKED ISSUE #386: `BMI.update_until` can throw an `InexactError`
### Wflow version checks

### Reproducible Example

While running `BMI.update_until` the following error can occur: `InexactError: Int64`

### Current behaviour

 `InexactError`.

### Desired behaviour

No `InexactError`.

### Additional Context

_No response_

verseve:
## Issue addressed
Fixes #386 

## Explanation
`BMI.update_until` could throw an `InexactError: Int64` caused by a not whole number. This is fixed by using `divrem` for the computation of the number of  `steps` in `BMI.update_until`, an error is thrown when the absolute remainder is larger than `eps()`. An error is also thrown when `steps` is a negative number. 

## Checklist

### DIFF
--- a/src/bmi.jl
+++ b/src/bmi.jl
@@ -66,19 +66,19 @@ end
 function BMI.update_until(model::Model, time::Float64)
     @unpack clock, network, config = model
     t = BMI.get_current_time(model)
-    steps = (time - t) / model.clock.Δt.value
+    _div, _rem = divrem(time - t, model.clock.Δt.value)
+    steps = Int(_div)
     if steps < 0
         error("The current model timestamp $t is larger than provided `time` $time")
-    elseif !isinteger(steps)
+    elseif abs(_rem) > eps()
         error_message = string(
             "Provided `time` $time minus the current model timestamp $t",
             " is not an integer multiple of model time step $(model.clock.Δt.value)",
         )
         error(error_message)
-    else
-        for _ = 1:steps
-            model = run_timestep(model)
-        end
+    end
+    for _ = 1:steps
+        model = run_timestep(model)
     end
     return model
 end
--- a/src/bmi.jl
+++ b/src/bmi.jl
@@ -65,11 +65,20 @@ end
 
 function BMI.update_until(model::Model, time::Float64)
     @unpack clock, network, config = model
-    curtime = BMI.get_current_time(model)
-    steps = round((time - curtime) / model.clock.Δt.value)
-    n = Int(max(0, steps))
-    for _ = 1:n
-        model = run_timestep(model)
+    t = BMI.get_current_time(model)
+    steps = (time - t) / model.clock.Δt.value
+    if steps < 0
+        error("The current model timestamp $t is larger than provided `time` $time")
+    elseif !isinteger(steps)
+        error_message = string(
+            "Provided `time` $time minus the current model timestamp $t",
+            " is not an integer multiple of model time step $(model.clock.Δt.value)",
+        )
+        error(error_message)
+    else
+        for _ = 1:steps
+            model = run_timestep(model)
+        end
     end
     return model
 end
--- a/src/bmi.jl
+++ b/src/bmi.jl
@@ -66,7 +66,8 @@ end
 function BMI.update_until(model::Model, time::Float64)
     @unpack clock, network, config = model
     curtime = BMI.get_current_time(model)
-    n = Int(max(0, (time - curtime) / model.clock.Δt.value))
+    steps = round((time - curtime) / model.clock.Δt.value)
+    n = Int(max(0, steps))
     for _ = 1:n
         model = run_timestep(model)
     end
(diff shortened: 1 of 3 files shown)


visr (commented):
Choosing round here means that you can either undershoot or overshoot if the desired time does not fall on a model timestep. I feel like either an error or a `cld`, where there can never be undershoot, may be a better option, though I'm not sure what the BMI best practices are for this.

verseve:
Thanks for your comment @visr, good point. I think an improved error message may indeed be a better option than rounding. Will convert this PR back to draft, as it is not yet clear where the deviation from a whole number for the `time` argument is coming from (wflow or openda).

visr:
Perhaps it is almost exactly and integer, but not quite since it comes from floating point math.

```julia
julia> t = 10 * (0.1 + 0.2)
3.0000000000000004

julia> Int(t)
ERROR: InexactError: Int64(3.0000000000000004)

julia> round(t)
3.0

julia> isapprox(round(t), t)
true
```

That should probably be allowed at least, and a better error message for the not so close times would also be an improvement.

verseve:
> Perhaps it is almost exactly and integer, but not quite since it comes from floating point math.

Yes, if `time` is almost equal to an allowed model timestamp this probably happens for `steps` (almost exactly and integer). I think the issue is probably caused by the `time` input, because both current model time and model time step are always whole numbers. First a bit more testing with openda-wflow is required (timestamps are rounded in openda) to check where the problem occurs and to decide if we need to allow almost exactly and integer `steps`. Besides that, if the problem is the external `time` input, you could also argue that is the responsibility of the external software package to deal with this properly (based on the error message).

src/bmi.jl
```diff
@@ -65,10 +65,20 @@ end
...
+    t = BMI.get_current_time(model)
+    steps = (time - t) / model.clock.Δt.value
+    if steps < 0
+        error("The current model timestamp $t is larger than provided `time` $time")
+    elseif !isinteger(steps)
+        error_message = string(
+            "Provided `time` $time minus the current model timestamp $t",
+            " is not an integer multiple of model time step $(model.clock.Δt.value)",
+        )
+        error(error_message)
```
CFBaptista, line 77: I definitely agree that is a good idea to throw an error for the exact reason as stated in the error message.
However, to check whether `steps` is an integer indicates something is wrong / ill-defined somewhere else in the code.

I suggest to use `divrem` for computing `steps`.
You can use the value of the remainder for deciding whether to throw the error message.

    _div, _rem = divrem(time - curtime, model.clock.Δt.value)
    steps = Int(_div)

    if steps < 0
       # Throw error 
    elseif _rem > eps()
        # Throw error
    end

src/bmi.jl
```diff
@@ -65,10 +65,20 @@ end
...
+    else
+        for _ = 1:steps
+            model = run_timestep(model)
+        end
```
CFBaptista, line 81: `else` is never necessary.
`else` is preferable if there is high cohesion among the branches.
In this case the first two branches throw errors and the final branch advances the model in time.
Due to low cohesion I suggest to take the loop out of the conditional block.

CFBaptista (commented):
I have a few suggestions regarding

1. How to compute `steps`
2. The exact conditional statements
3. Code style

src/bmi.jl
visr: Indeed `isinteger` does not help with floating point issues like mentioned in https://github.com/Deltares/Wflow.jl/pull/387#issuecomment-2039814894. There I suggested `isapprox(round(t), t)`, which I think is a bit better since it supports small deviations on either side where `divrem` only supports overshoot.

CFBaptista: `divrem` does support both undershoot and overshoot. Depending on the inputs the outputs can be positive or negative.

visr: Just discussed with @CFBaptista that `divrem` is also good (slightly better), as long as the undershoot is also handled with negative `div`.

visr: Though even when the internal time format is the same, Float64, I think it is very common to calculate that time using floating point arithmetic. Therefore I think it's better that we handle this ourselves, it is only a few lines of code extra.

src/bmi.jl
```diff
@@ -65,10 +65,20 @@ end
 
 function BMI.update_until(model::Model, time::Float64)
     @unpack clock, network, config = model
-    curtime = BMI.get_current_time(model)
-    n = Int(max(0, (time - curtime) / model.clock.Δt.value))
-    for _ = 1:n
-        model = run_timestep(model)
+    t = BMI.get_current_time(model)
+    steps = (time - t) / model.clock.Δt.value
```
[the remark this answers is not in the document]
verseve: I agree, see commit https://github.com/Deltares/Wflow.jl/pull/387/commits/da8948cd5dca574e049d3b43b1feace77dda4a16.

src/bmi.jl
verseve: Ok, sounds good, went with `divrem` approach, see commit https://github.com/Deltares/Wflow.jl/pull/387/commits/da8948cd5dca574e049d3b43b1feace77dda4a16.

src/bmi.jl
verseve: Good point, see commit https://github.com/Deltares/Wflow.jl/pull/387/commits/da8948cd5dca574e049d3b43b1feace77dda4a16.

CFBaptista (approved):

PR #412, opened 2024-05-06
Update publications.md

verseve:
Changed wflow_sbm paper in Geosci. Model Dev. from in review to published version.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -33,12 +33,17 @@
         },
         {
             "affiliation": "Deltares",
-            "name": "Pronk, Maarten",
-            "orcid": "0000-0001-8758-3939"
+            "name": "Weerts, Albrecht",
+            "orcid": "0000-0002-3249-8363"
         },
         {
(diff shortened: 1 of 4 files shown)


visr:
Should the [Citing section in the docs](https://deltares.github.io/Wflow.jl/dev/intro/publications/#Citing-wflow) and readme also be updated? Is it needed to keep the Zenodo as well? I'd think we should direct most citations to this paper.

verseve:
> Should the [Citing section in the docs](https://deltares.github.io/Wflow.jl/dev/intro/publications/#Citing-wflow) and readme also be updated? Is it needed to keep the Zenodo as well? I'd think we should direct most citations to this paper.

Good point, committed the update in https://github.com/Deltares/Wflow.jl/pull/412/commits/a27ff3d6069d5bc74f8c5d38d87a49241e411c08. I like keep Zenodo as well, to cite a specific software version.

JoostBuitink (approved):

PR #434, opened 2024-06-26
Cleanup metadata macros

### LINKED ISSUE #432: Metadata without macros
### Feature type

### Improvement Description

The metadata of struct fields is generated with macros. While fancy, it is not very readable/explicit (code), and it is only used for model coupling through BMI and thus not exposed to a wflow user (running a wflow model simulation). Implementing metadata without macros makes the code more readable/explicit.  

_No response_

### Additional Context

_No response_

verseve:
## Issue addressed
Fixes #432 

## Explanation
See issue #432. The macro `get_units` has not been removed. This macro is nice to have and also used by `BMI` without affecting code readability. Note that this branch has been started from the `vertical_concepts` branch.


src/flow.jl
```diff
@@ -38,29 +37,44 @@ abstract type SurfaceFlow end
...
+        1
+    end
+end
+
+function grid_location(::SurfaceFlow, var)
+    if var in (:dt, :its, :kinwave_it)
+        "none"
+    else
+        "node"
+    end
```
JoostBuitink, line 77: I'm not too familiar with these functions, but it feels like they are quite specific to BMI: is that right? Is it maybe better to keep track of these in the src/bmi.jl file, or do you prefer to keep it close to their source? If it is quite specific BMI information/functionality, maybe it is better to keep those functions in the scr/bmi.jl file (a bit similar to the src/states.jl functionality). If it is not specific for BMI, keeping them here is fine with me!

JoostBuitink (changes requested):
Looks good to me, code is indeed a lot cleaner with these changes! I have one question/comment regarding the location of the `exchange` and `grid_location` functions, see the comment related to that!

src/flow.jl
verseve: Yes, these functions are quite specific to BMI. I agree and will move these functions to src/bmi.jl.

JoostBuitink (approved):
Looks good, and nice to see CI being updated for the v1 branch as well!

PR #437, opened 2024-07-05
Refactor/style guide

CFBaptista:
This PR adds configuration files for both Julia Formatter and VS Code in order to format Julia source files according to the [default style](https://docs.julialang.org/en/v1/manual/style-guide/) (with some minor deviations, see [`.JuliaFormatter.toml`](https://github.com/Deltares/Wflow.jl/blob/refactor/style-guide/.JuliaFormatter.toml)) when saving a file.
In addition the existing codebase is re-formatted according to settings in the configuration files.


.JuliaFormatter.toml
```diff
@@ -0,0 +1,15 @@
+# Options for the JuliaFormatter auto syntax formatting tool.
+# https://domluna.github.io/JuliaFormatter.jl/stable/
+# https://docs.sciml.ai/SciMLStyle/stable/
+
+# We don't use SciML style since aligning with opening brackets doesn't look good if
+# the opening bracket is near the end of the line. This squeezes the code to the right,
+# and this alignment confuses auto indent.
```
visr, line 7:
```suggestion
```

This is outdated. I just tried switching to SciML style, and this particular issue is fixed, though overall I still find in worse looking than the default style. I just removed these lines in https://github.com/Deltares/Ribasim/pull/1614 as well.

.JuliaFormatter.toml
```diff
@@ -0,0 +1,15 @@
...
+# Based on the default style we do pick these non-default options from SciML style:
+whitespace_ops_in_indices = true
+remove_extra_newlines = true
+always_for_in = true
+whitespace_typedefs = true
+
+# And add other options we like:
+separate_kwargs_with_semicolon = true
```
visr, line 15:
```suggestion
separate_kwargs_with_semicolon = true

ignore = [".pixi"]
```

This made it format the Ribasim repo on Windows in 1s instead of 7s.

I see you have julia installations in `utils/juliaup` as well, so you may need `ignore = [".pixi", "utils"]` (I did, until I got rid of that in https://github.com/Deltares/Ribasim/pull/1614. I think it'd be good for you to do the same, but can be a separate PR.)

.vscode/settings.json
visr, line 1: You may want these as well:

```json
    "julia.lint.disabledDirs": [
        ".pixi"
    ],
    "julia.lint.run": true
```

At least until https://github.com/julia-vscode/julia-vscode/pull/3611 is merged.

visr (approved):
Left some suggestions, but looks good.

verseve (approved):
LGTM!

.JuliaFormatter.toml
CFBaptista: Thanks, that is good to know.

In the past I have used Blue without any modifications.
For my private projects I prefer Blue over SciML because I think it looks better.
You might want to give it a go and see whether aesthetically it is a better fit for us.
It will for sure have my vote.

.JuliaFormatter.toml
CFBaptista: How did you measure the formatting time?
I get less than 0.7 seconds doing this (without any of the pixi-related suggestions):

    julia> using BenchmarkTools

    julia> using JuliaFormatter

    julia> @btime format(".")
      672.831 ms (2384350 allocations: 143.38 MiB)

.JuliaFormatter.toml
visr: I agree with Blue over SciML, but this is default, not SciML. What do you like better about adding `style = "blue"`? I just tried it out, the differences are not large, but most differences I dislike:

- Blue removes spaces around = in kwargs
- Blue tries harder to put more args on the same line

Both make Blue more compact but less readable IMO. The default style is also more in line with Python's black formatter.

.JuliaFormatter.toml
visr: But is it faster with the ignore? I just did `@time format(".")`. There are actually two reasons for ignoring:

- We don't want to format any .jl files in there
- We don't want to spend time checking for tens of thousands of files if they are julia files or not

.JuliaFormatter.toml
verseve: In addition we were already using the default style.

.JuliaFormatter.toml
CFBaptista: Be aware that by using `@time` the timing might include warm-up time (something you avoid with `BenchmarkTools.@btime`.
You might want to run `@time` twice to see the actual timing.

The ignore had no impact on my machine.

.JuliaFormatter.toml
visr: Indeed, [BlueStyle](https://github.com/JuliaDiff/BlueStyle) definitely goes beyond the [Julia Style Guide](https://docs.julialang.org/en/v1/manual/style-guide/). I was only really thinking about the JuliaFormatter setting.

.JuliaFormatter.toml
visr: Yeah I'm aware. My timings were for Ribasim, where the pixi folder is much bigger, hence the effect more pronounced. I still see an improvement from 0.6s to 0.4s on my machine, which makes sense.

.vscode/settings.json
CFBaptista: When adding this to `wflow`'s `settings.toml` the configuration is grayed out and the warning in the screenshot is shown. Do you know what is going on here?

image

visr: Not sure, I don't see the same in Ribasim. Perhaps this is related to the way you open a workspace, e.g. with just Wflow, or a combined workspace.

PR #438, opened 2024-07-08
Stop using local JULIAUP_DEPOT_PATH

visr:
Same as https://github.com/Deltares/Ribasim/pull/1614


verseve (approved):
Thanks for this PR! 
We should not forget to merge these changes also with the `v1` branch.

PR #440, opened 2024-07-18
Stop using local JULIAUP_DEPOT_PATH (#438)

verseve:
Cherry-pick from Master: Stop using local JULIAUP_DEPOT_PATH

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,5 +24,3 @@
 
 # pixi environments
 .pixi
-# juliaup
-utils/juliaup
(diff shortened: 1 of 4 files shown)


visr:
Why are you using a v1 branch? Are you already committing things to master that you don't want in v1?

verseve:
> Why are you using a v1 branch? Are you already committing things to master that you don't want in v1?

Yeah, correct, we decided to work independently of the master branch. Mostly because of bug fixing or new features that are required in the master branch but cannot be (yet) included in the v1 branch.

Issue #382, opened 2024-03-28
Refactor modeltype `sediment`


verseve:
### Feature type

Refactor the sediment model, i.e., re-organize the model's code without adding new or removing existing features.

### Improvement Description

The initialization of the sediment model takes place in one function (`initialize_sediment_model`) of over 140 lines of code.
This is poorly maintainable and should be split up into smaller individual logical steps.
In addition, the parameters used for initializing the sediment model are, currently, stored alongside the parameters of all other models / concepts in wflow.
Sediment model parameters should be extracted into its own parameter `Struct`.
The extraction of the sediment model parameters should include the re-coupling of sediment model parameter to other models that require access to sediment model parameters.

see also main issue https://github.com/Deltares/Wflow.jl/issues/351

### Tasks

- [ ] #425 
- [ ] #451  
- [ ] #453  
- [ ] Store sediment model parameters in a separate `Struct`
- [ ] Refactor the sediment model to use its own parameter `Struct` instead of the `Model` struct
- [ ] Remove sediment model parameters from the `Model` struct
- [ ] Fix all sediment model parameter access invalidations

hboisgon:
Looking through the Model, I see the following possibilities for new ``Struct``.

The main points (still to decide) are:

- Move some of the parameter calculations to HydroMT: 
     - soil mean sediment diameter d50
     - Govers parameters: nGovers, cGovers 
     - soil fractions of clay/silt/sand/small and large aggregates
     - Kodatie parameter
     - Bed and bank erodibility based on critical shear stress: kdbed, kdbank
- Combine river and overland transport capacities in the same structs (they all depend on flow velocity be it shallow overland flow or more concentrated river flow)
- Create a separate ``Struct`` for waterbodies (no real distinction here between lake and reservoir)? Can the one for SBM be re-used?
- Create a SedimentParticle ``Struct`` containing at least the sediment properties for each class. Should this class also contains the different variables and states of the model (ie soil erosion, soil particles entering the river, river erosion, river deposition, sediment flux in suspended/bed load etc...)?
- Organisation of the RiverSediment struct


## Sediment model: potential new structs

### SoilErosion
Existing model can compute total erosion or sum of erosion for different sources (overland flow, rainfall, wind etc.). I guess we then need a main SoilErosion struct that could switch between summing erosion from different sources.
So far in wflow, we support erosion from overland flow (1 concept) and from rainfall (2 concepts).
[...]

PR #446, opened 2024-08-15
Refactor/init sediment model

CFBaptista:
This PR refactors the initialization of the sediment model by:

- Removing unused variables
- Removing stubs
- Renaming variables names to more descriptive ones
- Extracting large constructions to a separate function
- Simplifying some logic

### DIFF
--- a/src/sediment_model.jl
+++ b/src/sediment_model.jl
@@ -158,8 +158,8 @@ function initialize_sediment_model(config::Config)
     return model
 end
 
-function init_overland_flow_sediment(river, number_of_cells)
-    river_cell = float(river)
+function init_overland_flow_sediment(river::Vector, number_of_cells::Integer)
+    river_cell = convert(Vector{Float}, river)
 
     overland_flow_sediment = OverlandFlowSediment{Float}(;
         n = number_of_cells,
--- a/src/sediment_model.jl
+++ b/src/sediment_model.jl
@@ -139,6 +139,8 @@ function initialize_sediment_model(config::Config)
         reverse_indices = reverse_indices_river,
     )
 
+    sediment_model = SedimentModel()
+
     model = Model(
         config,
         (; land, river, reservoir = nothing, lake = nothing, index_river, frac_to_river),
@@ -147,7 +149,7 @@ function initialize_sediment_model(config::Config)
         clock,
         reader,
         writer,
-        SedimentModel(),
+        sediment_model,
     )
 
     model = set_states(model)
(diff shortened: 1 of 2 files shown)


verseve:
A general comment (did not go though the code in detail in yet): what about refactoring of the different structs that are part of the Sediment model? As discussed for the SBM concept we would like to sub-divide long structs in sub-structs, I think this is also applicable for the Sediment concept. Probably good to also ask @hboisgon to review this PR?

CFBaptista:
> A general comment (did not go though the code in detail in yet): what about refactoring of the different structs that are part of the Sediment model? As discussed for the SBM concept we would like to sub-divide long structs in sub-structs, I think this is also applicable for the Sediment concept. Probably good to also ask @hboisgon to review this PR?

I completely agree and this is also what I intend to do.
I, however, prefer to implement the changes across multiple PR's instead of one big one.
Please refer to the task list in #382 for an overview.
Each task will be a separate PR.
If you feel that list is incomplete please let me know.
You can also add potentially missing steps yourself if you would like to.

src/sediment_model.jl
```diff
@@ -13,138 +13,143 @@ function initialize_sediment_model(config::Config)
...
-        order = topological_sort_by_dfs(graph_riv),
-        indices = inds_riv,
-        reverse_indices = rev_inds_riv,
+        graph = graph_river,
+        order = topological_sort_by_dfs(graph_river),
+        indices = indices_river,
+        reverse_indices = reverse_indices_river,
     )
 
+    sediment_model = SedimentModel()
```
verseve, line 142: What is the advantage of doing this? Why not use directly `SedimentModel()`?

src/sediment_model.jl
```diff
@@ -153,6 +158,41 @@ function initialize_sediment_model(config::Config)
     return model
 end
 
+function init_overland_flow_sediment(river::Vector, number_of_cells::Integer)
+    river_cell = convert(Vector{Float}, river)
+
+    overland_flow_sediment = OverlandFlowSediment{Float}(;
```
verseve, line 164: Instead of `Float` you could also use `T` here, equal to the `SBM` model concept.

verseve (changes requested):
LGTM, only two small comments. 

As suggested, would be good if @hboisgon also reviews this PR.

src/sediment_model.jl
CFBaptista: It facilitates debugging.
Having the construction of an object as a separate statement instead of nesting its construction in the construction of another object add additional flexibility for placing breakpoints.
For the same reasoning I also refrain from constructing objects during a return statement.

src/sediment_model.jl
CFBaptista: I added an argument to the function call to generalize the construction of `OverlandModelSediment`.
It can now construct OverlandModelSediment with type parameters other than `Float` as long as it is a subtype of `AbstractFloat`.

PR #448, opened 2024-08-15
Move docs to quarto for v1

### LINKED ISSUE #224: Further improve documentation with examples/tutorials and move to Quarto
Further improve the documentation. In the current docs, the examples and explanations are somewhat combined, making it difficult for a (new) user to figure out best practices to work with Wflow. It would be nice to separate the examples from the explanations. Additionally, adding some tutorials and/or clear examples would really benefit the usability as well (see [hydromt_wflow docs](https://deltares.github.io/hydromt_wflow/latest/) for a great example).

This would require restructuring of the docs, including new sections. The Documenter.jl package (currently) only supports a menu bar on the left hand side, without different sections such as the hydromt_wflow docs. Adding more sections would make navigating even more difficult (as we are already reaching approaching the limit). Shifting to sphinx would be a solution, but we would lose the direct Julia integration of Documenter.jl: need to figure out the consequences if we want to switch to a different documentation generator.

JoostBuitink:
## Issue addressed
Fixes #224 

## Explanation
Moved all current docs to quarto format

Docs can be viewed via this link: https://deltares.github.io/Wflow.jl/quarto/

Some items are still missing/need to be picked up
- Many internal links to other pages are not yet working
- The version switcher needs to be updated (check Delft-FIAT repo)
- Examples are missing, but can be added in the future


JoostBuitink:
Please ignore all changes in the `src/` directory (I need to fix this still). These are inherited from (mainly) the recent water-demand updates. I included those to ensure all docs we aligned with the master branch when moving to quarto

PR #470, opened 2024-09-26
Reinfiltration of surface water and flow threshold for overland flow

### LINKED ISSUE #237: Re-infiltration of surface water
Re-infiltration of surface water can be an important process, for example in dry areas or for the simulating the effect of nature-based solutions (e.g. changing the roughness parameter for overland flow can have an affect on re-infiltration of surface water).

JoostBuitink:
## Issue addressed
Fixes #237 

## Explanation

## Checklist


verseve:
Nice work @JoostBuitink ! 

I will submit a review this Friday. One question I did already have: you did check the water balance of the vertical part (runoff), did you also check the water balance of the overland flow? I would recommend this as the code has changed quite a bit (I think mainly the kinematic wave).

verseve (changes requested):
Nice additions! Besides checking the water balance, it seems to me that the update of the `pond_volume_current` variable in the kinematic wave overland flow iterations is not completely correct (see also two comments added to this part of the code).

src/flow.jl
```diff
@@ -227,23 +240,47 @@ function update(sf::SurfaceFlowLand, network, frac_toriver)
...
+                        (sf.qin[v] * dt),
+                        0.0,
                     )
 
-                    # update h, only if surface width > 0.0
-                    if sf.width[v] > 0.0
-                        crossarea = sf.alpha[v] * pow(sf.q[v], sf.beta)
-                        sf.h[v] = crossarea / sf.width[v]
+                    # Start kinematic wave if pond volume exceeds threshold
+                    if pond_volume_pot >= threshold_volume[v]
```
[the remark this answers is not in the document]
verseve: Just a quick thought: an alternative is to pull the pond out of the kinematic wave solution, and only consider vertical fluxes, similar to the `Paddy` approach.

src/flow.jl
```diff
@@ -199,6 +206,12 @@ function update(sf::SurfaceFlowLand, network, frac_toriver)
     @. sf.alpha = sf.alpha_term * pow(sf.width, sf.alpha_pow)
     @. sf.qlat = sf.inwater / sf.dl
 
+    # Convert threshold to volume
+    threshold_volume = sf.h_thresh .* sf.width .* sf.dl
+
+    # Get pond volume at the start of the time step
+    pond_volume_current = sf.pond_height .* sf.width .* sf.dl
```
verseve, line 213: This will result in more allocation and could have an impact on performance. Would be nice if this can be avoided (in the for loop?).

src/flow.jl
```diff
@@ -207,10 +207,10 @@ function update(sf::SurfaceFlowLand, network, frac_toriver)
     @. sf.qlat = sf.inwater / sf.dl
 
     # Convert threshold to volume
-    threshold_volume = sf.h_thresh .* sf.width .* sf.dl
+    threshold_volume = sf.h_thresh .* area
```
verseve, line 210: Why the use of `area` here? It's confusing as it deviates from the overland flow dimensions `width` and `dl`. For river cells `area` is higher than `sf.width .* sf.dl`, for land cells it has the same value. I would suggest to use the overland flow dimensions for the computation of the `threshold_volume`, and the `pond_volume_current`. 

What most probably is not correct yet when using the overland flow dimensions, is the amount of surface water that can infiltrate from river cells. I think this amount should be scaled by the land fraction of these cells. We have a similar approach for evaporation and direct runoff by using `waterfrac`.

src/sbm.jl
```diff
@@ -916,6 +921,14 @@ function update_until_recharge(sbm::SBM, config)
             sbm.waterfrac[i] * sbm.potential_evaporation[i],
         )
 
+        # Update land waterlevel
+        waterlevel_land = sbm.waterlevel_land[i] - ae_openw_l
+
+        if do_surface_water_infiltration
+            # Add land waterlevel to infiltration
+            avail_forinfilt += waterlevel_land
```
verseve, line 929: The amount of surface water that can infiltrate needs to be scaled by the land fraction. We use a similar approach for runoff and evaporation from open water (using `waterfrac`).

src/sbm.jl
```diff
@@ -916,26 +919,31 @@ function update_until_recharge(sbm::SBM, config)
...
-        ae_openw_l = min(
-            sbm.waterlevel_land[i] * sbm.waterfrac[i],
-            sbm.waterfrac[i] * sbm.potential_evaporation[i],
+        # Determine remaining fraction of available potential_evaporation
+        soilevap_fraction = max(
+            sbm.canopygapfraction[i] - sbm.riverfrac[i] - sbm.glacierfrac[i],
+            0.0,
         )
+        # Determine amount of surface water
+        surface_water = sbm.waterlevel_land[i] * (1.0 - sbm.riverfrac[i])
```
verseve, line 928: Would recommend to use the same approach for `runoff_land` (L. 912) for consistency. Here `waterfrac` is still used.

verseve: This implementation (not using `waterfrac`) results in a hydrograph that is quite flashy (tested by @JoostBuitink ). Most likely this is caused by a disconnect between the overland flow kinematic wave and soil: water is available in the kinematic wave while the soil (unsaturated zone) has also capacity left (`ustorecapacitiy`), especially in the case when infiltration of surface water is not allowed.

So, while we model the kinematic wave routing for the complete land fraction, and allow for open water evaporation and surface runoff for that fraction (by not using `waterfrac`), most likely the soil is not completely saturated. As a consequence, this results most likely in an "overestimation" of surface runoff (`runoff_land`) and open water evaporation.

Based on this, my recommendation would be to:
1) Revert back to using `waterfrac`, the approach is then consistent between surface runoff and open water evaporation, or
2) Use `waterfrac` unless the soil in the grid cell is almost completely saturated (`ustorecapacitiy` $\approx$ 0.0), then use the land fraction.

Another approach (for the longterm) and similar to 2) could be to use subgrid variability, for example by using a pdf of soil storage capacity or of topographic information to derive a dynamic saturated fraction (e.g. see also this [paper](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2007JD008940)).

PR #471, opened 2024-10-01
Update julia to 1.10.5, update manifests

visr:


verseve (approved):
Thanks for this update, LGTM!

PR #472, opened 2024-10-01
Use FastPower.jl for faster power function

visr:
This package was registered today: https://github.com/SciML/FastPower.jl
I remember from profiling a while ago that the power function was the main bottleneck for the kinematic wave.  So I wonder how much difference this trade of precision for performance will be. This function by itself is about 2-2.5x faster.


visr:
Wow I'm surprised, but this actually leads to significantly worse performance. It looks like with the lower power precision the kinematic wave doesn't convergence well. The epsilon of 1e-12 may just be too low, making it hit the max_iters instead.

https://github.com/Deltares/Wflow.jl/blob/8a4d2e936f2f62bdb80efc7f3a0d9acba340ac95/src/horizontal_process.jl#L27-L29

If I set the epsilon higher, the performance difference goes down.

visr:
Closing this since it won't get merged like this. We should IMO still see if we can start using FastPower.jl, but the convergence criteria should be set differently for it to pay off.

PR #480, opened 2024-10-09
Remove git-tree-sha1 from metadata

### LINKED ISSUE #479: Build wflow_cli fails
### Wflow version checks

### Reproducible Example

Running Pixi task `build-wflow-cli` throws an ERROR.

### Current behaviour

```05:12:51   [ Info: PackageCompiler: Done
05:12:51   - PackageCompiler: compiling nonincremental system image
05:15:30   ERROR: LoadError: KeyError: key "git-tree-sha1" not found```

### Desired behaviour

Do not add metadata "git-tree-sha1" for Wflow.

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by verseve**
It is not part of the Wflow entry in the Manifest.toml file and as a result the build fails.

## Issue addressed
Fixes #479

### DIFF
--- a/build/create_binaries/add_metadata.jl
+++ b/build/create_binaries/add_metadata.jl
@@ -43,7 +43,19 @@ function add_metadata(project_dir, license_file, output_dir, git_repo)
         version = TOML.parsefile(normpath(git_repo, "Project.toml"))["version"]
         repo = GitRepo(git_repo)
         branch = LibGit2.head(repo)
+        commit = LibGit2.peel(LibGit2.GitCommit, branch)
         short_name = LibGit2.shortname(branch)
+        short_commit = string(LibGit2.GitShortHash(LibGit2.GitHash(commit), 10))
+
+        # get the release from the current tag, like `git describe --tags`
+        # if it is a commit after a tag, it will be <tag>-g<short-commit>
+        options = LibGit2.DescribeOptions(; describe_strategy = LibGit2.Consts.DESCRIBE_TAGS)
+        result = LibGit2.GitDescribeResult(repo; options)
+        suffix = "-dirty"
+        foptions =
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by visr** [comment]
If we don't have the git-tree-sha1 or the commit, we only have the branch and the version number, which is not enough to identify the source of the build. This started failing since the build script moved into this repository, hence it is not in the metadata. It's no problem to remove the git-tree-sha1 but then I'd at least add the commit.

For Ribasim we add this to the README.md in the root of our build:

```toml
release = "2024.10.0-128-ge09eeb5"
commit = "https://github.com/Deltares/Ribasim/tree/e09eeb5b9a"
branch = "https://github.com/Deltares/Ribasim/tree/main"
julia_version = "1.10.5"
core_version = "2024.11.0"
```

With additionally in `share/julia/Build.toml` the TeamCity build number, in case it was built on TeamCity:

```toml
BUILD_NUMBER = "1342"
BUILD_VCS_NUMBER = "e09eeb5b9ad46b9715a3aa5ba47fc69b03b48789"
```

You can take what you want from https://github.com/Deltares/Ribasim/blob/4048a50367dbc4f28975b28f6c6da76cb77e3ea5/build/build.jl#L92

verseve:
> If we don't have the git-tree-sha1 or the commit, we only have the branch and the version number, which is not enough to identify the source of the build. This started failing since the build script moved into this repository, hence it is not in the metadata. It's no problem to remove the git-tree-sha1 but then I'd at least add the commit.

Thanks for the info. I think this started failing with the latest update of the Manifest.toml file: https://github.com/Deltares/Wflow.jl/commit/59f5ef6653ca0952dfdde03978781455285f2029#diff-4d5f121308e3da7be8f85c55dba1429a0a14f54b42c9d40eedf2d6effdcf2446L655.

Adding the commit (and aligning with Ribasim) sounds good!

visr (approved):
Looks good! Assuming you tested locally.

PR #488, opened 2024-10-22
Quality of life improvements

SouthEndMusic:
What better way to get to know a new repository than to hack around in it :-) In this branch I did 2 things: 

- Removing the `@unpack` macro: this macro (and the `UnPack` package) is no longer needed because of the named tuple notation for obtaining fields from objects. An advantage of this notation is that VSCode indicates whether the field name is actually used in the given scope, which allowed me to clean up the code a bit
- Only import used names: This prevents that the name space gets cluttered up and shows in a nice overview in `Wflow.jl` what is actually used from the dependencies

~I accidentally also upgraded the Julia version to 1.11, if you don't want this let me know.~


visr (approved):
Looks good! The explicit imports (can be done using https://github.com/ericphanson/ExplicitImports.jl) help making it clear where functions come from. And I also prefer the NamedTuple unpacking over the unpack macro. The NamedTuple unpacking didn't exist when this code was first written.

PR #490, opened 2024-10-28
Clean up equations and units in docs

SouthEndMusic:
I am a sucker for neat mathematical notation (: One of my first PRs to the Ribasim repository was also like this.

These are most of the types of changes I made:
- Unified method for displaying units, without mixing $\TeX$ and non- $\TeX$
- Variable names and subscripts that consist of multiple letters are non-italic to indicate that they form one 'word' (I am not a fan of variable names that consist of multiple letters but I understand that it is unavoidable sometimes)
- Use proper symbols for partial derivatives

Some things still need to be fixed, I left some comments in a few places (most notably my question about what the unit `t` means). 

I found [DocumenterCitations.jl](https://juliadocs.org/DocumenterCitations.jl/stable/) for a nicer incorporation of references, but given the plan to migrate the docs (https://github.com/Deltares/Wflow.jl/pull/448) that might not be worth looking into.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 # Document generation
 /docs/build/
 /docs/site/
+/docs/Manifest.toml
 
 # Temporary files
 /dev/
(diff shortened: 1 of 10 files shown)


SouthEndMusic:
Merged into https://github.com/Deltares/Wflow.jl/pull/448.

PR #493, opened 2024-11-06
Update changelog

verseve:


hboisgon (approved):
LGTM! I wonder if a deprecated section would not make it even clearer to users instead of list in this list but I do not know if it follows semantic versioning conventions

PR #501, opened 2024-11-13
Fix docker building

JoostBuitink:
## Issue addressed
No issue, docker was failing to build on teamcity, due to missing test data (as a result from moving to pixi tasks).

## Explanation
Run script to download testdata in the docker file (and remove when done, as these files are not needed inside the docker).

## Checklist


evetion (approved):
If it works, it works. Note that in the future, you might be able to create a slightly smaller image by combining all Julia steps in one, just like the apt-get step. Every `RUN` creates it's own layer in the Docker image.

PR #513, opened 2024-12-03
Wflow ZMQ server: allow JSON reading and writing of `NaN` and `Inf`

### LINKED ISSUE #392: Wflow ZMQ Server: `BMI.getvalue` can throw JSON spec error
### Wflow version checks

### Reproducible Example

Requesting wflow variables with the default `ksat_profile` setting through the wflow ZMQ Server. 

### Current behaviour

The following error is thrown:

┌ Error: Wflow function `get_value` failed
│   exception =
│    NaN not allowed to be written in JSON spec
│    Stacktrace:
│      [1] error(s::String)
│        @ Base .\error.jl:35
│      [2] #write#90

### Desired behaviour

In this case the model variables `vertical.z_layered` and `lateral.subsurface.kh` are set at NaN during initialization of a wflow model with the default `ksat_profile` setting. Preferably we do not allow to exchange NaN with JSON because it is not standard.

### Additional Context

_No response_

vers-w:
## Issue addressed
Fixes #392 

## Explanation
See issue #392


JoostBuitink (approved):
Looks good to me (note: did not test the code myself, but the implementation looks straightforward and good)!

PR #514, opened 2024-12-06
Update Julia and Pixi manifests

visr:
The main benefit is that the Julia update from 1.10.5 to 1.10.7 will fix random failures on TeamCity related to artifact installation filesystem issues. See https://discourse.julialang.org/t/error-ioerror-rm-c-users-35879-julia-artifacts-jl-wek1vy-bin-directory-not-empty-enotempty/123142/5?u=visr if you are interested in details.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -11,10 +11,11 @@ platforms = ["win-64", "linux-64"]
 
 [tasks]
 # Installation
-install-julia = "juliaup add 1.10.5 && juliaup override unset && juliaup override set 1.10.5"
+install-julia = "juliaup add 1.10.7 && juliaup override set 1.10.7"
 # Julia
-update-registry-julia = "julia --eval='using Pkg; Registry.update()'"
-instantiate-julia = {cmd = "julia --project --eval='using Pkg; Pkg.instantiate()'",env = { JULIA_SSL_CA_ROOTS_PATH = "" } }
+# Clear SSL_CERT_DIR to avoid Julia warnings, see https://github.com/JuliaLang/Downloads.jl/issues/244
+update-registry-julia = { cmd = "julia --eval='using Pkg; Registry.update()'", env = { SSL_CERT_DIR = "" } }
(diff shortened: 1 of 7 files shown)


pixi.toml
```diff
@@ -11,10 +11,11 @@ platforms = ["win-64", "linux-64"]
 
 [tasks]
 # Installation
-install-julia = "juliaup add 1.10.5 && juliaup override unset && juliaup override set 1.10.5"
+install-julia = "juliaup add 1.10.7 && juliaup override set 1.10.7"
```
visr, line 14: This can be simplified due to https://github.com/JuliaLang/juliaup/pull/1002. Pixi will pull in the latest juliaup with this patch.

pixi.toml
```diff
@@ -11,10 +11,11 @@ platforms = ["win-64", "linux-64"]
...
 # Julia
-update-registry-julia = "julia --eval='using Pkg; Registry.update()'"
-instantiate-julia = {cmd = "julia --project --eval='using Pkg; Pkg.instantiate()'",env = { JULIA_SSL_CA_ROOTS_PATH = "" } }
+# Clear SSL_CERT_DIR to avoid Julia warnings, see https://github.com/JuliaLang/Downloads.jl/issues/244
+update-registry-julia = { cmd = "julia --eval='using Pkg; Registry.update()'", env = { SSL_CERT_DIR = "" } }
+instantiate-julia = {cmd = "julia --project --eval='using Pkg; Pkg.instantiate()'",env = { SSL_CERT_DIR = "" } }
```
visr, line 18: These will be less noisy, see https://github.com/JuliaLang/Downloads.jl/issues/244

visr (commented):
With Julia 1.12 we can use workspaces such that we only have a single Manifest.toml, that will be nicer to handle: https://github.com/JuliaLang/Pkg.jl/blob/master/CHANGELOG.md#pkg-v112-release-notes

SouthEndMusic (approved):

PR #518, opened 2024-12-10
Mapping of internal model variables and parameters to standard names

### LINKED ISSUE #481: Mapping of internal model variables and parameters to standard names
### Feature type

### Improvement Description

The mapping of model variables to external names (e.g. in the TOM config file and BMI) is one to one. For example, the `satwaterdepth` variable of the vertical SBM concept can be accessed in Julia through `model.vertical.satwaterdepth`, and should be mapped to `vertical.satwaterdepth` in the TOML config file to write this variable as output or read as a state input. As part of v1.0, the large structs of vertical concepts are divided into sub-models/structs. For example the SBM concept is divided into interception, snow, glacier and soil modules. Additionally, separate structs are used for model variables and parameters, and we make use of shared parameters (e.g. for vegetation).
 
As a consequence, this results in a mapping that can get complex for an end-user, with below an example for a part of the TOML config file:

```toml
[output.vertical.interception.variables]
canopy_storage = "canopystorage"

[output.vertical.soil.variables]
satwaterdepth = "satwaterdepth"
tsoil = "tsoil"
ustorelayerdepth = "ustorelayerdepth"

[output.vertical.snow.variables]
snow_storage = "snow"
snow_water = "snowwater"
```

Additionally, a disadvantage of the one to one mapping is that any change in the setup of the Model struct requires also a change in the TOML config file. An improvement would be to work with standard variable/parameter names resulting in a more user friendly external access of model variables.

For the variable names CSDMS provides a good standard that we can use: https://csdms.colorado.edu/wiki/CSDMS_Standard_Names.

Below an example of (part of) the TOML file using external names from the CSDMS standard:

```toml

[input.static]
soil_water__saturated_volume_fraction = "thetaS"
soil_water__residual_volume_fraction = "thetaR"

vegetation_woody-part__storage_capacity = "Swood"
vegetation__leaf-area_index = "LAI"
vegetation_canopy__light-extinction_coefficient = "Kext"

snowpack__degree-day_coefficient = "Cfmax"

channel_water_flow__manning_n_parameter = "N"

land_surface__slope = "Slope"

[output.variables]
soil_water_sat-zone_top__depth = "zi"
soil_water__infiltration_volume_flux = "actinfilt"
soil_surface__temperature = "tsoil"
soil_water_sat-zone__depth = "satwaterdepth"
soil_water_unsat-zone__depth-per-soil_layer = "ustorelayerdepth"

vegetation_canopy_water__storage = "canopystorage"

channel_water__volume_flow_rate = "q_river"
channel_water__time_average_of_volume_flow_rate = "q_av_river"

subsurface_water__volume_flow_rate = "ssf"
```

### Additional Context

_No response_

vers-w:
## Issue addressed
Fixes #481

- [x] model type `sbm`: change model parameters, states, and output variables (sub-selection) to standard names.
- [x] model type`sbm_gwf`: change model parameters, states, and output variables (sub-selection) to standard names.
- [x] model type `sediment`: change model parameters, states, and output variables (sub-selection) to standard names.
- [x] update/refactor check of state variables.
- [x] possibly refactor hydrological forcing in model type `sediment`, see also https://github.com/Deltares/Wflow.jl/pull/518#issuecomment-2551290932, for now we decided to go with improvement in commit https://github.com/Deltares/Wflow.jl/pull/518/commits/8b38fb2979b205d06ba490c4711d255aadf6de16. 
- [x] check list model variables allowed to exchange (`standard_name_map`).
- [x] check units `sediment` model, not always clear from code/metadata/docs as the unit is also part of the standard name (quantity).
- [x] reduce allocations (increased compared to base branch). the increase in allocations is introduced in commit https://github.com/Deltares/Wflow.jl/commit/fbc238325dbe50fbe994aacfd43a90848c66b123, most probably caused by a new version of `NCDatasets` (need to look further into this, not part of this PR).


SouthEndMusic:
@vers-w I see you got rid of the nested names with periods in your TOML headers, but I make use of this structure in https://github.com/Deltares/Wflow.jl/pull/519. I know I just started working on this without discussing with you, so let me know what your thoughts are (:

JoostBuitink:
> @vers-w I see you got rid of the nested names with periods in your TOML headers, but I make use of this structure in #519. I know I just started working on this without discussing with you, so let me know what your thoughts are (:

Just to jump in this discussion here: we indeed decided to reorganize the toml file, so perhaps better to wait with #519 to prevent double work ;) ).

vers-w:
Hydrological forcing data in model type `sediment` is stored as part of the `SoilLoss`, `OverlandFlowSediment` and `RiverSediment` components, independently. In commit https://github.com/Deltares/Wflow.jl/pull/518/commits/8b38fb2979b205d06ba490c4711d255aadf6de16 this data is now shared (in memory) between the `SoilLoss` and `OverlandFlowSediment` components. The same hydrological forcing time series are also present in the `sbm` and `sbm_gwf` model types stored at different internal model locations, compared to the `sediment` model. Because of this two different mappings (between the standard name and internal model name) are required.

Possible improvements:
1. Store hydrological forcing data at one location in the `sediment` model (two mappings are required).
2. Share the hydrological forcing data for  `SoilLoss` and `OverlandFlowSediment` components (commit https://github.com/Deltares/Wflow.jl/pull/518/commits/8b38fb2979b205d06ba490c4711d255aadf6de16, two mappings are required).
3. Store the hydrological forcing data at the same internal model locations as the `sbm` and `sbm_gwf` models. One mapping is required and is likely most future-proof (e.g. run `sbm` and `sediment` in a combined wflow simulation).

vers-w:
@JoostBuitink and @hboisgon, this PR is ready for review, couple of comments from my side:
* For the standard names I did follow as much as possible the [CSDMS standard names](https://csdms.colorado.edu/wiki/CSDMS_Standard_Names), also the [examples](https://csdms.colorado.edu/wiki/CSN_Examples) are quite useful.
* It is not possible to output model parameters and some variables through the standard names. It is possible to output these through the previous implementation (with dots) if needed. I suggest we consider only the standard names as the (official) API and not the previous implementation, otherwise we will have a breaking change (or make it backward compatible) each time when we change the internal structure. Please also check if variables are missing in the standard_name.jl file.
* As I am not that familiar with the sediment code it would be good if @hboisgon could focus on checking the standard names of this component (besides checking generally if the standard names of the other components make sense). Also, the units of the sediment variables are not always that clear to me (sometimes there are differences between code, metadata and docs).
* For a variable with dimension `:layer`, the standard name starts with `soil_layer` to indicate that this variable can represent multiple soil layers (allowed as input and output).
* The standard name that starts with `land_surface_water` can refer to overland flow (for example `land_surface_water__volume_flow_rate`) or total surface water on the land surface including rivers, reservoirs and lakes (`land_surface_water_abstraction__volume_flux` and `land_surface_water__potential_evaporation_volume_flux`).

src/standard_name.jl
```diff
@@ -0,0 +1,208 @@
...
+    "soil_erosion__mass_flow_rate" => @optic(_.land.soil_erosion.variables.amount),
+    "soil_erosion_clay__mass_flow_rate" => @optic(_.land.soil_erosion.variables.clay),
+    "land_surface_water_sediment_transport_capacity__mass_flow_rate" =>
+        @optic(_.routing.overland_flow.transport_capacity.variables.amount),
+    "land_surface_water_clay_transport_capacity__mass_flow_rate" =>
+        @optic(_.routing.overland_flow.transport_capacity.variables.clay),
+    "land_surface_water_sediment~to-river__mass_flow_rate" =>
+        @optic(_.routing.overland_flow.to_river.variables.amount),
+    "land_surface_water_clay~to-river__mass_flow_rate" =>
+        @optic(_.routing.overland_flow.to_river.variables.clay),
```
hboisgon, line 154: This one is needed for all particle class clay/silt/sand/sagg/lagg
```suggestion
    "land_surface_water_clay~to-river__mass_flow_rate" =>
        @optic(_.routing.overland_flow.to_river.variables.clay),
    "land_surface_water_silt~to-river__mass_flow_rate" =>
        @optic(_.routing.overland_flow.to_river.variables.silt),
    "land_surface_water_sand~to-river__mass_flow_rate" =>
        @optic(_.routing.overland_flow.to_river.variables.sand),
    "land_surface_water_aggregates~small~to-river__mass_flow_rate" =>
        @optic(_.routing.overland_flow.to_river.variables.sagg),
    "land_surface_water_aggregates~large~to-river__mass_flow_rate" =>
        @optic(_.routing.overland_flow.to_river.variables.lagg),
```

src/standard_name.jl
```diff
@@ -0,0 +1,208 @@
...
+        @optic(_.land.hydrological_forcing.q_land),
+    "land_surface_water__depth" => @optic(_.land.hydrological_forcing.waterlevel_land),
+    "vegetation_canopy_water__interception_volume_flux" =>
+        @optic(_.land.hydrological_forcing.interception),
+    "soil_erosion~rainfall__mass_flow_rate" =>
+        @optic(_.land.rainfall_erosion.variables.amount),
+    "soil_erosion~overland_flow__mass_flow_rate" =>
+        @optic(_.land.overland_flow_erosion.variables.amount),
+    "soil_erosion__mass_flow_rate" => @optic(_.land.soil_erosion.variables.amount),
+    "soil_erosion_clay__mass_flow_rate" => @optic(_.land.soil_erosion.variables.clay),
```
hboisgon, line 146: This one is needed for all particle class clay/silt/sand/sagg/lagg
```suggestion
    "soil_erosion_clay__mass_flow_rate" => @optic(_.land.soil_erosion.variables.clay),
    "soil_erosion_silt__mass_flow_rate" => @optic(_.land.soil_erosion.variables.silt),
    "soil_erosion_sand__mass_flow_rate" => @optic(_.land.soil_erosion.variables.sand),
    "soil_erosion_aggregates~small__mass_flow_rate" => @optic(_.land.soil_erosion.variables.sagg),
    "soil_erosion_aggregates~large__mass_flow_rate" => @optic(_.land.soil_erosion.variables.lagg),
```

src/standard_name.jl
```diff
@@ -0,0 +1,208 @@
...
+    "land_surface_water__volume_flow_rate" =>
+        @optic(_.land.hydrological_forcing.q_land),
+    "land_surface_water__depth" => @optic(_.land.hydrological_forcing.waterlevel_land),
+    "vegetation_canopy_water__interception_volume_flux" =>
+        @optic(_.land.hydrological_forcing.interception),
+    "soil_erosion~rainfall__mass_flow_rate" =>
+        @optic(_.land.rainfall_erosion.variables.amount),
+    "soil_erosion~overland_flow__mass_flow_rate" =>
+        @optic(_.land.overland_flow_erosion.variables.amount),
+    "soil_erosion__mass_flow_rate" => @optic(_.land.soil_erosion.variables.amount),
```
hboisgon, line 145: I think these may be more mass_flux? They are equivalent in sbm to the water available before transport ie (prec-interception-infiltration+exfiltration...) so they would be equivalent to these.
And I see you used flux volume_flux for prec and interception.
I always get confused bcause of course they depend on the model timestep so interception and erosion is mm/timestep and not just mm, so maybe that's why the confusion in the metadata?

src/standard_name.jl
```diff
@@ -0,0 +1,208 @@
...
+    "soil_erosion_clay__mass_flow_rate" => @optic(_.land.soil_erosion.variables.clay),
+    "land_surface_water_sediment_transport_capacity__mass_flow_rate" =>
+        @optic(_.routing.overland_flow.transport_capacity.variables.amount),
+    "land_surface_water_clay_transport_capacity__mass_flow_rate" =>
+        @optic(_.routing.overland_flow.transport_capacity.variables.clay),
+    "land_surface_water_sediment~to-river__mass_flow_rate" =>
+        @optic(_.routing.overland_flow.to_river.variables.amount),
```
hboisgon, line 152: This is equivalent to river_water__volume_inflow_rate so I guess the name is correct?

src/standard_name.jl
```diff
@@ -0,0 +1,208 @@
...
+    "land_surface_water_clay~to-river__mass_flow_rate" =>
+        @optic(_.routing.overland_flow.to_river.variables.clay),
+    "land_surface_water_sediment__mass_flow_rate" =>
+        @optic(_.routing.overland_flow.sediment_flux.variables.amount),
+    "land_surface_water_clay__mass_flow_rate" =>
+        @optic(_.routing.overland_flow.sediment_flux.variables.clay),
```
hboisgon, line 158: This one is not so useful and can be skipped if the total is available

test/sediment_config.toml
```diff
@@ -18,144 +20,119 @@ path_output = "outstates-moselle-sed.nc"
...
+
+land_surface_water_sediment__particle_density = "sediment_density"
+land_surface_water_sediment__govers_transport_capacity_coefficient = "c_govers"
+land_surface_water_sediment__govers_transport_capacity_exponent = "n_govers"
+land_surface_water_sediment__d50_diameter = "d50_soil"
+land_surface_water_clay__d50_diameter = "dm_clay"
+land_surface_water_silt__d50_diameter = "dm_silt"
+land_surface_water_sand__d50_diameter = "dm_sand"
+"land_surface_water_aggregates~small__d50_diameter" = "dm_sagg"
+"land_surface_water_aggregates~large__d50_diameter" = "dm_lagg"
```
hboisgon, line 95: So this was my issue without mapping that the ``d50_diameter`` of clay/silt/sand/sagg/lagg/gravel in ``land_surface_water``, ``river_bottom-and-bank`` and ``river_water`` are actually the same.
With the mapping is it now possible to simplify and only define these parameters once?
Like ``clay__d50_diameter``? Or should a sediment particle struct still be defined? 
For the total sediment, the d50 is not fixed as it depends on the clay/silt/sand etc composition in each cell.

hboisgon (commented):
I checked the sediment part and most of it looks good already! Thanks @vers-w :)

Sorry that the metadata or unit in the code is not clear enough sometimes, I think also for me it's not clear that variables like precipitation or interception in mm are volumes or fluxes so the same happens for the sediment. If you know places in the code where it can be made clearer let me know and I can assist.
Note that the docs were not updated yet for sediment so that will still be on the todo list.

src/standard_name.jl
vers-w: If these variables are defined per unit area, so SI unit is [ kg m-2 s-1 ], then indeed these should be `mass_flux`. Let me know if that's the case! The quantity `mass_flow_rate` [ kg s-1 ] is the total mass per unit time (product of area and flux).

I think I got confused whether a variable represented a storage or rate, for example these variables https://github.com/Deltares/Wflow.jl/blob/master/src/sediment/sediment_transport/river_transport.jl#L17 are described what is left in the cell (store) while the metadata has `"t dt-1"` and the unit in the docs is `t`.

src/standard_name.jl
vers-w: Yes, see also my response to the previous comment, it this variable represents total mass per time (and is not defined per unit area (`flux`), then the name is correct.

test/sediment_config.toml
vers-w: Yes, it is possible to define these parameters once for the input, without the need to define a sediment particle struct. I think using one data structure is preferred (now it is stored at three different locations), we could do this refactoring as part of another PR. 

I guess these values are always the same for each cell? Then a Vector field is not required (could also be part of another PR).

src/standard_name.jl
hboisgon: OK then I think ``mass_flux`` indeed as these variables are ton / cell / timestep. They are not strorage but they are like interception for which the unit is mm / cell / timestep as well.

src/standard_name.jl
hboisgon: Yes then this is correct in this case!

test/sediment_config.toml
hboisgon: Yes the values can actually be constant and not a vector/gridded. Let's change this in a separate PR then.

src/standard_name.jl
vers-w: Hm, the SI unit of interception is [m3 m-2 s-1] and thus [m s-1], and indeed we use [mm dt-1] with model time step `dt`. So, what do you mean with cell, is that cell area?

src/standard_name.jl
vers-w: And these output parameters are also required for standard use of the sediment model? 

For expert users we could also opt to remove these parameters and an user can then use then the previous implementation (with dot notation).

src/standard_name.jl
vers-w: See previous comment about standard use of the sediment model.

src/standard_name.jl
```diff
@@ -0,0 +1,206 @@
...
+        @optic(_.routing.river_flow.boundary_conditions.lake.variables.outflow_av),
+    "soil_water__infiltration_volume_flux" => @optic(_.land.soil.variables.actinfilt),
+    "soil_water__transpiration_volume_flux" =>
+        @optic(_.land.soil.variables.transpiration),
+    "soil_surface_water__runoff_volume_flux" => @optic(_.land.soil.variables.runoff),
+    "soil_surface_water__net_runoff_volume_flux" =>
+        @optic(_.land.soil.variables.net_runoff),
+    "soil_layer_water__volume_fraction" => @optic(_.land.soil.variables.vwc),
+    "soil_layer_water__volume_percentage" => @optic(_.land.soil.variables.vwc_perc),
+    "soil_water_root-zone__volume_fraction" => @optic(_.land.soil.variables.vwc_root),
```
JoostBuitink, line 56: Amount of water in mm as well?
```
"soil_water_root-zone__storage" => @optic(_.land.soil.variables.roostore),
```

src/standard_name.jl
```diff
@@ -0,0 +1,206 @@
...
+    "floodplain_water__time_average_of_volume_flow_rate" =>
+        @optic(_.routing.river_flow.floodplain.variables.q_av),
+    "reservoir_water__volume" =>
+        @optic(_.routing.river_flow.boundary_conditions.reservoir.variables.volume),
+    "reservoir_water~outgoing__time_average_of_volume_flow_rate" =>
+        @optic(_.routing.river_flow.boundary_conditions.reservoir.variables.outflow_av),
+    "reservoir_water__volume_fraction" =>
+        @optic(_.routing.river_flow.boundary_conditions.reservoir.variables.percfull),
+    "lake_water__volume" =>
+        @optic(_.routing.river_flow.boundary_conditions.lake.variables.storage),
```
JoostBuitink, line 45: Perhaps lake water level as well?
```
    "lake_water_level__elevation" =>
        @optic(_.routing.river_flow.boundary_conditions.lake.variables.waterlevel),
```

test/io.jl
```diff
@@ -17,30 +17,26 @@ config = Wflow.Config(tomlpath)
...
-    ) == "thetaS"
+    @test collect(keys(config.output)) == ["variables", "path"]
+
+    # test if soil_water__saturated_volume_fraction can also be provided under the alias theta_s
+    lens = @optic(_.input.parameters.soil_water__saturated_volume_fraction)
+    lens_alias = @optic(_.input.parameters.theta_s)
+    @test Wflow.get_alias(config, lens, lens_alias, nothing) == "thetaS"
+    val = pop!(config.input.parameters, "soil_water__saturated_volume_fraction")
+    config.input.parameters["theta_s"] = val
+    @test Wflow.get_alias(config, lens, lens_alias, nothing) == "thetaS" == "thetaS"
```
JoostBuitink, line 32: This last part is not necessary right?

```suggestion
    @test Wflow.get_alias(config, lens, lens_alias, nothing) == "thetaS"
```

JoostBuitink (commented):
Nice work, and great steps for wflow as well! I have added a couple of comments and suggestions for additional parameters. I agree with the names are they are currently there, I think they nicely describe what they represent.

test/sediment_config.toml
```diff
@@ -18,144 +20,108 @@ path_output = "outstates-moselle-sed.nc"
...
+river_water_sediment__kodatie_transport_capacity_a-coefficient = "a_kodatie"
+river_water_sediment__kodatie_transport_capacity_b-coefficient = "b_kodatie"
+river_water_sediment__kodatie_transport_capacity_c-coefficient = "c_kodatie"
+river_water_sediment__kodatie_transport_capacity_d-coefficient = "d_kodatie"
+
+river_bottom-and-bank_sediment__d50_diameter = "D50_River"
+river_bottom-and-bank_clay__mass_fraction = "ClayF_River"
+river_bottom-and-bank_silt__mass_fraction = "SiltF_River"
+river_bottom-and-bank_sand__mass_fraction = "SandF_River"
+river_bottom-and-bank_gravel__mass_fraction = "GravelF_River"
```
vers-w, line 105: @hboisgon : not sure if these standard names are now completely correct. For example both `river_bottom-and-bank_sediment__d50_diameter` and `river_water_sediment__d50_diameter` refer to the same netCDF parameter `"D50_River"`. Based on the comments in the code I think `river_bottom-and-bank_sediment__d50_diameter` is correct, as it is described as: # Mean diameter in the river bed/bank. But, is this parameter really different from `river_water_sediment__d50_diameter`? 

I think for some of the parameters the object `water` can probably be removed. For the transport capacity related parameters using `water` is fine as it refers to sediment transport by river water. For the quantities `particle_density` and `d50_diameter` I think `water` can be removed, as it not only refers to sediment in water (e.g. suspended and bedload) but to river sediment or land sediment/soil surface generally (if I understand it correctly).

src/sediment/sediment_transport/river_transport.jl
```diff
@@ -287,19 +237,21 @@ function SedimentRiverTransportParameters(dataset, config, indices)
...
             sel = indices,
             type = Float,
             fill = 0,
         )
+        lens = lens_input_parameter("lake_surface__area")
         lakearea = ncread(
             dataset,
             config,
-            "routing.river_flow.sediment_flux.parameters.lakearea";
+            lems;
```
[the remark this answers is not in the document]
vers-w: Yes, not sure if in the tests lakes are included, @hboisgon ?

src/standard_name.jl
hboisgon: yes cell area. It's really same as interception but m3 are replaced with tons.

test/sediment_config.toml
hboisgon: So they are still different but there is no easy data to characterise them. The bed and bottom is like the original channel composition (geology linked to the soil, think of the river bank) and the other one may change depending on deposition and erosion (sediments from upstream may be deposited on the original river material or with erosion, only gravels stay at the bottom of the river). Because there is no data for it then that's why in the default settings the same netcdf map is read. The river_water_sediment_d50 is actually a calibration parameter for some of the river transport concepts.
Bottom line: they are different!

src/sediment/sediment_transport/river_transport.jl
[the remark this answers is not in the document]
hboisgon: Maybe not... depend if there are any in the test case. But there are reservoirs for sure and in terms of sediment they behave exactly the same ie I read lake area and reservoir separately but I then create directly a waterbody variable which combines both. So if something is still wrong it should only be when reading.

src/standard_name.jl
vers-w: After some checking of the units with @hboisgon: `mass_flow_rate` is correct.

src/standard_name.jl
```diff
@@ -0,0 +1,206 @@
...
+        @optic(_.land.soil.variables.net_runoff),
+    "soil_layer_water__volume_fraction" => @optic(_.land.soil.variables.vwc),
+    "soil_layer_water__volume_percentage" => @optic(_.land.soil.variables.vwc_perc),
+    "soil_water_root-zone__volume_fraction" => @optic(_.land.soil.variables.vwc_root),
+    "soil_water_root-zone__volume_percentage" =>
+        @optic(_.land.soil.variables.vwc_percroot),
+    "soil_water_sat-zone_top__recharge_volume_flux" =>
+        @optic(_.land.soil.variables.recharge),
+    "soil_layer_water_unsat-zone__depth" =>
+        @optic(_.land.soil.variables.ustorelayerdepth),
```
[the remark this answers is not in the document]
vers-w: This variable is required for the state, I also did add `ustoredepth`.

src/standard_name.jl
vers-w: Sounds good!

test/sediment_config.toml
vers-w: See also #537.

src/standard_name.jl
vers-w: Nice catch, yes this variable is required (state)!
Did rename `level` to `surface` as it is used by CSDMS and more consistent with other standard names.

JoostBuitink (approved):
Great work; looks good to me!

PR #522, opened 2025-01-06
Rename to depends-on in pixi.toml

visr:
Will soon be deprecated.
See https://github.com/Deltares/Ribasim/pull/1993


vers-w (approved):
Thanks for updating this!

Issue #90, opened 2021-01-13
include mass/water balance checks


visr:
probably at Model level

PR #525, opened 2025-01-14
Water balance computation setup

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/90

## Explanation
I created a new script `water_balance.jl`, with:
- Structs with arrays for in-place computations of water balance errors
- Associated constructors
- Functions for evaluating the water balance errors for `vertical` and `over_land` separately

I have a few open questions about this:
- How to deal with the start of the simulation? I.e. data from before and after a timestep is needed to compute a water balance error. An explicit example: how to obtain `storage_total_prev` after the first timestep in `compute_water_balance!` for vertical processes
- What is the cleanest way to deduce how much cells (`n`) there are (for array initialization)?
- Please check which variables I have to use to cover all water storage in the model
- What should be done with the water balance data (in this PR)? Should we already implement an error when certain bounds are exceeded? Should the data be written to an output file? 

## Checklist

### DIFF
--- a/src/water_balance.jl
+++ b/src/water_balance.jl
@@ -1,0 +1,97 @@
+struct WaterBalanceVertical{T <: AbstractFloat} # Terms in [mm]
+    storage_total::Vector{T}
+    storage_total_prev::Vector{T}
+    inflow_total::Vector{T}
+    outflow_total::Vector{T}
+    balance_error::Vector{T}
+    relative_error::Vector{T}
+end
+
+function WaterBalanceVertical(n::Integer, float_type::Type{T}) where {T <: AbstractFloat}
+    WaterBalanceVertical(ntuple(_ -> fill(float_type(Nan), n), 6)...)
+end
+
+function compute_water_balance!(model, water_balance::WaterBalanceVertical)::Nothing
+    (;
+        storage_total,
+        storage_total_prev,
+        inflow_total,
+        outflow_total,
+        balance_error,
+        relative_error,
+    ) = water_balance
+    (; vertical, lateral) = model
+
+    (; soil, runoff) = vertical
+    (; satwaterdepth, ustoredepth, actevap) = soil.variables
+    (; runoff_land, net_runoff_river) = runoff.variables
+
+    (; subsurface) = lateral
+    (; ssf, ssfin) = subsurface.variables
+    (; flow_length, flow_width) = subsurface.parameters
+
+    storage_total .= satwaterdepth
+    storage_total .+= ustoredepth
+
+    inflow_total .= vertical.atmospheric_forcing.precipitation
+    inflow_total .+= runoff.variables.net_runoff_river
+    inflow_total .+= 1000.0 * ssfin ./ (flow_length .* flow_width)
+
+    outflow_total .= runoff_land
+    outflow_total .+= net_runoff_river
+    outflow_total .+= actevap
+    outflow_total .+= 1000.0 * ssf ./ (flow_length .* flow_width)
+
+    @. balance_error = (inflow_total - outflow_total - (storage_total - storage_total_prev))
+    @. relative_error = balance_error / ((inflow_total + outflow_total) / 2)
+
+    copyto!(storage_total_prev, storage_total)
+
+    #TODO : What to do with results?
+
+    return nothing
+end
+
+struct WaterBalanceOverLand{T <: AbstractFloat} # Terms in [?]
+    volume_t_prev::Vector{T}
+    volume_rate::Vector{T}
+    water_in::Vector{T}
+    water_out::Vector{T}
+    balance_error::Vector{T}
+    relative_error::Vector{T}
+end
+
+function WaterBalanceOverLand(n::Integer, float_type::Type{T}) where {T <: AbstractFloat}
+    WaterBalanceOverLand(ntuple(_ -> fill(float_type(Nan), n), 6)...)
+end
+
+# function compute_water_balance!(model, water_balance::WaterBalanceOverLand)::Nothing
+#     (; vertical, lateral) = model
+#     (; volume_t_prev, volume_rate, water_in, water_out, balance_error, relative_error) = water_balance_overland
+
+#     (; land, river) = lateral
+#     (; volume) = land.variables
+#     volume_t = volume
+
+#     water_in .= lateral.land.qin_av
+#     @. water_in += lateral.land.qlat_av * lateral.land.dl
+
+#     water_out .= lateral.land.q_av
+
+#     @. volume_rate = (volume_t - volume_t_previous) / vertical.dt
+
+#     @. balance_error = (water_in - water_out) - volume_rate
+
+#     #TODO : What to do with results?
+
+#     return nothing
+# end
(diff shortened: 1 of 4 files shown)


src/water_balance.jl
```diff
@@ -0,0 +1,97 @@
...
+    (; soil, runoff) = vertical
+    (; satwaterdepth, ustoredepth, actevap) = soil.variables
+    (; runoff_land, net_runoff_river) = runoff.variables
+
+    (; subsurface) = lateral
+    (; ssf, ssfin) = subsurface.variables
+    (; flow_length, flow_width) = subsurface.parameters
+
+    storage_total .= satwaterdepth
+    storage_total .+= ustoredepth
```
vers-w, line 34: During model initialization, `ustoredepth` is not set based on `ustorelayerdepth` (state). As discussed, it could for example be added in the initialization of the model, for that the `ustoredepth!` function can be used after the states are set (after this line for the model type `sbm`: https://github.com/Deltares/Wflow.jl/blob/master/src/sbm_model.jl#L390).

src/water_balance.jl
```diff
@@ -0,0 +1,97 @@
...
+
+    (; soil, runoff) = vertical
+    (; satwaterdepth, ustoredepth, actevap) = soil.variables
+    (; runoff_land, net_runoff_river) = runoff.variables
+
+    (; subsurface) = lateral
+    (; ssf, ssfin) = subsurface.variables
+    (; flow_length, flow_width) = subsurface.parameters
+
+    storage_total .= satwaterdepth
```
vers-w, line 33: A few storage variables are missing: 
`get_snow_storage(snow) .+ get_snow_water(snow)`
`get_glacier_store(glacier) .* get_glacier_fraction(glacier)`
`get_water_depth(demand.paddy)`
`model.land.interception.variables.canopy_storage`

src/water_balance.jl
```diff
@@ -0,0 +1,97 @@
...
+    storage_total .+= ustoredepth
+
+    inflow_total .= vertical.atmospheric_forcing.precipitation
+    inflow_total .+= runoff.variables.net_runoff_river
```
vers-w, line 37: `net_runoff_river` is not an inflow (it is an outflow). `allocation` refers to `model.land.allocation`.
```suggestion
    inflow_total .+= get_irrigation_allocated(allocation)
```

src/water_balance.jl
```diff
@@ -0,0 +1,97 @@
...
+    inflow_total .+= 1000.0 * ssfin ./ (flow_length .* flow_width)
+
+    outflow_total .= runoff_land
+    outflow_total .+= net_runoff_river
```
vers-w, line 41: As part of water demand computations water can be abstracted: `land.allocation.variables.act_groundwater_abst`. This should be added to `outflow_total`. Suggest to use a wrapper method here similar to `get_irrigation_allocated(allocation)`.

src/water_balance.jl
```diff
@@ -0,0 +1,97 @@
...
+        relative_error,
+    ) = water_balance
+    (; vertical, lateral) = model
+
+    (; soil, runoff) = vertical
+    (; satwaterdepth, ustoredepth, actevap) = soil.variables
+    (; runoff_land, net_runoff_river) = runoff.variables
+
+    (; subsurface) = lateral
+    (; ssf, ssfin) = subsurface.variables
```
vers-w, line 30: Note that these variables refer to the `LateralSSF` struct (kinematic wave for lateral subsurface flow). It would be nice to also have water balance computations for the `sbm_gwf` model type that makes use of Darcy groundwater flow for a single unconfined layer (see also \src\sbm_gwf_model.jl and \src\groundwater\aquifer.jl).

src/water_balance.jl
vers-w: Lateral snow transport is often used when snow modelling is included, see also https://github.com/Deltares/Wflow.jl/blob/master/src/sbm.jl#L88. For the water balance this means the lateral snow fluxes need to be included.

src/water_balance.jl
```diff
@@ -0,0 +1,97 @@
...
+    outflow_total .+= net_runoff_river
+    outflow_total .+= actevap
+    outflow_total .+= 1000.0 * ssf ./ (flow_length .* flow_width)
+
+    @. balance_error = (inflow_total - outflow_total - (storage_total - storage_total_prev))
+    @. relative_error = balance_error / ((inflow_total + outflow_total) / 2)
+
+    copyto!(storage_total_prev, storage_total)
+
+    #TODO : What to do with results?
```
vers-w, line 50: For now I think it is fine to compute the errors and these variables can be written as output (netCDF (gridded and scalar) or CSV).

src/water_balance.jl
```diff
@@ -0,0 +1,97 @@
...
+    water_out::Vector{T}
+    balance_error::Vector{T}
+    relative_error::Vector{T}
+end
+
+function WaterBalanceOverLand(n::Integer, float_type::Type{T}) where {T <: AbstractFloat}
+    WaterBalanceOverLand(ntuple(_ -> fill(float_type(Nan), n), 6)...)
+end
+
+# function compute_water_balance!(model, water_balance::WaterBalanceOverLand)::Nothing
```
vers-w, line 68: For most routing options an internal time step (different from the model time step) is used. The variable `volume` is available and initialized (either cold or warm state). Outflow variables (e.g. `q_av`) are available for the model time step, but inflow variables are mostly only available at the internal time step.

**Kinematic wave overland flow**
- outflow: `q_av`
- inflow: `qin + qlat * flow_length` (internal time step), instead of `qlat * flow_length`,  inwater` can also be used (valid for model time step)

**Kinematic wave river flow**
- outflow: `q_av`
- inflow: `q_in + (qlat + _inflow) * flow_length` (internal time step) (as for the kinematic wave overland flow, `inwater` (valid for model time step) can also be used (to replace `qlat * flow_length`)

**Local inertial river flow (with optional floodplain routing scheme)**
A cumulative volume error is computed, see also the code starting here https://github.com/Deltares/Wflow.jl/blob/master/src/routing/surface_local_inertial.jl#L479. I think it would be useful to have also error(s) available per model time step as part of water_balance.jl. If floodplain routing is included I would recommend to compute the water balance for river and floodplain (not separated).    

**Local inertial river and overland flow**
A cumulative error is computed, see also the code starting here https://github.com/Deltares/Wflow.jl/blob/master/src/routing/surface_local_inertial.jl#L1033. For the water balance computation per model time step it is enough to consider `land_v.volume`.

When many variables are involved (as for local inertial routing) it probably makes sense to compute a `net_flux`.

 **Kinematic wave subsurface flow**
outflow: `ssf + exfiltwater * flow_length * flow_width`
inflow: `ssfin + recharge * flow_length`
Note: see for units of variables https://github.com/Deltares/Wflow.jl/blob/master/src/routing/subsurface.jl#L87.

**Groundwater flow**
Here `Q` is used as the net flux, see also: https://github.com/Deltares/Wflow.jl/blob/master/src/groundwater/aquifer.jl#L432

PR #534, opened 2025-01-23
Give (non-type) global constants upper case names

SouthEndMusic:
My apologies for the many PR's, but this is a very simple one. All I did was give (non-type) global constants upper case names. This helps with readability of the code for people who are not familiar with all the variable names.

Edit: fixes https://github.com/Deltares/Wflow.jl/issues/535


visr:
Float is a type though right? `const Float = Float64` it is a type alias. I don't see an issue for it, but I think it should be removed altogether, so we just use Float64 instead. The original idea was to start using it to support running in single precision but this is only half (so not) supported. And if we would start supporting it we'd use type parameters rather than a type alias.

visr:
Yeah I made a separate issue, #535. I'm fine with renaming it as well in this PR.

vers-w:
> Then I think indeed it can be removed, but that might require some refactoring of type parameters

Yes, I agree this can be removed, for me it's fine to do this as part of this PR.

vers-w:
The wflow server tests are failing now.
You can run these tests (from Wflow dir):
````
julia --project=server
(WflowServer) pkg> test
````

File: server/src/bmi_service.jl [hunk 1]
```diff
@@ -85,35 +85,35 @@ struct GetVarLocation
     name::String
 end
 
-struct GetValue
+struct GetValue{T}
```
vers-w, line 88: Do we really need parametric type structs here? The tests are failing now because the JSON message with the `dest` entry is now mapped to `Vector{Any}`.

vers-w (changes requested):
LGTM! Two small comments, one related to the failing wflow server tests.

server/src/bmi_service.jl
visr: Yeah indeed I also suggested to @SouthEndMusic to just remove them for now. If we want to support parametric precision we should probably do that in a separate PR and think about which structs would need type parameters.

visr: Yes, remove them from all structs.

vers-w (approved):

server/src/bmi_service.jl
vers-w: Hi @visr, I did remove them from all structs and added more function type signatures (was also required for dispatching of constructor functions). Perhaps you could do a final (quick) review of these changes?

PR #544, opened 2025-02-03
Update dev environment to Julia 1.11

visr:
The pixi.toml now installs julia 1.11.3. I directly updated all manifests such that they are resolved with this version. Note that in the Wflow.jl's Project.toml compat it still says `julia = "1.10"`, meaning the LTS release is still supported. We are not using any Julia 1.11+ features yet. It might be a good idea to make sure that 1.10 is also tested in CI, and not only 1.11, but that should be a separate PR.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,8 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.8"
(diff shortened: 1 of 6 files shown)


vers-w (approved):
LGTM, thanks for the update.

PR #548, opened 2025-02-06
Fix typo

vers-w:
Typo was causing a large increase in memory allocation.


visr (approved):
Strange, any idea how this would have any effect on allocations? This is only assigning binding to existing vectors, doing that twice for ony variable shouln't make a difference.
```jl
julia> nt=(;a=ones(1000),b=zeros(1000))

julia> @allocated (; a, b, a) = nt
0
```

vers-w:
> Strange, any idea how this would have any effect on allocations? This is only assigning binding to existing vectors, doing that twice for ony variable shouln't make a difference.

Output from `@code_warntype` showed `Core.Box()` for this variable, and thus I guess type instability (but no clue why this variable is showing up as `Core.Box()`) as a result of the typo).

PR #549, opened 2025-02-11
Model type `sbm_gwf`: limit river flux to groundwater

### LINKED ISSUE #543: Model type `sbm_gwf`: limit river flux to groundwater
### Wflow version checks

### Reproducible Example

This issue was found when running a simulation for the Hindon catchment.

### Current behaviour

For the computation of the river flux to the groundwater (boundary condition) a check of the available river storage volume (routing) is missing. This can result in river fluxes larger than available river storage volume.

### Desired behaviour

Limit river flux.

### Additional Context

_No response_

vers-w:
Fixes #543

### DIFF
--- a/src/groundwater/boundary_conditions.jl
+++ b/src/groundwater/boundary_conditions.jl
@@ -19,11 +19,16 @@ end
 
 @get_units @grid_loc @with_kw struct RiverVariables{T}
     stage::Vector{T} | "m"
+    storage::Vector{T} | "m3"
     flux::Vector{T} | "m3 d-1"
 end
 
 function RiverVariables(n)
-    variables = RiverVariables{Float}(; stage = fill(mv, n), flux = fill(mv, n))
+    variables = RiverVariables{Float}(;
+        stage = fill(mv, n),
+        storage = fill(mv, n),
+        flux = fill(mv, n),
+    )
     return variables
 end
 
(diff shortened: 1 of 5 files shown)


src/groundwater/boundary_conditions.jl
```diff
@@ -51,19 +56,21 @@ function River(dataset, config, indices, index)
...
 
-function flux!(Q, river::River, aquifer)
+function flux!(Q, river::River, aquifer; dt = 1.0)
     for (i, index) in enumerate(river.index)
         head = aquifer.variables.head[index]
         stage = river.variables.stage[i]
         if stage > head
             cond = river.parameters.infiltration_conductance[i]
             delta_head = min(stage - river.parameters.bottom[i], stage - head)
+            flux = min(cond * delta_head, river.variables.storage[i] / dt)
```
JoostBuitink, line 66: Doesn't the conductivity also needed to be corrected by `dt`? As far as I see, the `infiltration_conductance` is defined as `m2 day-1`.

JoostBuitink (commented):
Looks good to me! Just one comment related to the units of the `infiltration_conductance`.

src/groundwater/boundary_conditions.jl
vers-w: The `flux` unit is `m3 day-1`, so as far as I see it is not required to correct by `dt`.

JoostBuitink: Ah that is correct!

JoostBuitink (approved):

PR #558, opened 2025-02-24
Fix diagonal flow width and surface flow width

### LINKED ISSUE #556: Computation of surface flow width
The computation of diagonal flow width in `get_flow_width(ldd, x_length, y_length)` (used by the `sediment` model type) is not correct, it should be based on the flow length and area of a cell (area conservative) which is not the case. In addition, the remaining surface flow width (overland flow) for river cells should also be be based on the remaining cell area.

vers-w:
Fixes #556


JoostBuitink (approved):
LGTM!

PR #567, opened 2025-03-24
Sediment model state file

### LINKED ISSUE #564: Test sediment model and incorrect state file
When running the sediment model from the test folder with warm states the following error is thrown:

```
[ Info: Set initial conditions from state file `C:\Users\verseve\.julia\dev\Wflow\test\data\input\instates-moselle-sed.nc`.
┌ Info: Setting initial state from netCDF.
│   ncpath = "C:\\Users\\verseve\\.julia\\dev\\Wflow\\test\\data\\input\\instates-moselle-sed.nc"
│   ncvarname = "siltstore"
└   state = "river_bed_silt__mass"
ERROR: arrays contains missing values (values equal to the fill values attribute in the NetCDF file)
```

It seems the states in the state file are not correct (were maybe not updated when the model was updated (the versions are different)).

vers-w:
## Issue addressed
Fixes #564

### DIFF
--- a/test/run_sediment.jl
+++ b/test/run_sediment.jl
@@ -6,6 +6,21 @@ config = Wflow.Config(tomlpath)
 model = Wflow.initialize_sediment_model(config)
 (; network) = model
 
+@testset "initial default states" begin
+    river = model.routing.river_flow
+    mean(river.sediment_flux.variables.store_clay) == 0.0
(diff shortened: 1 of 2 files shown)


JoostBuitink (approved):
Looks good to me!

hboisgon (approved):
Indeed I did not realise there was also states in the test model and I did not update those. Thanks @vers-w for correcting this!

PR #574, opened 2025-04-10
Bugfix lakes

hboisgon:
## Issue addressed
To be ddefined

## Explanation
I started here with some fixes but not complete.
I guess testing with lakes should be added to the test bench.

## Checklist

### DIFF
--- a/src/routing/lake.jl
+++ b/src/routing/lake.jl
@@ -17,13 +17,13 @@ function LakeParameters(dataset::NCDataset, config::Config, network::NetworkWate
     (; indices_outlet) = network
 
     lens = lens_input_parameter(config, "lake_surface__area"; optional = false)
-    lakearea = ncread(dataset, config, lens; sel = indices_outlet, type = Float, fill = 0)
+    lakearea = ncread(dataset, config, lens; sel = indices_outlet, type = Float64, fill = 0)
     lens = lens_input_parameter(
         config,
         "lake_water__rating_curve_coefficient";
         optional = false,
     )
-    lake_b = ncread(dataset, config, lens; sel = inds_lake, type = Float64, fill = 0)
+    lake_b = ncread(dataset, config, lens; sel = indices_outlet, type = Float64, fill = 0)
     lens =
         lens_input_parameter(config, "lake_water__rating_curve_exponent"; optional = false)
     lake_e = ncread(dataset, config, lens; sel = indices_outlet, type = Float64, fill = 0)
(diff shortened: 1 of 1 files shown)


vers-w:
I think this is ready for review, we can update the test model as part of another PR (#579).

Wflow/src/standard_name.jl
```diff
@@ -303,10 +302,6 @@ const sediment_standard_name_map = Dict{String, NamedTuple}(
         lens = @optic(_.routing.overland_flow.transport_capacity.variables.amount),
         unit = "t dt-1",
     ),
-    "land_surface_water_clay_transport_capacity__mass_flow_rate" => (
-        lens = @optic(_.routing.overland_flow.transport_capacity.variables.clay),
-        unit = "t dt-1",
-    ),
```
JoostBuitink, line 309: Perhaps I missed it, but what is the reason for dropping this standard name? I this variable being used in the toml, so doesn't it make sense to have a standard name as well?

Wflow/src/routing/lake.jl
```diff
@@ -17,13 +17,13 @@ function LakeParameters(dataset::NCDataset, config::Config, network::NetworkWate
...
     lens = lens_input_parameter(config, "lake_surface__area"; optional = false)
-    lakearea = ncread(dataset, config, lens; sel = indices_outlet, type = Float, fill = 0)
+    lakearea = ncread(dataset, config, lens; sel = indices_outlet, type = Float64, fill = 0)
     lens = lens_input_parameter(
         config,
         "lake_water__rating_curve_coefficient";
         optional = false,
     )
-    lake_b = ncread(dataset, config, lens; sel = inds_lake, type = Float64, fill = 0)
+    lake_b = ncread(dataset, config, lens; sel = indices_outlet, type = Float64, fill = 0)
```
JoostBuitink, line 26: Apparently we don't cover these lines with tests, and it looks like none of our test tomls actually have `lakes = true`. Do want want to include tests as well (maybe part of a separate PR)?

JoostBuitink (approved):
Just two questions for clarification, all code changes look good!

Wflow/src/standard_name.jl
vers-w: Because only `clay` fraction and `amount` (total sediment transport capacity) are part of the standard names for transport_capacity. Total amount makes sense I guess, but if we want to support `clay` fraction we should also add the other fractions (`silt`, `sand`, etc.). @hboisgon : do you agree with dropping this variable?

Wflow/src/routing/lake.jl
vers-w: Yes, see above https://github.com/Deltares/Wflow.jl/pull/574#issuecomment-2796028841 (part of another PR, @hboisgon did update the Piave test model so it includes lakes.

JoostBuitink: Ah overlooked that, sorry!

Issue #576, opened 2025-04-10
Move Wflow to subdir


visr:
Right now the Wflow package lives in the root of the repository. I'd like to propose creating a new `/Wflow/` directory and moving `src` and `test` as well as `/Project.toml` in there.

This will make it easier to add a new shared module next to the Wflow package; https://github.com/Deltares/Ribasim/issues/2232.

This has the additional advantage that if a user installs Wflow from Julia they don't also get `build`, `docs`, `server` as well as other files in the repo root, that are not relevant to the Wflow Julia package.

Ribasim has had this setup since we use it as a monorepo together with Python and QGIS developments: https://github.com/Deltares/Ribasim/tree/v2025.2.0/core.

- [ ] This also needs to be updated in the General registry, similar to https://github.com/JuliaRegistries/General/pull/118470.
- [x] Probably needs some updates in TeamCity as well after merging.
- [ ] Update links in the docs (installing) after including a subdir in the General registry
- [ ] Update links to example tomls in the docs if we decide to use a different subdir name

JoostBuitink:
Maybe good to reconsider the name of the subdir to maybe something like `core`, as that might be a bit nicer than the current name,

visr:
Yeah either works. Ribasim also uses core. I went with Wflow because normally the convention is to use the package name for the dir/report, and there was already WflowServer doing that as well. Though in the case of Ribasim I found core clearer, because the Python package is also called ribasim.

PR #578, opened 2025-04-10
Move Wflow to /Wflow

visr:
## Issue addressed
Fixes most of #576, except for the General registry update.

## Explanation
Moved files, updated Manifests, CI, and pixi.toml where needed.
Copied LICENSE.md since the Wflow subdir is a separate artifact for the registry that needs a license.


vers-w (commented):
LGTM!

PR #580, opened 2025-04-11
Create a single root environment with all dependencies

### LINKED ISSUE #577: Create shared Manifest.toml
Related to #576, I'd like to move to having only one Manifest.toml in the repo root. This was also done for Ribasim in https://github.com/Deltares/Ribasim/pull/740 and simplified things. This will be easier to manage then each Manifest separately, for instance when doing updates like #544. The repo root Manifest would have all Wflow test dev and server dependencies together. During development this is usually the one that you'd want to have active so you can directly use you dev dependencies as well.

Later with Julia 1.12 we can then make the next step and start using the `[workspace]` section: https://pkgdocs.julialang.org/dev/toml-files/#The-%5Bworkspace%5D-section.

visr:
## Issue addressed
Fixes #577

## Explanation
This folds the build projects into the root environment, and only retains 3 separate projects, with one manifest.
- Wflow: the package
- WflowServer: the package
- root: dev environment with build tooling like PackageCompiler, and all Wflow and WflowServer dependencies

The root environment does not maintain a separate [compat] section, the package projects do need that, and it isn't convenient to maintain that in two places. Note that the Project.toml now have:

```toml
[sources]
Wflow = {path = "Wflow"}
```

to point to Wflow, making a `Pkg.dev` in CI not needed.

and

```toml
manifest = "../Manifest.toml"
```

is used to point to the shared manifest.


visr:
Closing, see https://github.com/Deltares/Wflow.jl/issues/577#issuecomment-2796910094.

PR #587, opened 2025-04-15
Change soil infiltration capacity parameter

### LINKED ISSUE #385: Duplicate infiltration parameters (`infiltcapsoil` and `kv_0`)
### Feature type

### Improvement Description

Currently, we use `infiltcapsoil` to describe the infiltration capacity of the soil, while we also have a conductivity (`kv_0`, at the surface) being set that is representative for the soil. It currently does not really make sense to have two parameters describing essentially the same process.

It would be better to use the `kv_0` values instead of the `infiltcapsoil` values. Also because the `infiltcapsoil` values are currently produced as constant values in hydromt_wflow (of 600 mm/day), and it makes more sense to link those to the soil properties. 

The `infiltcappath` still would make sense to continue supporting, to simulated the reduced infiltration capacities of paved areas. 

When this issue is being picked up, changes in hydromt_wflow are also required.

_No response_

### Additional Context

_No response_

vers-w:
## Issue addressed
Fixes #385


JoostBuitink (approved):
LGTM!

PR #589, opened 2025-04-16
Local inertial routing: limit negative external inflow

### LINKED ISSUE #588: Local inertial routing: limit negative inflow
For river local inertial routing the river volume can be abstracted completely. It would be better to use for example a fixed fraction of total river storage as maximum allowed abstraction similar to the kinematic wave implementation.

For the river-overland flow local inertial model a check on a maximum allowed abstraction is missing.

vers-w:
## Issue addressed
Fixes #588


JoostBuitink (approved):
Looks good to me!

PR #591, opened 2025-04-18
Update test model Piave

### LINKED ISSUE #579: Update test model Piave
Add lakes (reservoirs are already included) to the Piave test model, so this functionality is also tested as part of a model run (see also #575). 

This test model is also part of water demand and allocation tests, so abstractions from water bodies are then also included and tested.

vers-w:
## Issue addressed
Fixes #579


JoostBuitink (approved):
Nice to see these tests added, all looks good to me!

PR #594, opened 2025-04-21
Expose the rainfall erosion factor of ANSWERS equation

hboisgon:
## Issue addressed
No issue. While working on a project, I would like to be able to calibrate a rainfall erosion factor to balance effect of rainfall erosion versus overland flow erosion but cannot as this parameter is encoded as a fixed value for rainfall erosion (0.108) but is already flexible for overland flow erosion (answers_k).

Because it needs to be added to standard names in v1, I thought better to do it now before the release (it was also on my mind for a long time).

## Checklist
- [x] Updated tests or added new tests -> not needed, the default can still be used
- [x] Updated changelog.qmd if needed -> it's a small change compared to the rest so not sure it's needed

When the docs PR is merged, the new parameter `soil_erosion__answers_rainfall_factor` can be added to the list below in sediment_parameters.qmd ( ANSWERS rainfall erosion factor | -  | 0.108 |

image


JoostBuitink (approved):
All code changes seem fine to me, I just have one comment/question related to the naming/use of this parameter!

Wflow/src/sediment/erosion/rainfall_erosion.jl
```diff
@@ -186,8 +188,11 @@ function RainfallErosionAnswersParameters(
     usle_k = ncread(dataset, config, lens; sel = indices, defaults = 0.1, type = Float64)
     lens = lens_input_parameter(config, "soil_erosion__usle_c_factor")
     usle_c = ncread(dataset, config, lens; sel = indices, defaults = 0.01, type = Float64)
+    lens = lens_input_parameter(config, "soil_erosion__answers_rainfall_factor")
+    answers_overland_flow_factor = ncread(dataset, config, lens; sel = indices, defaults = 0.108, type = Float64)
+    
     answers_parameters =
-        RainfallErosionAnswersParameters(; usle_k = usle_k, usle_c = usle_c)
+        RainfallErosionAnswersParameters(; usle_k = usle_k, usle_c = usle_c, answers_rainfall_factor = answers_overland_flow_factor)
```
vers-w, line 195:
```suggestion
        RainfallErosionAnswersParameters(; usle_k = usle_k, usle_c = usle_c, answers_rainfall_factor = answers_rainfall_factor)
```

hboisgon:
> Isn't the naming here a little bit confusing? Perhaps because I don't know the sediment model too well, but I see the same factor being used for overland flow and rainfall erosion. Is it a coincidence that they have the same parameter value, or does that make sense? E.g., is it unwanted to have one parameter for rainfall erosion, and one parameter for overland flow erosion?

Sorry they are different but it was bad "copy pasting" from my part and I did not change the internal names. Thanks WIllem for the suggestions! I also then merged with `master` and updated the docs so hopefully this should now be ready to merge :)

PR #596, opened 2025-04-22
Drop `GroundwaterExchange` component

### LINKED ISSUE #593: Drop `GroundwaterExchange` component
For v1.0-beta I think we should drop the `GroundwaterExchange` component (can be used to replace the kinematic wave lateral subsurface flow routing with an external groundwater model (e.g. Modflow). This feature has been used only (to my knowledge) as part of the following [MSc thesis ](https://studenttheses.uu.nl/handle/20.500.12932/41091) work. One of the potential improvements recommended by this work was to re-activate the lateral subsurface flow routing for coupling with Modflow 6, as a rapid runoff (interflow) generating mechanism was missing and the current coupling did not yield satisfactory results. So, to summarize, more work is needed to support this kind of coupling in a good way.

vers-w:
## Issue addressed
Fixes #593


JoostBuitink (approved):
Indeed good to drop this! Changes look good to me!

PR #597, opened 2025-04-23
Extend standard names

### LINKED ISSUE #573: Extend standard names
+ for DELWAQ coupling
+ possibly for 1D river routing coupling, although it is probably cleaner to let the external model deal with this (e.g. dealing with precip-evap (net runoff) on their own and subtract this quantity from the lateral inflow rate (`river_water_inflow~lateral__volume_flow_rate`).

vers-w:
## Issue addressed
Fixes #573

### DIFF
--- a/docs/model_docs/parameters_landhydrology_sbm.qmd
+++ b/docs/model_docs/parameters_landhydrology_sbm.qmd
@@ -99,8 +99,10 @@ td, th {
 ### Output
 | Standard Name | Description | Unit |
 | ---- | --- | - |
+| `vegetation__leaf-area_index` | Leaf area index | m² m⁻² |
 | `vegetation_canopy_water__stemflow_volume_flux` | Stemflow | mm Δt⁻¹ |
 | `vegetation_canopy_water__throughfall_volume_flux` | Throughfall | mm Δt⁻¹ |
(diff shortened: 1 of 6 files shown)


hboisgon (approved):
Looks good! I think all the variables I mentioned are added to standard names and the docs! I didn't check the other adjustments in the docs very carefully but trust you.

docs/model_docs/parameters_landhydrology_sbm.qmd
```diff
@@ -99,8 +99,10 @@ td, th {
 ### Output
 | Standard Name | Description | Unit |
 | ---- | --- | - |
+| `vegetation__leaf-area_index` | Leaf area index | m² m⁻² |
```
JoostBuitink, line 102: We already listed this one a input, do we want to mention it here again?

JoostBuitink (approved):
Just one question related to LAI being in the output section, and removed a couple of duplicate spaces. But other than that, looks good to me!

docs/model_docs/parameters_landhydrology_sbm.qmd
vers-w: Nice catch, I did remove it, so it is consistent with other cyclic input parameters.

Issue #585, opened 2025-04-15
Update/check TOML model setting names


vers-w:
We did already change some names (more explicit), need to check if other setting names are clear or need to change.

hboisgon:
Kind of related but I did a bad review and the exftilsatstore points to the ustore variables (line 164 in standard_names.jl)

```
    "soil_surface_water_unsat-zone__exfiltration_volume_flux" =>
        (lens = @optic(_.land.soil.variables.exfiltustore), unit = "mm dt-1"),
    "soil_surface_water_sat-zone__exfiltration_volume_flux" =>
        (lens = @optic(_.land.soil.variables.exfiltustore), unit = "mm dt-1"),
```

sould be 

```
    "soil_surface_water_unsat-zone__exfiltration_volume_flux" =>
        (lens = @optic(_.land.soil.variables.exfiltustore), unit = "mm dt-1"),
    "soil_surface_water_sat-zone__exfiltration_volume_flux" =>
        (lens = @optic(_.land.soil.variables.exfiltsatwater), unit = "mm dt-1"),
```

PR #602, opened 2025-05-02
Sediment renaming

hboisgon:
## Issue addressed
I imagine linked to #585.

## Explanation
I went through all standard names and made them fully explicit (d50 for median diameter for still in short name).

I also noticed that sediment__particle_density was still split for land and river while it's a fixed property like clay mean particle diameter so does not need to be defined twice.

## Checklist


vers-w (changes requested):
LGTM. Just two small comments related to `d50` renaming to `mean`, is this correct? As `d50` normally refers to `median`. I think it is fine to use `sediment__particle_density` if a distinction between `land` and `river` is not required (also in the long term by users).

vers-w (approved):

Wflow/src/sediment/sediment_transport/river_transport.jl
```diff
@@ -181,17 +181,17 @@ function SedimentRiverTransportParameters(
     if any(abs.(river_fractions .- 1.0) .> 1e-3)
         error("Particle fractions in the river bed must sum to 1")
     end
-    lens = lens_input_parameter(config, "clay__d50_diameter")
+    lens = lens_input_parameter(config, "clay_mean_diameter")
```
JoostBuitink, line 184: Just a very quick drive-by comment, shouldn't it have double underscores as well? Same holds for other changes as well
```suggestion
    lens = lens_input_parameter(config, "clay__mean_diameter")
```

hboisgon: You are very right! Thanks for spotting this

PR #603, opened 2025-05-06
Check and update model setting names

### LINKED ISSUE #585: Update/check TOML model setting names

vers-w:
## Issue addressed
Fixes #585


JoostBuitink (approved):
Looks good to me! Much clearer names, and also very nice that the local inertial settings can now be controlled for river and land parts separately!

PR #606, opened 2025-05-08
Check optional parameters

### LINKED ISSUE #561: Consider which parameters are optional/can get a default value
### Feature type

### Improvement Description

Some parameters currently get a default value if they are not supplied by the user. While this is fine for some layers, it is not for other more essential parameter maps (such as theta_s, theta_r, kv_0, etc). I would be in favor that if one of these essential parameters are not provided by the user, the wflow simulation would not be possible to start. If you really want to use a default/fixed value for these parmeters, you can always pass that via the toml using the `variable.value = 3000` syntax.

A bit similar to the glacierfrac mentioned in this issue: https://github.com/Deltares/Wflow.jl/issues/322

_No response_

### Additional Context

_No response_

vers-w:
## Issue addressed
Fixes #561

### DIFF
--- a/Wflow/src/groundwater/aquifer.jl
+++ b/Wflow/src/groundwater/aquifer.jl
@@ -134,6 +134,7 @@ function UnconfinedAquiferParameters(
     bottom::Vector{Float64},
     area::Vector{Float64},
 )
+    conductivity_profile = get(config.model, "conductivity_profile", "uniform")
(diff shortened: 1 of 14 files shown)


JoostBuitink (approved):
Nice, thanks for addressing this! I only have one small comment related to the changelog, but other than that it is good to be merged!

PR #608, opened 2025-05-12
Support computation and output of actual river water abstraction

### LINKED ISSUE #598: Keep track of actual negative external inflow
Negative external inflow (abstraction) is limited by available river volume (and inflows in case of kinematic wave routing). Keeping track of actual abstraction and exposing this variable through a standard name would be useful.

vers-w:
## Issue addressed
Fixes #598


JoostBuitink (approved):
LGTM!

PR #623, opened 2025-05-22
Fix init state and boundary condition local inertial model

### LINKED ISSUE #622: Fix init state and reservoir/lake state 1d-2d local inertial model
Setting the state of the local inertial model for land routing is not working, broken during refactoring of the code for v1.0.

Improved the accuracy of reservoir or lake state after initialization with a warm state by re-ordering computations of 1d-2d local inertial model. The inflow from land to a reservoir or lake was updated during the local inertial overland flow routing (final computation during each internal time step of the 1d-2d local inertial model). This results in small inflow deviations when starting with a warm state as the inflow from land to a reservoir or lake is not yet updated while a reservoir or lake is updated as part of the river routing. Order of computations is now: 1) overland flow routing, 2) reservoir or lake inflow and 3) river routing.

vers-w:
## Issue addressed
Fixes #622

### DIFF
--- a/Wflow/src/sbm_model.jl
+++ b/Wflow/src/sbm_model.jl
@@ -168,11 +168,11 @@ function set_states!(model::AbstractModel{<:Union{SbmModel, SbmGwfModel}})
             end
             land_v.storage .= land_v.h .* surface_flow_width .* flow_length
         elseif land_routing == "local-inertial"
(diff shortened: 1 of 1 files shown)


JoostBuitink (approved):
Good that you spotted this issue! Changes look good and correct to me!

PR #625, opened 2025-05-30
Add pixi task to genenate SBOM

deltamarnix:
## Explanation
Generate an SPDX standardized SBOM with all dependencies used by Wflow.jl

Still misses Julia and its depending third parties. See https://github.com/JuliaLang/julia/blob/master/THIRDPARTY.md

Opened an issue to add even more information by PkgToSoftwareBOM: https://github.com/SamuraiAku/PkgToSoftwareBOM.jl/issues/38

This task could be added to when the project.toml file is being updated in GitHub actions.

We are also checking how this works out for Ribasim in https://github.com/Deltares/Ribasim/pull/2306

## Checklist


JoostBuitink:
Thanks for adding this. I see it breaks our tests (as the `PkgToSoftwareBOM` is marked as a "stale" dependency by Aqua, since it is not actively used by wflow). Given it's role is only to perform this task, do we want to list this package as a dependency of the Wflow.jl package? Or can we perhaps add it as part of the pixi task?

deltamarnix:
> Thanks for adding this. I see it breaks our tests (as the `PkgToSoftwareBOM` is marked as a "stale" dependency by Aqua, since it is not actively used by wflow). Given it's role is only to perform this task, do we want to list this package as a dependency of the Wflow.jl package? Or can we perhaps add it as part of the pixi task, or within the generate-sbom.jl script?

Yeah it seems to be a little different than the Ribasim repository. They have two Project.toml files, where the root file also contains development packages. So I didn't have that problem there.
I was hoping to make custom targets so I could put the package in `[extras]`, but that is not yet implemented in Julia package manager.

Would it be an option to add it as a Julia test and ignore the test usually, except for when we explicitly call `Pkg.test("generate_sbom")`?
Or is it possible to ignore it in Aqua?

JoostBuitink:
> Would it be an option to add it as a Julia test and ignore the test usually, except for when we explicitly call `Pkg.test("generate_sbom")`? Or is it possible to ignore it in Aqua?

Not sure if I understand your first suggestion, but did a quick look at the Aqua call, and we can ignore certain packages with `Aqua.test_all(Wflow; ambiguities = false, persistent_tasks = false, stale_deps = (ignore = [:PkgToSoftwareBOM],))`

But agree with @vers-w, if possible, it would be nicer to not have this package as a (direct) dependency to the "wflow-core"

deltamarnix:
I think I can make a separate folder next to Wflow and make a new Project.toml next to it that will only generate the SBOM.

```toml
name = "Wflow SBOM Generator"
uuid = "f63e2d3d-2108-4529-8446-85f2b0f0c21f"

[deps]
PkgToSoftwareBOM = "6254a0f9-6143-4104-aa2e-fd339a2830a6"
Wflow = "d48b7d99-76e7-47ae-b1d5-ff0c1cf9a818"

[sources]
Wflow = { path = "../../Wflow" }
```

vers-w:
> I think I can make a separate folder next to Wflow and make a new Project.toml next to it that will only generate the SBOM.

Sound good, we do this also for the Wflow `server` (folder).

utils/sbom/generate-sbom.jl
```diff
@@ -0,0 +1,30 @@
+using PkgToSoftwareBOM
+using UUIDs
+using Pkg
+
+Pkg.activate("../../Wflow")
+
+wflowLicense = SpdxLicenseExpressionV2("MIT")
+organization = SpdxCreatorV2("Organization", "Deltares", "software@deltares.nl")
```
vers-w, line 8: What about using here the Wflow@deltares.nl email?

vers-w (approved):
LGTM, one small comment about the email address.

utils/sbom/generate-sbom.jl
vers-w: @deltamarnix : do you agree with changing the email address?

deltamarnix: @alimeshgi If you want to make use of the SST channels for software support, I would keep it software@deltares.nl. If you have a separate wflow support channel you might want to change it to wflow@.

PR #628, opened 2025-06-10
Fix lake abstraction (water demand and allocation)

vers-w:


JoostBuitink (approved):
Good spot, thanks for fixing this!

PR #633, opened 2025-06-12
feat: adding LandSurfaceTemperature module for LST modeling

### LINKED ISSUE #635: Support land surface temperature (LST) computation
Implementation of methods to compute LST based on the following paper: "Identifying irrigated areas using land surface temperature and hydrological modelling: application to the Rhine basin" [doi](https://doi.org/10.5194/hess-29-1483-2025).

This can be implemented as:
1. Part of the Wflow core (`\Wflow\src\`) or 
2. As a separate module outside the core like `WflowServer`.

@aweerts and Jaap Schellekens: Do you have a preference (e.g. what is the use case)?

michaelohanrahan:
## Explanation
Working towards implementation of the methods from the Devi Purnamasari et al. 2025 paper: "Identifying irrigated areas using land surface temperature and hydrological modelling: application to the Rhine basin" [doi](https://doi.org/10.5194/hess-29-1483-2025)

## Implementation details:
The land surface temperature (LST) module is an optional component of the `SBM` land hydrology model, controlled by the `land_surface_temperature__flag` in the model config. When enabled, a `LandSurfaceTemperatureModel` is created and stored as a field in `LandHydrologySBM`, alongside `interception`, `snow`, `soil`, and other land processes. The LST model uses atmospheric forcing (temperature, wind speed, net radiation), actual evapotranspiration, land surface roughness parameters (momentum and heat transfer), zero plane displacement height and land surface "skin" layer height above the ground to compute aerodynamic resistance, sensible heat flux, latent heat flux and land surface temperature per cell. When the flag is set to `false` (default), a `NoLandSurfaceTemperatureModel` is used instead, and the LST update is skipped.

## Checklist

A unit test has been added for LST, would be good to add an integration test for LST.


Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
+module LandSurfaceTemperature
+
+using Wflow
+using Dates
+import Wflow: AbstractLandModel, AtmosphericForcing, SbmSoilModel, Config
+using NCDatasets
+using Parameters
```
vers-w: For now I would suggest not to use `module` here, but to implement this feature as part of the `LandHydrologySBM` model (like `snow` etc.).

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+
+#abstract should define a common interface
+abstract type AbstractLSTModel end
```
vers-w: I would recommend to use more explicit naming (e.g. here `LandSurfaceTemperature`) as much as possible throughout the code.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+#a struct or "object" is a collection of fields
+#a field is a variable that belongs to a struct
+#@with_kw is a macro that creates a struct with fields and default values
+const sigma = 5.67e-8  # Stefan-Boltzmann constant for blackbody radiation
+const cp = 1013.0      # Specific heat capacity of air in J/kg/K
+const rho_a = 1.29     # Density of air in kg/m3
+const rho_w = 1000.0     # Density of water in kg/m3
+#TODO: make crop height a distributed parameter
+const hc = 2.0        # crop height in m
+const k = 0.41        # von Kármán constant
```
vers-w: We work with shared parameters, I would say most of these can be part of `LandParameters` (\src\domain.jl) or vegetation parameters (\src\vegetation\parameters.jl).

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+#HYRAS input is 10m we correct to 2m in calculation for aerodynamic resistance
```
vers-w: I think computation of windspeed at 2 m should be done by hydromt_wflow?

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+const zm = 2.0        # height of wind speed measurement = zh
+const zh = 2.0        # height of air temperature measurement = zm
+const dt = 86400.0     # timestep in seconds
+
+@with_kw struct LSTModel <: AbstractLandModel
```
vers-w: `AbstractLSTModel`?

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+    # Input forcing from atmospheric module
+    atmospheric_forcing::AtmosphericForcing
```
vers-w: Would recommend to make a distinction between variables and parameters here like the other wflow models.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+
+    # Input radiation
+    RS_in::Vector{Float64}    # Downward shortwave radiation (W/m2)
```
vers-w: This can be part of `AtmosphericForcing`.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+const cp = 1013.0      # Specific heat capacity of air in J/kg/K
+const rho_a = 1.29     # Density of air in kg/m3
+const rho_w = 1000.0     # Density of water in kg/m3
+#TODO: make crop height a distributed parameter
+const hc = 2.0        # crop height in m
+const k = 0.41        # von Kármán constant
+#HYRAS input is 10m we correct to 2m in calculation for aerodynamic resistance
+const zm = 2.0        # height of wind speed measurement = zh
+const zh = 2.0        # height of air temperature measurement = zm
+const dt = 86400.0     # timestep in seconds
```
vers-w: If we run this model as part of updating `LandHydrologySBM` then `dt` is already part of that function.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+    # Input forcing from atmospheric module
+    atmospheric_forcing::AtmosphericForcing
+
+    # Input radiation
+    RS_in::Vector{Float64}    # Downward shortwave radiation (W/m2)
+
+    # Static parameters (from netCDF or defaults)
+    albedo::Vector{Float64}   # Surface albedo
+    crop_height::Vector{Float64} # Crop height in m
+    emissivity::Vector{Float64} # Surface emissivity
```
vers-w: These can be part of shared land or vegetation parameters.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+    latitude::Vector{Float64}  # Latitude of each grid cell (degrees)
```
vers-w: We also support other projections. This parameter can be stored as part of the `Network` information, e.g. `NetworkLand` in \src\network.jl.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+
+    # Time information
+    doy::Vector{Int}          # Day of year for each timestep
```
vers-w: We also use `doy` information for reservoirs, at a later stage probably good to store this more centrally, e.g. as part of `Clock`. Or just using `dayofyear(clock.dt)` in the update function is also fine. I don't think a `Vector{Int}` is required here?

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,305 @@
...
+    ra::Vector{Float64}      # Aerodynamic resistance (s/m)
+    u2m::Vector{Float64}     # Wind speed at 2m height (m/s)
+
+    # Energy balance components
+    RSN::Vector{Float64}     # Net shortwave radiation (W/m2)
+    RLN::Vector{Float64}     # Net longwave radiation (W/m2)
+    Rnet::Vector{Float64}    # Net radiation (W/m2)
+    lambda::Vector{Float64}  # Latent heat of vaporization (J/kg)
+    LE::Vector{Float64}      # Latent heat flux (W/m2)
+    H::Vector{Float64}       # Sensible heat flux (W/m2)
```
vers-w: Would be good to check/discuss what variables should be computed by hydromt_wflow and what is computed by wflow.

vers-w (commented):
Nice start with this new feature @michaelohanrahan ! 

I added some comments, main suggestion is to include this new feature as part of `LandHydrologySBM`, make use of variables and parameter fields in the model struct, to move some parameters or forcing to other structs (e.g. for shared land parameters) and to check/discuss what hydromt_wflow should compute.

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Thanks @vers-w for now I think we agree that wind speed (u2m) shortwave radiation (RS_in) and albedo will be added to the forcing and are therefore computed in preprocessing, soon after migrating to hydromt_wflow. A similar story with canopy height hc, that is now a static param. Ra, lambda, etc still remain calculated within the "LandSurfaceTemperatureModel"

Wflow/src/land_surface_temperature.jl
michaelohanrahan: For now DOY remains calculated in func update_timestep_land_surface_temperature, it is no longer a vector indeed

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Latitude now part of network

Wflow/src/land_surface_temperature.jl
michaelohanrahan: no longer static params, they are in forcing, though emissivity is not yet supported. Is there a way to support fallback values in forcing supplied? i.e. if albedo is not attached in forcing now it will not work, though one could feasibly proceed with a constant albedo for testing purposes

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,347 @@
+# after Devi Purnamasari et al. 2025 
+# https://doi.org/10.5194/hess-29-1483-2025
+#TODO: review comments that are too instructional, retain during dev.
+
+using Wflow
+using Dates
+import Wflow: AbstractLandModel, AtmosphericForcing, SbmSoilModel, Config
+using NCDatasets
+using Parameters
```
vers-w: These are not required here, and `using Wflow` only makes sense if this would be an external package (not within the module `Wflow`).

```suggestion
```

Wflow/src/forcing.jl
```diff
@@ -6,6 +6,12 @@
     potential_evaporation::Vector{Float64}
     # Temperature [ᵒC]
     temperature::Vector{Float64}
+    # Downward shortwave radiation [W m-2]
+    RS_in::Vector{Float64}
+    # Surface albedo [-]
+    albedo::Vector{Float64}
```
vers-w: Suggest to move `albedo` to struct `LandParameters` (src\domain.jl).

Wflow/src/forcing.jl
```diff
@@ -6,6 +6,12 @@
     potential_evaporation::Vector{Float64}
     # Temperature [ᵒC]
     temperature::Vector{Float64}
+    # Downward shortwave radiation [W m-2]
+    RS_in::Vector{Float64}
```
vers-w: Preference is to use lower case explicit names for variables in the code, e.g. here:  `shortwave_radiation_in` and `windspeed_2m`.

Wflow/src/forcing.jl
```diff
@@ -6,6 +6,12 @@
     potential_evaporation::Vector{Float64}
```
vers-w: Would be good to use default values for the struct fields, e.g.: `potential_evaporation::Vector{Float64} = Float64[]`.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,347 @@
...
+const dt = 86400.0     # timestep in seconds
+
+"Struct for storing LST model variables"
+@with_kw struct LandSurfaceTemperatureVariables
+    # Input forcing (updated each timestep)
+    RS_in::Vector{Float64}   # Downward shortwave radiation (W/m2)
+    Ta::Vector{Float64}      # Air temperature (K)
+    u2m::Vector{Float64}     # Wind speed at 2m height (m/s)
+    albedo::Vector{Float64}  # Surface albedo
+    #TODO: emissivity is a constant, not a variable
```
vers-w: These fields can be removed as these are now part of forcing/land parameters.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,347 @@
...
+    # emissivity::Vector{Float64} # Surface emissivity
+
+    #From LandHydrologySBM
+    ET_a::Vector{Float64}    # Actual evapotranspiration from SBM (mm/Δt)
```
vers-w: Including `ET_a` is not really required here, as it is already stored as part of the `SbmSoilModel`.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,347 @@
...
+
+    # Calculated variables
+    ra::Vector{Float64}      # Aerodynamic resistance (s/m)
+    RSN::Vector{Float64}     # Net shortwave radiation (W/m2)
+    LE::Vector{Float64}      # Latent heat flux (W/m2)
+    H::Vector{Float64}       # Sensible heat flux (W/m2)
+    RLN::Vector{Float64}     # Net longwave radiation (W/m2)
+    lambda::Vector{Float64}  # Latent heat of vaporization (J/kg)
+    Rnet::Vector{Float64}    # Net radiation (W/m2)
+    LST::Vector{Float64}     # Land surface temperature (K)
```
vers-w: Try to make use of lower case explicit names for variables as much as possible.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,347 @@
...
+        LST = fill(MISSING_VALUE, n),
+    )
+end
+
+"Struct for storing LST model parameters"
+@with_kw struct LandSurfaceTemperatureParameters
+    # Static parameters (from netCDF or defaults)
+    crop_height::Vector{Float64} # Crop height in m
+    latitude::Vector{Float64}  # Latitude of each grid cell (degrees)
+end
```
vers-w: As `crop_height` and `latitude` can be useful for other wflow models suggest to store `latitude` as part of struct `NetworkLand` in \src\network.jl and `crop_height` in struct `VegetationParameters` in \src\vegetation\parameters.jl.

Maybe replace `crop_height` by `canopy_height` (better term)?

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,347 @@
...
+    land_surface_temperature_model.variables.Ta .= atmospheric_forcing.temperature
+
+    land_surface_temperature_model.variables.RS_in .= atmospheric_forcing.RS_in
+
+    land_surface_temperature_model.variables.albedo .= atmospheric_forcing.albedo
+
+    land_surface_temperature_model.variables.u2m .= atmospheric_forcing.u2m
+
+    return nothing
+end
```
vers-w: This function is not required as it is done at the model level (`load_dynamic_input!(model)`).

```suggestion
```

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,347 @@
...
+    params = LandSurfaceTemperatureParameters(dataset, config, indices)
+    vars = LandSurfaceTemperatureVariables(n)
+
+    return LandSurfaceTemperatureModel(; parameters = params, variables = vars)
+end
+
+"""
+Update LST model for a single timestep
+"""
+function update_timestep_land_surface_temperature(
```
vers-w:
```suggestion
function update_land_surface_temperature(
```

Wflow/src/sbm_gwf_model.jl
```diff
@@ -72,6 +72,17 @@ function update!(model::AbstractModel{<:SbmGwfModel})
...
+    do_lst = get(config.model, "lst__flag", false)::Bool
+    if do_lst
+        update!(
+            land.land_surface_temperature,
+            land.soil,
+            land.atmospheric_forcing,
+            clock.time,
+            dt,
+        )
+    end
```
vers-w: Suggest to move this to sbm.jl (update function of `LandHydrologySBM`).

Wflow/src/sbm_model.jl
```diff
@@ -98,6 +98,18 @@ function update_until_recharge!(model::AbstractModel{<:SbmModel})
...
+    do_lst = get(config.model, "lst__flag", false)::Bool
+    if do_lst
+        update!(
+            land.land_surface_temperature,
+            land.soil,
+            land.atmospheric_forcing,
+            clock.time,
+            dt,
+        )
+    end
```
vers-w: Suggest to move this to sbm.jl (update function of `LandHydrologySBM`).

vers-w (commented):
Thanks for the changes @michaelohanrahan ! Some additional comments from my side.

Wflow/src/sbm.jl
```diff
@@ -32,6 +33,12 @@ function LandHydrologySBM(dataset::NCDataset, config::Config, domain::DomainLand
     else
         snow = NoSnowModel()
     end
+    do_lst = get(config.model, "lst__flag", false)::Bool
```
vers-w: Suggest to use a more explicit name for this model setting:
```suggestion
    do_lst = get(config.model, "land_surface_temperature__flag", false)::Bool
```

Wflow/src/standard_name.jl
```diff
@@ -203,6 +203,20 @@ const sbm_standard_name_map = Dict{String, NamedTuple}(
...
+    "land_surface__net_longwave_radiation" =>
+        (lens = @optic(_.land.land_surface_temperature.variables.RLN), unit = "W m-2"),
+    "land_surface__net_radiation" =>
+        (lens = @optic(_.land.land_surface_temperature.variables.Rnet), unit = "W m-2"),
+    "land_surface__latent_heat_flux" =>
+        (lens = @optic(_.land.land_surface_temperature.variables.LE), unit = "W m-2"),
+    "land_surface__sensible_heat_flux" =>
+        (lens = @optic(_.land.land_surface_temperature.variables.H), unit = "W m-2"),
+    "land_surface__aerodynamic_resistance" =>
+        (lens = @optic(_.land.land_surface_temperature.variables.ra), unit = "s m-1"),
```
vers-w: Some suggested improvements for these standard names, also including a quantity name.

```suggestion
    "land_surface_radiation~net~shortwave__energy_flux" =>
        (lens = @optic(_.land.land_surface_temperature.variables.RSN), unit = "W m-2"),
    "land_surface_radiation~net~longwave__energy_flux" =>
        (lens = @optic(_.land.land_surface_temperature.variables.RLN), unit = "W m-2"),
    "land_surface_radiation~net~total__energy_flux" =>
        (lens = @optic(_.land.land_surface_temperature.variables.Rnet), unit = "W m-2"),
    "land_surface_air_heat~latent~evaporation__energy_flux" =>
        (lens = @optic(_.land.land_surface_temperature.variables.LE), unit = "W m-2"),
    "land_surface_air_heat~net~sensible__energy_flux" =>
        (lens = @optic(_.land.land_surface_temperature.variables.H), unit = "W m-2"),
    "atmosphere_bottom_air__bulk_momentum_aerodynamic_resistance" =>
        (lens = @optic(_.land.land_surface_temperature.variables.ra), unit = "s m-1"),
```

Wflow/src/forcing.jl
michaelohanrahan: Thanks for the suggestion, changing should also allow for flexibility in providing default values

`
@with_kw struct LandSurfaceTemperatureVariables

    # Input forcing (updated each timestep)

    RS_in::Vector{Float64} = Float64[]   # Downward shortwave radiation (W/m2)

... etc.
`

Wflow/src/forcing.jl
michaelohanrahan: Adopted, thanks  e.g. sensible_heat_flux, net_radiation, net_shortwave_radiation and shortwave_radiation_in

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Indeed the constant for testing is now removed and handled in preprocessing before handling per timestep with atmospheric forcing

Wflow/src/land_surface_temperature.jl
michaelohanrahan: dt now accessed via clock

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Indeed, now corrected

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Made more explicit throughout land_surface_temperature.jl

Wflow/src/land_surface_temperature.jl
michaelohanrahan: NetworkLand updated, Vegetation params updated, canopy_height adopted :)

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Adopted, Thanks!

Wflow/src/land_surface_temperature.jl
michaelohanrahan: adopted

Wflow/src/sbm.jl
michaelohanrahan: adopted, thanks

Wflow/src/sbm_gwf_model.jl
michaelohanrahan: Moved, thanks 😄

Wflow/src/sbm_model.jl
michaelohanrahan: Adopted, thanks

Wflow/src/standard_name.jl
michaelohanrahan: adopted, thanks

Wflow/src/forcing.jl
michaelohanrahan: currently thinking about how to first go with albedo as part of forcing, otherwise fall back on static param/constant

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Attempted to implement the fallback mechanism commit @ [72a74e9](https://github.com/Deltares/Wflow.jl/pull/633/commits/72a74e9e771958c70e0900b5a1fc1ebf8377cd59)

Wflow/src/land_surface_temperature.jl
michaelohanrahan: now using `Clock`

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Still working on moving constants to land parameters

Wflow/src/forcing.jl
michaelohanrahan: fallback attempted with this commit [72a74e9](https://github.com/Deltares/Wflow.jl/pull/633/commits/72a74e9e771958c70e0900b5a1fc1ebf8377cd59)

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,272 @@
...
+    lat_rad = deg2rad(lat)
+    doy = dayofyear(clock.current_time)
+    decl = solar_declination(doy)
+    dist = relative_distance(doy)
+    sha = acos(-tan(lat_rad) * tan(decl))
+    Ra =
+        dist * gsc / π *
+        (cos(lat_rad) * cos(decl) * sin(sha) + sha * sin(lat_rad) * sin(decl))
+    return Ra
+end
```
vers-w: I would suggest to remove the computation of extraterrestrial radiation from wflow and let hydromt_wflow take care of this (input as forcing). And maybe also the net longwave radiation computation? This variable does not seem to rely on parameters in wflow that may vary through time (like `albedo` and net shortwave radiation).

Wflow/src/land_surface_temperature.jl
vers-w: Parameters can also vary in time (cyclic/forcing) as long as there is a mapping between the standard name and the internal model name, stored in \src\standard_name.jl. I have added the `albedo` parameter to the standard name mapping `Dict` and removed it from the forcing struct. so the code is a bit more simple/efficient.

Wflow/src/land_surface_temperature.jl
vers-w: If we make use of `emissivity` as a parameter, I think we only need surface downward longwave radiation flux as input forcing for the computation of net longwave radiation.

Wflow/src/standard_name.jl
```diff
@@ -203,6 +211,34 @@ const sbm_standard_name_map = Dict{String, NamedTuple}(
...
+        unit = "W m-2",
+    ),
+    "land_surface__sensible_heat_flux" => (
+        lens = @optic(_.land.land_surface_temperature.variables.sensible_heat_flux),
+        unit = "W m-2",
+    ),
+    "land_surface__aerodynamic_resistance" => (
+        lens = @optic(_.land.land_surface_temperature.variables.aerodynamic_resistance),
+        unit = "s m-1",
+    ),
```
vers-w: In a previous comment https://github.com/Deltares/Wflow.jl/pull/633#discussion_r2156720138, these suggestions were added but not adopted by selecting "Commit suggestion".

```suggestion
    "land_surface_radiation~net~shortwave__energy_flux" => (
        lens = @optic(_.land.land_surface_temperature.variables.net_shortwave_radiation),
        unit = "W m-2",
    ),
    "land_surface_radiation~net~longwave__energy_flux" => (
        lens = @optic(_.land.land_surface_temperature.variables.net_longwave_radiation),
        unit = "W m-2",
    ),
    "land_surface_radiation~net~total__energy_flux" => (
        lens = @optic(_.land.land_surface_temperature.variables.net_radiation),
        unit = "W m-2",
    ),
    "land_surface_air_heat~latent~evaporation__energy_flux" => (
        lens = @optic(_.land.land_surface_temperature.variables.latent_heat_flux),
        unit = "W m-2",
    ),
    "land_surface_air_heat~net~sensible__energy_flux" => (
        lens = @optic(_.land.land_surface_temperature.variables.sensible_heat_flux),
        unit = "W m-2",
    ),
    "atmosphere_bottom_air__bulk_momentum_aerodynamic_resistance" => (
        lens = @optic(_.land.land_surface_temperature.variables.aerodynamic_resistance),
        unit = "s m-1",
    ),
```

Wflow/src/standard_name.jl
```diff
@@ -12,10 +12,18 @@ const sbm_standard_name_map = Dict{String, NamedTuple}(
     ),
     "land_surface__evapotranspiration_volume_flux" =>
         (lens = @optic(_.land.soil.variables.actevap), unit = "mm dt-1"),
+    "land_surface__albedo" =>
+        (lens = @optic(_.domain.land.parameters.albedo), unit = "-"),
     "land_water~storage~total__depth" =>
         (lens = @optic(_.land.soil.variables.total_storage), unit = "mm"),
     "atmosphere_air__temperature" =>
         (lens = @optic(_.land.atmospheric_forcing.temperature), unit = "°C"),
+    "atmosphere_air__shortwave_radiation_in" => (
```
vers-w: Consistent with CSDMS standard name:
```suggestion
    "land_surface_radiation~incoming~shortwave__energy_flux" => (
```

Wflow/src/network.jl
```diff
@@ -89,6 +91,35 @@ function NetworkLand(dataset::NCDataset, config::Config, modelsettings::NamedTup
...
+        if length(lat_1d) < new_rows
+            # Pad with the last latitude value if needed
+            extended_lat = vcat(
+                extended_lat,
+                repeat([lat_1d[end]], new_rows - length(lat_1d), new_cols),
+            )
+        end
+        latitude_2d = extended_lat
+    end
+    latitude = latitude_2d[indices]
```
vers-w: When using the Allen et al. (1998) approach for computing net longwave radiation we decided to do this is part of pre-processing (e.g. hydromt_wflow) and not part of the Wflow kernel. The variable `latitude` is only required as part of this approach (to compute extraterrestrial radiation), so I think it is better to remove this part. We also do not have to deal with wflow models in meter-based projections.

Wflow/src/sbm.jl
```diff
@@ -78,11 +87,21 @@ function update!(
     domain::Domain,
     config::Config,
     dt::Float64,
+    clock,
```
vers-w: Suggest to remove `clock`. It is only required for the extraterrestrial radiation computation as part of the Allen et al. (1998) approach for computing net longwave radiation, and we decided to move this (computing net longwave radiation) to a pre-processing step (e.g. hydromt_wflow).

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Added to hydromt_wflow/wflow setup_landsurfacetemp: https://github.com/michaelohanrahan/hydromt_wflow/tree/feature/landsurfacetemp

Wflow/src/network.jl
michaelohanrahan: Adopted with the new preprocessing func setup_landsurfacetemp: https://github.com/michaelohanrahan/hydromt_wflow/tree/feature/landsurfacetemp

Wflow/src/sbm.jl
michaelohanrahan: done!

Wflow/src/standard_name.jl
michaelohanrahan: apologies for having not implemented the suggestions before they became outdated. Of course they are strong suggestions and I have now included them in the latest commit https://github.com/Deltares/Wflow.jl/pull/633/commits/d133977639c5b2b24a099452a5e6e511916d78c4

Wflow/src/land_surface_temperature.jl
michaelohanrahan: moved to src/utils.jl ... but I doubt this is the preferred placement. Suggestion?

vers-w: It seems these parameters are used locally (e.g. within a function), so I would suggest to define these locally as part of the associated functions. If we want to share these parameters between (many) different functions I would say it is preferred to store these for example as part of the land parameters struct, but for now I think including these parameters as part of a function is fine.

Wflow/src/vegetation/parameters.jl
```diff
@@ -29,6 +31,12 @@ function VegetationParameters(
     rootingdepth = ncread(dataset, config, lens; sel = indices, type = Float64)
     lens = lens_input_parameter(config, "vegetation__crop_factor")
     kc = ncread(dataset, config, lens; sel = indices, defaults = 1.0, type = Float64)
+
+    # Read canopy height
+    lens = lens_input_parameter(config, "vegetation__canopy_height")
```
vers-w: Same as standard name used in Sediment model:
```suggestion
    lens = lens_input_parameter(config, "vegetation_canopy__height")
```

Wflow/src/vegetation/parameters.jl
```diff
@@ -29,6 +31,12 @@ function VegetationParameters(
...
+    canopy_height =
+        ncread(dataset, config, lens; sel = indices, defaults = 0.12, type = Float64)
```
vers-w: Any reason to use a value of 0.12 here? The default value as part of the Sediment model is 0.5.

Wflow/src/network.jl
vers-w: Sounds good, probably good to remove this part then. Or do you still need it for some testing?

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,260 @@
...
+    zm_ref::Float64 = 2.0,
+    k::Float64 = 0.41,
+)
+    # Handle measurement height below canopy
+    if z_measured < canopy_height
+        z_measured = canopy_height
+    end
+
+    # Ensure minimum canopy height using Allen FAO reference
+    canopy_height = max(canopy_height, 0.12)
```
vers-w: You could also only call this function if `canopy_height` is >= 0.12. I think that is a bit cleaner.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,260 @@
...
+        z0m_ratio = 1.23e-1
+        z0h_ratio = 0.1
+
+    elseif canopy_height >= 1
+        z0m_ratio = 1.23e-1 * (canopy_height / 2.0) #z0m increases with canopy height
+        ref_h = 0.12
+        dh_ratio = 2.0 / 3.0
+        z0h_ratio = 0.2
+    else
+        error("Canopy height $canopy_height is not supported")
```
vers-w: Wondering if this error message is really needed? 

What height would not be supported as it is already set to a minimum of 0.12? If this function is only called for cells with the correct range of `canopy_height` (see also comment above) I think catching this error is not needed. Raising an error in a function that is called in a for loop and for each timestep can get computationally expensive and also for a user it may get overwhelming, depending how many times the error is raised.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,260 @@
...
+    # Handle measurement height below canopy
+    if z_measured < canopy_height
+        z_measured = canopy_height
+    end
+
+    # Ensure minimum canopy height using Allen FAO reference
+    canopy_height = max(canopy_height, 0.12)
+
+    # Simplified empirical d/h ratios and roughness height adjustments
+    if canopy_height < 1
```
vers-w: Would make use of floating-point type, 1.0, also below.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,260 @@
...
+        #AWRA05
+        f_h = log((813 / canopy_height) - 5.45)
+        ku = 0.305 / (f_h * (f_h + 2.3))
+        ga = ku * wind_speed_ref
+        ra = (1 / ga) / 10.0
+    end
+
+    # Ensure positive aerodynamic resistance
+    if ra <= 0
+        error("Aerodynamic resistance is negative: $ra")
```
vers-w: Instead of throwing an error I would recommend to solve this with a minimum value for `ra` or if possible resolve this before `ra` is computed.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,260 @@
...
+function compute_land_surface_temperature(
+    sensible_heat_flux::Float64,
+    aerodynamic_resistance::Float64,
+    air_temperature::Float64,
+)
+    # Handle invalid aerodynamic resistance values
+    if isnan(aerodynamic_resistance) ||
+       isinf(aerodynamic_resistance) ||
+       aerodynamic_resistance <= 0
+        @warn "Invalid aerodynamic resistance: $aerodynamic_resistance, using air temperature as land surface temperature"
```
vers-w: As mentioned above probably better to leave out log messages in functions that are applied in a for loop for each cell and at each timestep. Would recommend to give it a minimum value (or not including cells that are invalid for these kind of computations).

Wflow/src/network.jl
michaelohanrahan: Thankfully this is now cleaned up by removal in [f6a234f](https://github.com/Deltares/Wflow.jl/pull/633/commits/f6a234f784684e10490906b3b357816e53be0f04) , hence marking as resolved :)

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Resolved now with local, function level definition with [c2bef5b](https://github.com/Deltares/Wflow.jl/pull/633/commits/c2bef5b7cbdc6e9e7cadde5c2ce86337c0a62020)

Turns out I had carried forward some unnecessary definitions from the beginning, so only air density and heat capacity was required after moving net radiation to preprocessing

Wflow/src/vegetation/parameters.jl
michaelohanrahan: yes i use this to default to the FAO (Allen '98) grass reference as default

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Agreed! A remnant of testing that should now be removed. It was indeed an overwhelming error at one point, so I completely understand the suggestion.

Wflow/src/land_surface_temperature.jl
michaelohanrahan: I don't think it is possible any longer to calculate a negative with the floor in wind speed, and the handling of the taller canopies. I would be hesitant to assuming a defualt value for ra as the end result is quite sensitive and the combination of vars that lead to negative values are not so predictable. I will retain for testing and add a TODO to remove after I am confident the issue is eliminated.

Wflow/src/land_surface_temperature.jl
michaelohanrahan: is it allowed to return a nan (or nodata) value in the case of an invalid iteration?

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Done!

Wflow/src/vegetation/parameters.jl
vers-w: OK, sounds good, thanks for the explanation!

Wflow/src/land_surface_temperature.jl
vers-w: I would not recommend to return a NaN value, but that is not the case here, right? If `aerodynamic_resistance` is invalid then `air_temperature` is returned. I would suggest to ensure  `aerodynamic_resistance` has always a valid value, is that possible?

Wflow/src/land_surface_temperature.jl
vers-w: I don't think this has changed?

michaelohanrahan: instead of mutating inside the ` wind_and_aero_resistance ` , favouring line 80 insertion         `canopy_height = max(vegetation_parameters.canopy_height[i], 0.12)` to achieve the same effect.

Wflow/src/land_surface_temperature.jl
michaelohanrahan: solved: return `max(ra, 1.0)`

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Solved with the return max(ra,1.0) with above comment. `@warn` also removed from sensible heat flux

Wflow/src/land_surface_temperature.jl
vers-w: Do we maybe have a literature references for this minimum value for `ra`?

Wflow/src/land_surface_temperature.jl
vers-w: See comment above.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,212 @@
...
+        latent_heat_of_vaporization * WATER_DENSITY * actual_evapotranspiration_ms
+    return latent_heat_flux
+end
+
+""" 'sensible heat flux' :: H  ≈ RNet - LE - G"""
+function compute_sensible_heat_flux(net_radiation::Float64, latent_heat_flux::Float64)
+    # Handle NaN values in net radiation
+    if isnan(net_radiation)
+        return 0.0
+    end
```
vers-w: Note that we already check for missing values in forcing data when updating the forcing data each time step and throw an error is this happens. So, I think this check is not required.

Wflow/src/land_surface_temperature.jl
vers-w: If I understand correctly, we want to do these computations only if we have vegetation (transpiration) in a grid cell, right? If that's the case, is it not cleaner to apply this only to grid cells with vegetation/transpiration and have a default value (air temperature?) for the other cells.

Wflow/src/land_surface_temperature.jl
michaelohanrahan: in terms of reference grass the formulations simplifies to 208/{wind speed at reference height} m/s so the values can range realistically from a max case: ra (wind 0.1 m/s) = 2080 m/s down to a realistic minimum case: ra (wind 20m/s) = 10,4 m/s ... that floor was devised to avoid incompatible values rather than with the physical range in mind. We can of course revise that

Wflow/src/land_surface_temperature.jl
michaelohanrahan: As discussed not necessarily, but indeed there is a simplification in this approach that gives some vegetated state (via canopy height) to otherwise a bare soil. Should this be raised as separate, future issue or PR?

Wflow/src/land_surface_temperature.jl
michaelohanrahan: updated minimum value in 187a1031a

Wflow/src/land_surface_temperature.jl
michaelohanrahan: resolved with updated max(ra, 10)?  187a1031a

Wflow/src/land_surface_temperature.jl
michaelohanrahan: removed in 7b16eda2e

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,203 @@
...
+Forest cap avoids z - d < 0 issues
+Alternative aerodynamic conductance calculation from AWRA05
+https://www.researchgate.net/publication/233757155_AWRA_Technical_Report_3_Landscape_Model_version_05_Technical_Description
+
+"""
+function wind_and_aero_resistance(
+    wind_speed_measured::Float64,
+    z_measured::Float64,
+    canopy_height::Float64;
+    zm_ref::Float64 = 2.0, # reference height for wind speed (m)
```
vers-w: I understand that the reference height `zm_ref` of 2 m is based on the FAO approach, related to the Penman-Monteith equation for a reference surface. 

Based on the work of Zink et al. (2018) and other literature I think we can deviate from this reference height of 2 m. My recommendation would be to use a reference height of ~2 m above the canopy (add 2 m to the average canopy height and round to nearest meter). Additionally, I would recommend to stick to the aerodynamic resistance equation based on Thom’s equation (Thom, 1975) and to remove the alternative aerodynamic conductance calculation from AWRA05. And for now just use the default values: 

```
dh_ratio = 2.0 / 3.0
z0m_ratio = 0.123
z0h_ratio = 0.1
```

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,203 @@
...
+    k::Float64 = 0.41, # von Kármán constant
+)
+    # Handle measurement height below canopy
+    if z_measured < canopy_height
+        z_measured = canopy_height
```
vers-w: Maybe better to check if `z_measured` is smaller than the zero plane displacement height, and than move it to `canopy_height`?

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,203 @@
...
+    end
+
+    # Calculate canopy height and roughness height
+    d = dh_ratio * ref_h
+    z0m = z0m_ratio * ref_h
+    z0h = z0h_ratio * z0m  # Canopy sublayer roughness
+
+    # Wind speed conversion to reference height (2m) for consistency with FAO-56
+    wind_speed_ref =
+        max(wind_speed_measured * (log(zm_ref / z0m) / log(z_measured / z0m)), 0.5)
```
vers-w: I noticed that a minimum wind speed of 0.5 is mentioned in Purnamasari et al. (2025) based on Allen et al., (1998) but cannot find it. Do you know where this is mentioned in Allen et al., (1998)?

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,203 @@
...
+
+    if canopy_height < 1.0
+        # Aerodynamic resistance using reference height
+        # ra = (log((zm_ref - d) / z0m)) * (log((zm_ref - d) / z0h)) / (k^2 * wind_speed_ref)
+        ra = log((zm_ref - d) / z0m) / (k^2 * wind_speed_ref)
```
vers-w: Is this not missing the inclusion of `z0h` in the equation?

Wflow/src/land_surface_temperature.jl
vers-w: Thought a bit more about using canopy height to estimate roughness lengths and the zero plane displacement height in wflow. In the end, I think it is cleaner to provide the roughness lengths and zero plane displacement height as parameters to wflow, HydroMT-Wflow can extract these parameters from data, even from data that already provides these parameters, see for example this [paper](https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2023EA003027). We also don't have to deal then with the magically canopy heights for non-vegetated classes in wflow.

vers-w: Now using zero‐plane displacement height `d0`, aerodynamic roughness length for momentum transfer `zm0`, aerodynamic roughness length for heat transfer `zh0` and surface "skin" layer height above ground `skin_layer_height` as input parameters (also allowed dynamically).

Wflow/src/land_surface_temperature.jl
vers-w: I did remove the check of a minimum value, I don't think it is required.

Wflow/src/land_surface_temperature.jl
vers-w: Did remove this check, I think it is not required.

Wflow/src/land_surface_temperature.jl
vers-w: The reference height is now set to `skin_layer_height` + ~2.0 m. Additionally, the resistance equation is now only based on Thom’s equation (Thom, 1975), including the part for heat transfer.

Wflow/src/land_surface_temperature.jl
vers-w: Removed the check for `z_measured`. The expectation is that the provided wind speed (and measurement height) is consistent with `d0` and `z0m`. Need to decide if validation of this input is needed (e.g. measurement height should be higher than `d0 + z0m`).

Wflow/src/land_surface_temperature.jl
vers-w: Need to decide if we want to stick with this minimum wind speed value of 0.5 m/s.

Wflow/src/land_surface_temperature.jl
vers-w: Now using Thom’s equation.

Wflow/src/land_surface_temperature.jl
vers-w: Did find it in Allen et al. (1998), in Chapter 3 - Meteorological data, so fine to use this value.

Wflow/src/vegetation/parameters.jl
```diff
@@ -16,6 +16,8 @@
     rooting_depth::Vector{Float64} = []
     # Crop coefficient Kc [-]
     crop_coefficient::Vector{Float64} = []
+    # Canopy height [m]
```
michaelohanrahan: Read but not used I think, the `ra` calculation uses a skin_layer that I am not yet familiar with. Maybe left over from the old version? Maybe useful to retain for filtering out shaded areas / tall canopy from output.

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,174 @@
...
+    n = length(land_surface_temperature_model.variables.land_surface_temperature)
+
+    for i in 1:n
+        # Use pre-calculated net radiation from forcing
+        land_surface_temperature_model.variables.latent_heat_of_vaporization[i] =
+            compute_latent_heat_of_vaporization(atmospheric_forcing.temperature[i])
+
+        land_surface_temperature_model.variables.latent_heat_flux[i] =
+            compute_latent_heat_flux(
+                atmospheric_forcing.temperature[i],
```
michaelohanrahan: Wflow/src/soil/soil.jl:35 
`actual_evapotranspiration::Vector{Float64} = fill(MISSING_VALUE, n)`

Should it match the long form?

Wflow/src/domain.jl
michaelohanrahan, line 215: should this also have the new z0m z0h d0 and skin_layer_height?

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,174 @@
...
+function compute_sensible_heat_flux(net_radiation::Float64, latent_heat_flux::Float64)
+    # estimation of ground heat flux can be improved, currently estimated with a fixed
+    # linear coefficient (0.1) applied to net_radiation
+    ground_heat_flux = 0.1 * net_radiation
+    sensible_heat_flux = net_radiation - latent_heat_flux - ground_heat_flux
+    return sensible_heat_flux
+end
+
+"""
+Compute aerodynamic resistance `ra` of the land surface "skin" (vegetation canopy, soil,
```
michaelohanrahan, line 124: edit for canopy

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,174 @@
...
+function compute_aerodynamic_resistance(
+    wind_speed::Float64,
+    z_measured::Float64,
+    skin_layer_height::Float64,
+    d0::Float64,
+    z0m::Float64,
+    z0h::Float64,
+)
+    # set reference height (~2.0 m above surface skin layer height).
+    zm_ref = round(skin_layer_height + 2.0)
```
michaelohanrahan, line 144: is it meant to be round?

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,174 @@
...
+    # wind speed at reference height and constrained to be greater than 0.5 m/s to consider
+    # vapour exchange on the surface induced by air buoyancy and layer instability effects.
+    min_wind_speed = 0.5
+    wind_speed_ref = max(
+        wind_speed * (log((zm_ref-d0) / z0m) / log((z_measured-d0) / z0m)),
+        min_wind_speed,
+    )
+
+    # compute aerodynamic resistance based on Thom's equation.
+    ra = (log((zm_ref-d0)/z0m)*log((zm_ref-d0)/z0h)) / (VON_KARMAN^2 * wind_speed_ref)
```
michaelohanrahan: docstring warns of d0+z0m and measurement height... is that legacy? Or should there be a check for negatives / nans in the log ..

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,174 @@
...
+end
+
+"Update land surface temperature model for a single timestep."
+function update_land_surface_temperature!(
+    land_surface_temperature_model::LandSurfaceTemperatureModel,
+    soil_model::SbmSoilModel,
+    atmospheric_forcing::AtmosphericForcing,
+    parameters::LandParameters,
+    wind_measurement_height::Float64,
+    dt::Float64,
```
michaelohanrahan: i dont see dt used here, legacy?

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,174 @@
...
+    air_temperature_degc = from_SI(air_temperature, ABSOLUTE_DEGREES)
+    return (2501.0 - 2.375 * air_temperature_degc) * 1000.0 # J/kg converted from kJ/kg
+end
+
+"Compute latent heat flux"
+function compute_latent_heat_flux(
+    air_temperature::Float64,
+    actual_evapotranspiration::Float64,
+)
+    latent_heat_of_vaporization = compute_latent_heat_of_vaporization(air_temperature)
```
michaelohanrahan, line 108: Called twice! also on 51

Wflow/src/land_surface_temperature.jl
```diff
@@ -0,0 +1,174 @@
...
+    latent_heat_flux =
+        latent_heat_of_vaporization * WATER_DENSITY * actual_evapotranspiration
+    return latent_heat_flux
+end
+
+"Compute sensible heat flux"
+function compute_sensible_heat_flux(net_radiation::Float64, latent_heat_flux::Float64)
+    # estimation of ground heat flux can be improved, currently estimated with a fixed
+    # linear coefficient (0.1) applied to net_radiation
+    ground_heat_flux = 0.1 * net_radiation
```
michaelohanrahan, line 118: Not constants? Also Air density? Asking for my own curiosity

docs/model_docs/land/landhydrology_sbm.qmd
```diff
@@ -1221,7 +1222,93 @@ continuity equation of the local inertial model. For reservoir locations surface
...
+and $\rho_\mathrm{water}$ is the density of water ($\SI{1000.0}{kg\ m^{-3}}$). The soil heat
+flux $G$ is estimated with a fixed linear coefficient (0.1) applied to net radiation
+$R_\mathrm{n}$. With $R_\mathrm{n}$ as input, sensible heat flux $H$ can be estimated using
+the energy balance equation. LST can be computed as follows:
+
+$$
+    T_\mathrm{s} = \frac{H r_\mathrm{air}}{\rho_\mathrm{air} c_\mathrm{p}} + T_\mathrm{air}
+$$
+
+where $\SIb{r_\mathrm{air}}{s\ m^{-1}}$ is aerodynamic conductance, $\SIb{c_\mathrm{p}}{J\
```
michaelohanrahan, line 1265: conductance or resistance?

docs/model_docs/land/landhydrology_sbm.qmd
```diff
@@ -1221,7 +1222,93 @@ continuity equation of the local inertial model. For reservoir locations surface
...
+
+The aerodynamic resistance $r_\mathrm{air}$ that controls the vapour and heat transfer is
+computed based on Thom’s equation (Thom, 1975):
+
+$$
+    r_\mathrm{air} = \frac{\ln(\frac{z_\mathrm{m} - d}{z_\mathrm{0m}}) \ln(\frac{z_\mathrm{h} - d}{z_\mathrm{0h}})}{k^2 u_\mathrm{z}}
+$$
+
+where $\SIb{d}{m}$ is the zero plane displacement height, $\SIb{z_\mathrm{m}}{m}$ is the
+height of wind measurement, $\SIb{z_\mathrm{h}}{m}$ is the height of humidity measurement,
```
michaelohanrahan, line 1277: code has a conversion to reference height I see, best to mention?

docs/model_docs/model_settings.qmd
```diff
@@ -33,6 +33,8 @@ The Tables in this subsection show the available models settings for the model t
 | `soil_infiltration_reduction__flag` | Enable reduction factor applied to the soil infiltration capacity | - | `false` |
 | `soil_layer__thickness` | Thickness of soil layers | mm | [100, 300, 800] |
 | `saturated_hydraulic_conductivity_profile` | Saturated hydraulic conductivity depth profile for `SBM` `soil` model (also used for kinematic wave lateral subsurface flow) | - | "exponential" |
+| `land_surface_temperature__flag` | Include land surface temperature modelling | - | `false` |
+| `land_surface_wind__speed_reference_height` | Wind speed (forcing) reference height | m | - |
```
michaelohanrahan, line 37: 10m?

michaelohanrahan:
Hi @vers-w my compliments first of all on the nice work, and my thanks for taking over a project I never had time to finish. Much appreciated! 

Since you took this over, I've left comments rather than pushing changes. If anything is unclear or confusing do ask me to expand on what are mostly short requests for clarification.

Wflow/src/vegetation/parameters.jl
vers-w: Nice catch, see commit https://github.com/Deltares/Wflow.jl/pull/633/commits/088f0ccc480353001afc593ceec5a0617f04d6f2. I did remove it as we are indeed not using this parameter as part of the `SBM` concept. LST is often referred to as `skin` temperature, hence the use of `skin_layer_height`. This more generic parameter replaces `canopy_height`.

Wflow/src/land_surface_temperature.jl
vers-w: Yes, it should, thanks for spotting this!

Wflow/src/domain.jl
vers-w: I would say for the `sediment` model these parameters are not required.

Wflow/src/land_surface_temperature.jl
vers-w: Yes, the more generic term "skin" now replaces `canopy`. Is that what you mean?

Wflow/src/vegetation/parameters.jl
michaelohanrahan: Nice thanks for the clarification

Wflow/src/land_surface_temperature.jl
vers-w: Good catch, removed `dt` in https://github.com/Deltares/Wflow.jl/pull/633/commits/03f3ad30009b6f1b29c32e1ae448cadc254d779d.

Wflow/src/land_surface_temperature.jl
vers-w: I did remove the rounding. It is now set at `skin_layer_height` with a minimum of 2.0 m (consistent with FAO standard height and I think also the previous implementation), see commit https://github.com/Deltares/Wflow.jl/pull/633/commits/63ccbc6c90e0b18f4e224b517269041eb5a96aaf.

Wflow/src/land_surface_temperature.jl
vers-w: Nice catch, see commit https://github.com/Deltares/Wflow.jl/pull/633/commits/c4d6f1e4f6320825e3c19f19381d92ca15cc3bd3.

Wflow/src/land_surface_temperature.jl
vers-w: Yes, good point, I did add air density and specific heat capacity as constants. For this linear coefficient for ground heat flux I am leaning toward not using a constant, we want to improve this method, the coefficient is only used here, and probably it is a bit too specific for a constant.

docs/model_docs/land/landhydrology_sbm.qmd
vers-w: Correct, should be resistance.

docs/model_docs/land/landhydrology_sbm.qmd
vers-w: Yes, I agree, this is described from L 1298, did update it, related to https://github.com/Deltares/Wflow.jl/pull/633#discussion_r4014073540.

Wflow/src/land_surface_temperature.jl
vers-w: Yes, I think this is related to https://github.com/Deltares/Wflow.jl/pull/633#discussion_r3638991852. Is validation of `d0`, `z0m` and the measurement height of wind speed required in Wflow? Personally, I think it makes more sense to handle this in hydromt_wflow.

The other comment in the docstring is about the expectation that wind speed is consistent with actual surface terrain, see also for example [era5-wind-speeds](https://forum.ecmwf.int/t/interpretation-of-era5-wind-speeds-regarding-surface-roughness/1259).

Wflow/src/standard_name/standard_name_domain.jl
```diff
@@ -31,4 +31,32 @@ const domain_standard_name_map = OrderedDict{String, ParameterMetadata}(
         description = "Water allocation area ids",
         tags = [:generic_input_map],
     ),
+    "atmosphere_bottom_air_flow__roughness_length" => ParameterMetadata(;
+        lens = @optic(_.domain.land.parameters.z0m),
+        unit = Unit(; m = 1),
+        description = "Land surface aerodynamic roughness length for momentum transfer",
+        allow_dynamic_input = true,
+        tags = [:lst_input],
```
michaelohanrahan, line 39:
```suggestion
        tags = [:lst_input],
        default = 0.01476 # 0.123 x 0.12
```

Wflow/src/standard_name/standard_name_domain.jl
```diff
@@ -31,4 +31,32 @@ const domain_standard_name_map = OrderedDict{String, ParameterMetadata}(
...
+    ),
+    "atmosphere_bottom_air_heat_flow__roughness_length" => ParameterMetadata(;
+        lens = @optic(_.domain.land.parameters.z0h),
+        unit = Unit(; m = 1),
+        description = "Land surface aerodynamic roughness length for heat transfer",
+        allow_dynamic_input = true,
+        tags = [:lst_input],
```
michaelohanrahan, line 46:
```suggestion
        tags = [:lst_input],
        default = 0.001476 # 0.1 x z0m
```

Wflow/src/standard_name/standard_name_domain.jl
```diff
@@ -31,4 +31,32 @@ const domain_standard_name_map = OrderedDict{String, ParameterMetadata}(
...
+    ),
+    "atmosphere_bottom_air_flow__zero_plane_displacement_length" => ParameterMetadata(;
+        lens = @optic(_.domain.land.parameters.d0),
+        unit = Unit(; m = 1),
+        description = "Zero plane displacement height",
+        allow_dynamic_input = true,
+        tags = [:lst_input],
```
michaelohanrahan, line 53:
```suggestion
        tags = [:lst_input],
        default = 0.08,   # 2/3 x 0.12
```

Wflow/src/standard_name/standard_name_domain.jl
```diff
@@ -31,4 +31,32 @@ const domain_standard_name_map = OrderedDict{String, ParameterMetadata}(
...
+    ),
+    "atmosphere_bottom_air__above_ground_height" => ParameterMetadata(;
+        lens = @optic(_.domain.land.parameters.skin_layer_height),
+        unit = Unit(; m = 1),
+        description = "Land surface \"skin\" layer height above the ground",
+        allow_dynamic_input = true,
+        tags = [:lst_input],
```
michaelohanrahan, line 60:
```suggestion
        tags = [:lst_input],
        default = 0.12   # FAO-56 reference crop height
```

Wflow/src/land_surface_temperature.jl
michaelohanrahan: Agreed on thom's

Issue #648, opened 2025-06-30
Node to source edge mapping has ragged array; not GPU compatible


BSchilperoort:
The `EdgesAtNode` struct contains a mapping from nodes to source edges. This is represented in a ragged array. For example;

```jl
mapping = Vector([
    Vector([1]),
    Vector([2, 3]),
    Vector([4, 5, 6]),
])
```
This nested vector structure is not compatible with (julia's/KernelAbstractions) GPU arrays\*. If the array cannot be copied to GPU we cannot use it inside a kernel. 
To make it GPU compatible, we can use a 2D array, with `0` as padding value (as julia has no 0 index);

```jl
mapping = Array([
    1 0 0
    2 3 0
    4 5 6
])
```

Seeing as this will be an array of size (nodes, \<node with maximum number of source edges\>) the array shouldn't become much bigger than its current size.

The [`sum_at` function](https://github.com/Deltares/Wflow.jl/blob/ecce360265299f8df0f0d3a7a874f5b6e031d753/Wflow/src/utils.jl#L457) needs to be modified to accommodate this, by skipping any `0` values.

```jl
function sum_at(A, inds)
    v = zero(eltype(A))
    for i in inds
        if i != 0
            v += A[i]
        end
    end
    return v
end
```

\* Technically CUDA and ROCm support ragged tensors, but this is not available in Julia's generic GPU packages.

Issue #647, opened 2025-06-30
Skipping modules with `!isnothing(...)` incompatible with GPU kernels


BSchilperoort:
If inside a GPU kernel definition Wflow modules are skipped using `if !isnothing`, the kernel cannot be compiled.

For example;
```jl
@kernel function update(model::LocalInertialRiverFlow)
    if !isnothing(model.floodplain)
        model.floodplain.variables.q0 .= model.floodplain.variables.q
    end
    ...
end
```

The GPU compiler tries to compile this kernel, but seeing as `model.floodplain.variables` does not exist when `model.floodplain = nothing`, the compiler fails. To fix this issue, instead of defining `model.floodplain` as `nothing`, it should be a valid floodplain struct with 0-element arrays instead.

Note that this is only needed when the `if !isnothing` statement is _inside_ the kernel (i.e. inertial river routing). I am not sure if this needs to change anywhere else but in `surface_local_inertial.jl`.

BSchilperoort:
I think the main problem is that the compiler tries to compile the code, including everything in the `if !isnothing` block.

As the code in the `if !isnothing` block tries to access variables of `nothing`, the compiler fails to compile the kernel.

```jl
@kernel function update(model::LocalInertialRiverFlow)
    # compiler will compile everything in if statement, as it will only
    #    evaluate the result of this if statement on kernel execution:
    if !isnothing(model.floodplain)
        model.floodplain.variables.q .= 0.0f0 # compiler fails on this line
    end
end
```

Normal julia code doesn't suffer from this as it will only compile the `model.floodplain.variables.q .= ... ` line once the if statement resolves to True. However, the entire `@kernel function` block has to be compiled to GPU-executable code, including any branching.

This would also go wrong if you would branch on two different model concepts, e.g. an `ExampleModelWithVariableQ` struct and an `ExampleModelWithVariableX` struct. The following kernel would be impossible to compile if the structs don't share the accessed variables;

```jl
@kernel function update(model::LocalIntertialRiverFlow)
    if model.floodplain isa ExampleModelWithVariableQ
        model.floodplain.q .= 0.0
    elseif model.floodplain isa ExampleModelWithVariableX
        model.floodplain.x .= 0.0
    end
end
```

This kind of branching needs to be done _outside_ GPU kernels; in native Julia code. It cannot be done inside GPU kernels.

Issue #646, opened 2025-06-30
Add configurable floating point precision


BSchilperoort:
In #534 (related issue #535) the `const Float` definition was removed, leaving no way to configure the floating point precision of Wflow. However, to be able to more efficiently run the model on GPUs, the floating point precision needs to be configurable, as; 

1. Consumer grade GPUs have very bad fp64 performance (or can't run on 64-bit at all, like Apple Metal).
2. Professional grade GPUs (e.g. NVIDIA A100) still have a 2x better performance on fp32.

In our fork of Wflow at https://github.com/ESiWACE-S1/Wflow.jl I added the `const Float` definition back in, but from the discussion in #535 it seems that this is not the preferred way to do this in the future.

BSchilperoort:
In #650 I have added parametric types to the LocalInertialRiverFlow structs. This allows for them to hold arbitrary floating point precision data. 

Upon loading the data from netCDF, the config is already available, so the types can be set from there, e.g.;
```jl
ncread(dataset, config, lens; sel = ..., type = config.precision)
```

This will make most of the data ending up in routines/functions will be using he specified precision.

For numbers defined inside functions, we can get the element type of a (floating point) variable and cast every number;

```jl
function example_function!(model::ExampleStruct)
    fp = eltype(model.example_var)  # Float32 or Float64
    model.example_var .+= fp(5.0)
end
```
and for a GPU kernel;
```jl
@kernel function example_function!(model::ExampleStruct)
    fp = eltype(model.example_var)  # Float32 or Float64
    i = @index(Global)
    model.example_var[i] += fp(5.0)
end
```

This seems to compile to the same instructions as defining the float with `5.0f0` or `Float32(5.0)`, and I could discern no performance difference at all.

vers-w:
I was wondering if something like this is also possible:

```julia
function example_function!(model::ExampleStruct{FT}) where {FT}
    model.example_var .+= FT(5.0)
end
```
I think it is a bit cleaner. Or otherwise have the float type as a function arg?

Maybe Julia packages part of the CliMA project (e.g. Oceananigans.jl or ClimaLand.jl) could provide some insight here as these packages also make use of GPU.

SouthEndMusic:
See https://github.com/Deltares/Wflow.jl/pull/912 for a possible approach.

PR #650, opened 2025-07-02
Preparations to run intertial river routing on GPU

BSchilperoort:
## Issue addressed
Related to #646 - allows for parametric types of vectors in structs. The parametric types are also `<:AbstractArray` to allow for moving the data to GPU arrays instead of Julia native arrays.

To do;

- [ ] Modify function types to allow for `<:AbstractFloat` instead of `Float64`
- [ ] Refactor river network structs to allow for moving to GPU (related to #648)
- [ ] Refactor structs so that modules cannot be `Nothing` (see #647)
- [ ] Allow for configuration of GPU backend through config
- [ ] Implement GPU Kernel for river routing

## Explanation
[Adapt](github.com/JuliaGPU/Adapt.jl) has been added as a dependency, to allow for moving arrays to GPU (e.g. `CuArray` or `ROCArray`)

## Checklist

### DIFF
--- a/Wflow/src/routing/surface_local_inertial.jl
+++ b/Wflow/src/routing/surface_local_inertial.jl
@@ -1,21 +1,26 @@
 "Struct for storing local inertial river flow model parameters"
-@with_kw struct LocalInertialRiverFlowParameters
-    n::Int                                  # number of cells [-]
-    ne::Int                                 # number of edges [-]
-    active_n::Vector{Int}                   # active nodes [-]
-    active_e::Vector{Int}                   # active edges [-]
-    g::Float64                              # acceleration due to gravity [m s⁻²]
-    froude_limit::Bool                      # if true a check is performed if froude number > 1.0 (algorithm is modified) [-]
-    h_thresh::Float64                       # depth threshold for calculating flow [m]
-    zb::Vector{Float64}                     # river bed elevation [m]
-    zb_max::Vector{Float64}                 # maximum channel bed elevation [m]
-    bankfull_storage::Vector{Float64}       # bankfull storage [m³]
-    bankfull_depth::Vector{Float64}         # bankfull depth [m]
-    mannings_n_sq::Vector{Float64}          # Manning's roughness squared at edge [(s m-1/3)2]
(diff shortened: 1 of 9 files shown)


Wflow/src/routing/surface_local_inertial.jl
```diff
@@ -1028,13 +1035,17 @@ Floodplain `storage` is a function of `depth` (flood depth intervals). Based on
 cumulative floodplain `storage` a floodplain profile as a function of `flood_depth` is
 derived with floodplain area `a` (cumulative) and wetted perimeter radius `p` (cumulative).
 """
-@with_kw struct FloodPlainProfile{N}
-    depth::Vector{Float64}        # Flood depth [m]
-    storage::Array{Float64, 2}    # Flood storage (cumulative) [m³]
-    width::Array{Float64, 2}      # Flood width [m]
-    a::Array{Float64, 2}          # Flow area (cumulative) [m²]
-    p::Array{Float64, 2}          # Wetted perimeter (cumulative) [m]
+@with_kw struct FloodPlainProfile{
```
BSchilperoort, line 1038: Not sure what the {N} parameter was supposed to do here. Adding parametric types here broke that feature.

It also did not seem to be used anywhere, except for in the `profile = FloodPlainProfile{n_depths}(; ...` definition below. Removing this did not break any tests.

Wflow/test/reservoir.jl
```diff
@@ -159,15 +159,15 @@ end
         Wflow.read_hq_csv(joinpath(datadir, "input", "reservoir_hq_2.csv")),
     ])
     res_params = Wflow.ReservoirParameters(;
-        area = [200_000_000],
+        area = [2.0e8],
```
BSchilperoort, line 162: Tests were passing a `Vector{Int}`, which did not fail before despite the type of `area` being defined as `Vector{Float64}`. I guess some magical type casting was performed automatically?

Defining the type of `area` as `<:AbstractArray{<:AbstractFloat}` broke this however, so the arrays in this test had to be modified to have float elements.

PR #652, opened 2025-07-03
Support external inflows at reservoir locations

### LINKED ISSUE #464: Allow water abstractions from lake/reservoir
### Feature type

Adding new functionality

### Improvement Description

Allow water abstractions from lake and reservoir for example for irrigation purposes.
Without the water demands concept, wflow can allow for external water inflow/abstractions but only from the river water volume and not the reservoir. It would be good to allow for this, especially if the data is available (for example for reservoir operators).

It is possible to add or remove external inflows by adding an inflow variable to the forcing file (together with precipitation, temperature and potential evapotranspiration). This variable contains 0 by default and then inflow or abstractions for the concerned grid cell locations. However, these inflow or abstractions are currently taken out of the river water volume rather than the reservoir. A minor change is then needed so that abstractions in reservoir cells are taken out from the reservoir volume rather than from the river. Alternatively, reservoir inflow could become an independent variable from the river inflow and CSV or NetCDF-CF_TIMESERIES input format for these timeseries could be considered in order to simplify processing.

To check in refinements: how abstractions from reservoirs are calculated if any when the water demands concept is on and can it work if local data is available.

### Additional Context

Timely and accurate forecast is crucial to predict and respond to flood or drought risk. For reservoir operators or basins that contains reservoirs, adapting reservoir operations in time might help reduce the impact of drought or floods. During our work with Sunwater, we noticed that in forecast mode, it is important that reservoir models, here Wflow, can reflect accurately reservoir operations such as:

- Modification of the release outflow and/or rating curves.
- Account for varying abstractions from the reservoirs (e.g. for irrigation).

If not considered properly, there is a risk that the model starts differing from the observations, reducing the accuracy of the forecast.

This issue is part of these improvements.

vers-w:
## Issue addressed
Fixes #464


Wflow/src/demand/water_demand.jl
```diff
@@ -642,10 +643,19 @@ function surface_water_allocation_area!(
...
                 k = inds_reservoir[j]
-                available_surfacewater[j] = reservoir.variables.storage[k] * 0.98 # limit available reservoir storage
-                sw_available += available_surfacewater[j]
+                external_inflow = reservoir.boundary_conditions.external_inflow[k]
+                if external_inflow < 0.0
+                    available_volume = reservoir.variables.storage[k] * 0.98
+                    max_res_abstraction = min(external_inflow * dt, available_volume)
+                    available_volume = max(available_volume - max_res_abstraction, 0.0)
+                else
+                    available_volume = reservoir.variables.storage[k] * 0.98
```
JoostBuitink, line 655: Just to be sure, in case of positive inflows, this is handled in the routing processes, right?

Wflow/src/demand/water_demand.jl
```diff
@@ -642,10 +643,19 @@ function surface_water_allocation_area!(
...
+                # for reservoir locations use reservoir storage, check for abstraction
+                # through external negative inflow first and adjust available volume.
                 k = inds_reservoir[j]
-                available_surfacewater[j] = reservoir.variables.storage[k] * 0.98 # limit available reservoir storage
-                sw_available += available_surfacewater[j]
+                external_inflow = reservoir.boundary_conditions.external_inflow[k]
+                if external_inflow < 0.0
+                    available_volume = reservoir.variables.storage[k] * 0.98
+                    max_res_abstraction = min(external_inflow * dt, available_volume)
+                    available_volume = max(available_volume - max_res_abstraction, 0.0)
```
JoostBuitink, line 653: This would mean that external abstractions get higher priority than the "internal demands", which seem perfectly fine to me. I thought we had that also mentioned in the docs, but I cannot find it. Perhaps good to explicitly mention this in the docs as well (in case it is indeed missing currently ;) )

Wflow/test/run_sbm.jl
```diff
@@ -206,41 +206,73 @@ Wflow.run_timestep!(model)
...
+(; actual_external_abstraction_av, external_inflow) =
+    model.routing.river_flow.boundary_conditions
 (; q_av) = model.routing.river_flow.variables
-@testset "river negative inflow" begin
+@testset "river external negative inflow" begin
     Wflow.run_timestep!(model)
-    @test actual_external_abstraction_av[44] ≈ -2.8679304138607975
-    @test q_av[44] ≈ 0.1843871721222499
+    @test actual_external_abstraction_av[44] ≈ 1.5977234689598292
+    @test q_av[44] ≈ 1.4346381048266201
```
JoostBuitink, line 268: How is it possible that q_av has increased by so much? Is see that less water is extracted, but what code change affects this?

JoostBuitink (approved):
Looks good to me, just a couple of small questions!

Wflow/src/demand/water_demand.jl
vers-w: Yes, this check is only to restrict the available volume during the water demand computation, negative external inflow has priority. Both positive and negative inflows are handled in the routing process (reservoir location as part of the river routing).

Wflow/src/demand/water_demand.jl
vers-w: Good point, I have added this info to the docs.

Wflow/test/run_sbm.jl
vers-w: Yes, good question! 

The approach for limiting the amount of river water that can be abstracted is now the same for local inertial and kinematic wave routing, and based only on river storage (this was already the case for local inertial routing). This is a bit more conservative, which is fine I think, and the advantage is that the same approach is now used for both routing options.

PR #656, opened 2025-07-14
Dispatch lateral snow transport function

### LINKED ISSUE #651: snow and snow_gravitional_transport flags
`snow__flag = false` requires `snow_gravitional_transport__flag = false` otherwise it throws an `“type NoSnowModel has no field variables”` error. It makes sense that we cannot have ``snow_gravitional_transport`` (flag = true) when snow modelling is not included (snow__flag = false) but we should decide whether we want to throw an error. Note that this was not the case in old versions (at least in v0.7.3) where we could have at the same time ``masswasting = true`` and ``snow = false``. 

This can cause an error when converting an old toml to v1 with hydromt (https://deltares.github.io/hydromt_wflow/latest/_examples/upgrade_to_wflow_v1.html). If we want to keep the error in wflow might be good to check whether ``snow=false`` in ``upgrade_to_v1_wflow`` and then write the ``snow_gravitional_transport__flag`` accordingly (i.e. false)

vers-w:
## Issue addressed
Fixes #651 

Dispatch lateral snow transport on `snow` model type. This is consistent with simulating snow dynamics and avoids running lateral snow transport when snow modelling is not included (`snow__flag = false`) while snow transport is included (`snow_gravitional_transport__flag = true`) and as a result the following `type NoSnowModel has no field variables` error is not thrown.

### DIFF
--- a/Wflow/src/routing/routing_process.jl
+++ b/Wflow/src/routing/routing_process.jl
@@ -365,17 +365,22 @@ function accucapacityflux(material, network, capacity)
 end
 
 """
-    lateral_snow_transport!(snow, snowwater, slope, network)
+    lateral_snow_transport!(snow, slope, network)
 
-Lateral snow transport. Transports snow downhill. Mutates `snow` and `snowwater`.
+Lateral snow transport. Transports snow downhill. Mutates `snow_storage` and `snow_water` of
+a `snow` model.
 """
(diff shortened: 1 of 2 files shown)


JoostBuitink (approved):
Nice and clean solution!

PR #660, opened 2025-07-15
Fix RutterInterception

### LINKED ISSUE #659: Bug in `RutterInterceptionModel` (has no `vegetation_parameter_set`)
### Wflow version checks

### Reproducible Example

When running a model on an hourly timestep, I get the error message that the `RutterInterceptionModel.parameters `has no `vegetation_parameter_set` , see the lines below:

https://github.com/Deltares/Wflow.jl/blob/cf377e7043f35d6a074648b60b94b67e81301756/Wflow/src/vegetation/canopy.jl#L111-L112

### Current behaviour

The way how the RutterInterceptionModel is slightly different from the GashInterceptionModel, which causes this issue. 

### Desired behaviour

Have a working interception model for both Gash and Rutter options. My current solution is to setup the RutterInterceptionModel in the same way (see linked PR)

### Additional Context

_No response_

JoostBuitink:
## Issue addressed
Fixes #659 

## Explanation

## Checklist

### DIFF
--- a/Wflow/src/vegetation/canopy.jl
+++ b/Wflow/src/vegetation/canopy.jl
@@ -68,7 +68,7 @@ function update!(model::GashInterceptionModel, atmospheric_forcing::AtmosphericF
     e_r = model.parameters.e_r
     n = length(precipitation)
     if !isnothing(leaf_area_index)
-        update_canopy_parameters!(model)
+        update_canopy_parameters!(model.parameters.vegetation_parameter_set)
         threaded_foreach(1:n; basesize = 1000) do i
             canopyfraction = 1.0 - canopygapfraction[i]
             ewet = canopyfraction * potential_evaporation[i] * kc[i]
@@ -113,7 +113,7 @@ function update!(model::RutterInterceptionModel, atmospheric_forcing::Atmospheri
         model.variables
     (; precipitation, potential_evaporation) = atmospheric_forcing
     if !isnothing(leaf_area_index)
-        update_canopy_parameters!(model)
+        update_canopy_parameters!(model.parameters)
     end
     n = length(precipitation)
     threaded_foreach(1:n; basesize = 1000) do i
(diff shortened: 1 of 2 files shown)


vers-w:
Thanks for spotting this!
 
An alternative solution and I think a bit cleaner is to unpack differently in the `update!` function of the `RutterInterceptionModel`. In this case an additional `struct RutterParameters` is not required (including possible additional interception models that only make use of `VegetationParameters`). Also a change to the `function update_canopy_parameters!` is then needed (dispatch on `vegetation_parameter_set`).

Wflow/src/vegetation/canopy.jl
```diff
@@ -131,16 +130,33 @@ function update!(model::RutterInterceptionModel, atmospheric_forcing::Atmospheri
...
+        storage_wood,
+        kext,
+        storage_specific_leaf,
+        canopygapfraction,
+        cmax,
+    )
+end
+
+"Update canopy parameters `cmax` and `canopygapfraction` based on `leaf_area_index` for a single timestep"
+function update_canopy_parameters!(model::AbstractInterceptionModel)
```
vers-w: An alternative approach is to dispatch here on `VegetationParameters`, and adapt the function calls in lines 71 and 116. It is a bit shorter, also for possible additional interception models.

docs/changelog.qmd
```diff
@@ -11,7 +11,8 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 - Introduced in v1.0.0-rc1, fixed unit conversion model parameter `kv_0` to `kh_0` for
-  sub-daily model timestep. 
+  sub-daily model timestep.
+- Fix Rutter interception by correctly extracting the vegetation parameters.
```
vers-w: Maybe also add here when this bug was introduced (v1.0.0-rc1)? 

For the official 1.0.0 release we can then easily remove these from the changelog.

vers-w (approved):
LGTM (made two small suggestions)!

docs/changelog.qmd
JoostBuitink: Good suggestion, I have added this!

Wflow/src/vegetation/canopy.jl
JoostBuitink: That is indeed a much cleaner solution! I went for this implementation now!

PR #662, opened 2025-07-17
Fix unit conversion `kv_0` to `kh_0`

### LINKED ISSUE #661: Unit `kh_0` not correct for sub-daily model timestep
@JoostBuitink reported large differences in model results for the Meuse at an hourly timestep between v0.8.1 and v1.0.0-rc1. These large differences do not occur when running the model at a daily timestep. 

This seems to be caused by an incorrect conversion of `kv_0`, the vertical hydraulic conductivity [mm Δt⁻¹] at the soil surface.

vers-w:
This was not correct for a sub-daily model timestep.

## Issue addressed
Fixes #661


JoostBuitink (approved):
Thanks for fixing this!

PR #669, opened 2025-07-30
SBOM generation

### LINKED ISSUE #630: Automate SBOM Generation and Include in End-User ZIP
We need to ensure that a Software Bill of Materials (SBOM) is:

- Automatically generated whenever manifest.toml is updated.
- Included in the ZIP file that is distributed to end users.

- [ ]  Set up a script or tool to generate an SBOM 
- [ ]  Trigger SBOM generation automatically on manifest changes.
- [ ]  Modify the build pipeline to include the SBOM in the final ZIP.
- [ ]  Validate that the SBOM is up to date and complete.

savente93:
## Issue addressed
Fixes #630 

## Explanation
 TBC

## Checklist

### DIFF
--- a/.github/workflows/update-sbom.yml
+++ b/.github/workflows/update-sbom.yml
@@ -1,0 +1,34 @@
+name: Update SBOM
+on:
+  push:
+      branches:
+        - 'main'
+      paths:
+        - 'docs/Manifest.toml'
+        - 'server/Manifest.toml'
+        - 'Wflow/Manifest.toml'
+  # on demand
+  workflow_dispatch:
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ssh-key: ${{ secrets.SSH_PRIVATE_KEY }}
+      - uses: prefix-dev/setup-pixi@v0.8.14
+        with:
+          pixi-version: "latest"
+      - name: Install Julia and update SBOM file
+        run: |
+          pixi run install-julia
+          pixi run generate-sbom
+        env:
+          JULIA_PKG_PRECOMPILE_AUTO: "0"
+      - uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.CI_PR_PAT }}
+          branch: update/sbom
+          title: Update SBOM
+          commit-message: "Update SBOM"
+          author: "GitHub <noreply@github.com>"
--- a/Wflow/src/sbom-generation.jl
+++ b/Wflow/src/sbom-generation.jl
@@ -1,0 +1,30 @@
+using PkgToSoftwareBOM
+using UUIDs
+using Pkg
+
+Pkg.activate("Wflow")
+
+wflowLisence = SpdxLicenseExpressionV2("MIT")
+organization = SpdxCreatorV2("Organization", "Deltares", "software@deltares.nl")
+packageInstructions = spdxPackageInstructions(;
+    spdxfile_toexclude = ["Wflow.spdx.json"],
+    originator = organization,
+    declaredLicense = wflowLisence,
+    copyright = "Copyright (c) 2025 Deltares <software@deltares.nl>",
+    name = "Wflow",
+)
+
+dependencies = Pkg.project().dependencies;
+spdxData = spdxCreationData(;
+    Name = "Wflow.jl",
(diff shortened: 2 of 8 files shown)


savente93:
I'm getting a little confused by all the different projects. If you are in the office tomorrow, would you mind us just sitting down and figuring out? i feel like it'll be a very short change if we sit together

Wflow/src/sbom-generation.jl
JoostBuitink, line 1: I think this file can be removed right?

.github/workflows/update-sbom.yml
```diff
@@ -0,0 +1,34 @@
+name: Update SBOM
+on:
+  push:
+      branches:
+        - 'master'
+      paths:
+        - 'docs/Manifest.toml'
+        - 'server/Manifest.toml'
```
JoostBuitink: I think these two can be ignored, as the SBOM is only related to Wflow itself

docs/Manifest.toml
JoostBuitink, line 1: These changes can be reverted I think, as the project.toml is no longer changed

utils/sbom/Project.toml
```diff
@@ -0,0 +1,6 @@
+[deps]
+PkgToSoftwareBOM = "6254a0f9-6143-4104-aa2e-fd339a2830a6"
+Wflow = "d48b7d99-76e7-47ae-b1d5-ff0c1cf9a818"
```
JoostBuitink, line 3: Not that it really matters a lot, but does wflow need to be a dependency here? Since the environment that contains wflow is activated in the sbom-generation.jl script

pixi.toml
```diff
@@ -49,6 +50,12 @@ quarto-check = { cmd = "quarto check all" }
 quarto-render = { cmd = "julia --project=docs --eval 'using Pkg; Pkg.build(\"IJulia\")' && quarto render docs --to html --execute" }
 quarto-preview = { cmd = "quarto preview docs" }
 
+[tasks.generate-sbom]
+cmd = "julia --project=utils/sbom --check-bounds=yes utils/sbom/sbom-generation.jl"
+depends-on = ["initialize-julia"]
```
JoostBuitink: I think it is important there that both the `utils/sbom` and the `wflow` are initialized. I like the solution in PR #625 that uses an `arg` to initialize the different enviroments (see https://github.com/Deltares/Wflow.jl/pull/625/files#diff-b092682f0ad15a352f9fa2b0a67c992454b586275ee76b74a5839ab9afdf9894)

.github/workflows/update-sbom.yml
aslowwriter: Good point

utils/sbom/Project.toml
aslowwriter: I'm unsure, doesn't the dependency need to be there to be able to activate it? THis will also be done in CI, so that's why I thought it was necessary but perhaps not. What do you think?

pixi.toml
```diff
@@ -15,43 +15,60 @@ install-julia = "juliaup add 1.11.3 && juliaup override set 1.11.3"
...
+		"utils/sbom",
+	] },
+	{ "task" = "instantiate-julia", args = [
+		"Wflow",
+	] },
+]
+cmd = "julia --project -- sbom-generation.jl"
+cwd = "utils/sbom"
+inputs = ["Wflow/Project.toml"]
+outputs = ["Wflow.spdx.json"]
```
JoostBuitink: I am not too big of a pixi expert, but if I run this task, the Wflow.spdx.json is placed in the utils/sbom folder, probably due to the cwd argument. Or perhaps I am doing something incorrect! If not, would the option below be a better alternative?

```suggestion
cmd = "julia --project=utils/sbom utils/sbom/sbom-generation.jl"
inputs = ["Wflow/Project.toml"]
outputs = ["Wflow.spdx.json"]
```

aslowwriter: good point, yes I think you're correct! thanks

utils/sbom/Project.toml
JoostBuitink: You are activating a Julia project that contains Wflow as a dependency. I tried disabling it (that is, removing Wflow from the SBOM project), but got an error I didn't understand :P, so keeping is like this is fine with me!

PR #670, opened 2025-07-30
feat: copy all lisence files during app creation

### LINKED ISSUE #631: Automate Inclusion of All OSS Licenses in Build Output
We must ensure that all OSS licenses (direct and transitive) are automatically included in the build output.

aslowwriter:
## Issue addressed
Fixes #631 

## Explanation
Added some code to the `add_metadata` functionality in creating binaries. It will look at all the dependencies in the `wflow_cli` environment, and copy any file that is named license (with optional extension and case insensitive) and copy them into a folder of the output called `dep_licenses`. Because it is added to the existing functionality, any time it is built the licenses will automatically be included. It only includes the licenses of the version that are actually used in the environment to avoid adding too much bloat to the distribution artifacts. 

## Checklist

### DIFF
--- a/build/create_binaries/add_metadata.jl
+++ b/build/create_binaries/add_metadata.jl
@@ -81,6 +81,7 @@ function add_metadata(project_dir, license_file, output_dir, git_repo)
     ctx = PackageCompiler.create_pkg_context(project_dir)
     license_dir = joinpath(output_dir, "dep_licenses") 
     mkpath(license_dir)
+
     for (uuid,pkg_entry) in ctx.env.manifest.deps
 	if isnothing(pkg_entry.tree_hash)
 	    # seems as though stdlib packages don't have a tree_sha. as there doesn't seem to be 
@@ -88,11 +89,11 @@ function add_metadata(project_dir, license_file, output_dir, git_repo)
 	    continue 
 	end
 	install_path = Pkg.Operations.find_installed(pkg_entry.name, uuid, pkg_entry.tree_hash)
-	files = readdir(install_path, join=true)
-	license_file_index =findfirst(x-> isfile(x) && !isnothing(match(r"(?i).*/license.*\b$",x)), files) 
-	if !isnothing(license_file_index)
-	    license_file_path = files[license_file_index] 
-	    cp(license_file_path, joinpath(license_dir,pkg_entry.name), force=true)
+
+	license = LicenseCheck.find_license(install_path)
+	if !isnothing(license) 
+	    license_file_path = joinpath(install_path,license.license_filename)
+	    cp(license_file_path , joinpath(license_dir,pkg_entry.name), force=true)
 	end
     end
 end
--- a/build/create_binaries/Manifest.toml
+++ b/build/create_binaries/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.11.3"
 manifest_format = "2.0"
-project_hash = "ea2528725d34650b23f042d15e2cf1ec29a55822"
+project_hash = "3d9fa91f1a60d4d39e72c3ce91d0447242ed5033"
 
 [[deps.AbstractTrees]]
 git-tree-sha1 = "2d9c9a55f9c93e8887ad391fbae72f8ef55e1177"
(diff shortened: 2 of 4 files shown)


build/create_binaries/add_metadata.jl
```diff
@@ -72,4 +76,23 @@ function add_metadata(project_dir, license_file, output_dir, git_repo)
...
+    mkpath(license_dir)
+    for (uuid,pkg_entry) in ctx.env.manifest.deps
+	if isnothing(pkg_entry.tree_hash)
+	    # seems as though stdlib packages don't have a tree_sha. as there doesn't seem to be 
+	    # a different way of detecting those, I'm going to assume that is characteristic 
+	    continue 
+	end
+	install_path = Pkg.Operations.find_installed(pkg_entry.name, uuid, pkg_entry.tree_hash)
+	files = readdir(install_path, join=true)
+	license_file_index =findfirst(x-> isfile(x) && !isnothing(match(r"(?i).*/license.*\b$",x)), readdir(install_path, join=true))
```
visr: Have you considered using `find_license` from https://github.com/ericphanson/LicenseCheck.jl?
Note also you can use your `files` variable instead of repeating `readdir`.

build/create_binaries/add_metadata.jl
```diff
@@ -72,4 +76,23 @@ function add_metadata(project_dir, license_file, output_dir, git_repo)
...
     end
+
+    # collect lisences of all dependencies
+    ctx = PackageCompiler.create_pkg_context(project_dir)
+    license_dir = joinpath(output_dir, "dep_licenses") 
+    mkpath(license_dir)
+    for (uuid,pkg_entry) in ctx.env.manifest.deps
+	if isnothing(pkg_entry.tree_hash)
+	    # seems as though stdlib packages don't have a tree_sha. as there doesn't seem to be 
+	    # a different way of detecting those, I'm going to assume that is characteristic
```
JoostBuitink, line 88: Related to this, there is this file https://github.com/JuliaLang/julia/blob/master/THIRDPARTY.md

And I think @deltamarnix opened an issue at PkgToSoftwareBOM.jl to see if it could be modified to also cover these libraries, but I don't think there has been any progress

build/create_binaries/add_metadata.jl
aslowwriter: using `files` instead of `readdir` is a good one I'll modify that. As for the `find_license` I didn't know that existed, but given that it's a repo with 15 commits the last of which happened over 3 years ago I'm not super stoked about taking it on as a dependency. I'll defer to the maintainers for that decision though

build/create_binaries/add_metadata.jl
aslowwriter: I'm not entirely sure how your comment relates to this PR. Does it mean you want to wait for that external update or is there something I should modify?

build/create_binaries/add_metadata.jl
visr: Sounds good. The maintainer is active in the community. Not particularly on this repo, but I believe that is mostly because it just works.

aslowwriter: Okay, if the maintainer is still active within the community that's less of a concern. I'll leave it up to maintainers to decide whether I should move over to that or if this is okay

visr: I see now that LicenseCheck is actually a dependency of PkgToSoftwareBOM, so it is already an indirect dependency.
May also help with being consistent with the SBOM.

https://github.com/Deltares/Wflow.jl/pull/669/files#diff-54a801f523c8f767061bbcd8c8da625a8d8ce3354a1513a7d674badb77fa14e9R1096-R1100

aslowwriter: Quickly discussed this with Joost, and it does check more files than my implementation does, so we'll use that instead. I've updated the PR

build/create_binaries/add_metadata.jl
aslowwriter: Discussed this briefly over a call, and we've decided to just go ahead and merge this. When anything new is necessary on the THIRDPARTY front, we can revisit but that way we can move on for the time being

JoostBuitink (approved):
Thanks! LGTM!

visr:
I think I gave you rights some days ago?

savente93:
image

PR #683, opened 2025-08-29
Support use of observed reservoir outflow

### LINKED ISSUE #465: Fixing reservoir outflow
### Feature type

Adding new functionality

### Improvement Description

The current SimpleReservoir model does not easily allow to control reservoir outflows even when data is available probably because we also need to tune parameters such as the min and max target volumes or the shape of the s curve used.
In case reservoir outflow data is available directly, it would be nice to have a more simple reservoir concept where the outflow is not computed from demand or rating curves but directly uses timeseries provided by the user (still checking if enough water is available in the reservoir volume).

In cases where the rating curve or Wflow model approach is not precise enough to correctly predict outflows, it could be useful to use user-defined outflow timeseries directly in Wflow.

To allow for this, a new reservoir outflow variable should be added to the Wflow model and if there is enough water available in the reservoir, the reservoir outflow is then equal to the user-defined outflow rather than calculated using a rating curve. Similar to RTC-Tools, it would be useful if a check or switch could be implemented so that a user could easily either use the normal rating curve concept or the user-defined outflows if provided in operational use (ie easy to switch from one concept to the other in the toml file).

As for the inflow and abstractions issue #464, it could be useful if the format of the user-defined reservoir outflows timeseries could be in CSV or NetCDF-CF_TIMESERIES format rather than NetDCF_CF_GRID to avoid extra processing.

### Additional Context

Timely and accurate forecast is crucial to predict and respond to flood or drought risk. For reservoir operators or basins that contains reservoirs, adapting reservoir operations in time might help reduce the impact of drought or floods. During our work with Sunwater, we noticed that in forecast mode, it is important that reservoir models, here Wflow, can reflect accurately reservoir operations such as:

Modification of the release outflow and/or rating curves.
Account for varying abstractions from the reservoirs (e.g. for irrigation).
If not considered properly, there is a risk that the model starts differing from the observations, reducing the accuracy of the forecast.

This issue is part of these improvements.

vers-w:
## Issue addressed
Fixes #465 

Use of netCDF gridded input (cyclic or forcing) of observed reservoir outflow is supported. If a reservoir location has a non-missing value for observed reservoir outflow, the reservoir is updated using this outflow limited by the reservoir storage, otherwise (missing value) the reservoir model is used defined by the rating and storage curve types. To simplify the code, two output reservoir variables (simple reservoir operational rules) have been removed, these were only defined at the internal routing timestep and not part of model output (standard names).

Supporting input through NetCDF-CF_TIMESERIES format can be part of another PR (e.g. also for external inflow variable).

@hboisgon : no need to check the code in detail, but probably good to check if this PR covers the requested functionality in #465.

### DIFF
--- a/Wflow/src/routing/reservoir.jl
+++ b/Wflow/src/routing/reservoir.jl
@@ -557,3 +557,12 @@ function log_message_observed_outflow(reservoir::Reservoir)
     end
     return msg
 end
+
+"Check if observed outflow is used for reservoirs"
+function using_observed_outflow(reservoir::Union{Reservoir, Nothing}, config::Config)
+    par = "reservoir_water~outgoing~observed__volume_flow_rate"
+    check =
+        !isnothing(reservoir) &&
+        (haskey(config.input.forcing, par) || haskey(config.input.cyclic, par))
+    return check
+end
--- a/Wflow/src/routing/surface_routing.jl
+++ b/Wflow/src/routing/surface_routing.jl
@@ -37,8 +37,7 @@ function surface_routing!(model)
         (; overland_flow, subsurface_flow),
         domain.reservoir.network,
     )
-    # 
-    if !isnothing(reservoir)
+    if using_observed_outflow(reservoir, config)
         @debug log_message_observed_outflow(reservoir)
     end
     # update river flow
@@ -57,7 +56,7 @@ Run surface routing (land and river) for a model type that contains the routing 
 function surface_routing!(
     model::Model{R},
 ) where {R <: Routing{<:LocalInertialOverlandFlow, <:LocalInertialRiverFlow}}
-    (; routing, land, domain, clock) = model
+    (; routing, land, domain, clock, config) = model
     (; soil, runoff) = land
     (; overland_flow, river_flow, subsurface_flow) = routing
     (; reservoir) = river_flow.boundary_conditions
@@ -72,7 +71,7 @@ function surface_routing!(
     # update reservoir inflow (subsurface flow), inflow from river and overland flow is
     # added within the river and overland routing schemes
     update_inflow!(reservoir, river_flow, subsurface_flow, domain.reservoir.network)
-    if !isnothing(reservoir)
+    if using_observed_outflow(reservoir, config)
         @debug log_message_observed_outflow(reservoir)
     end
     # update overland and river flow


hboisgon (approved):
Looks good in terms of functionality, tanks @vers-w ! Also nice that is very easy to switch between observed outflow or "modelled" outlfow when data is just missing

Wflow/src/routing/reservoir.jl
```diff
@@ -474,6 +469,22 @@ function update_reservoir_free_weir(
...
+    i::Int,
+    boundary_vars::NamedTuple,
+    dt::Float64,
+)
+    res_v = model.variables
+    (; precipitation, actevap, inflow) = boundary_vars
+
+    storage_input = (res_v.storage[i] + precipitation - actevap) / dt + inflow
+    outflow = min(res_v.outflow_obs[i], storage_input)
+    storage = (storage_input - outflow) * dt
```
JoostBuitink, line 486: Perhaps we do need to check of the storage does not exceed the maximum storage? Or do we want essentially ignore the parameters of the reservoirs when outflow data is used?

JoostBuitink: Also thinking related to this, maybe it it good to log for which reservoir outflow data is used?

JoostBuitink (commented):
Thanks for this! Looks good to me, just two small remarks!

Wflow/src/routing/reservoir.jl
vers-w: > Perhaps we do need to check of the storage does not exceed the maximum storage? Or do we want essentially ignore the parameters of the reservoirs when outflow data is used?

This check is of course possible. And perhaps we should then also check for the `threshold` (below that level outflow is zero)? Anyway, from the description in issue #465 I thought we basically want to ignore reservoir parameters. @hboisgon : what do you think, what are the client expectations?

vers-w: > Also thinking related to this, maybe it it good to log for which reservoir outflow data is used?

Yes, I think a good suggestion. I guess at `debug` level? 
We do not store the external reservoir IDs, but I would say for a useful log message we should add this info.

hboisgon: > > Perhaps we do need to check of the storage does not exceed the maximum storage? Or do we want essentially ignore the parameters of the reservoirs when outflow data is used?
> 
> This check is of course possible. And perhaps we should then also check for the `threshold` (below that level outflow is zero)? Anyway, from the description in issue #465 I thought we basically want to ignore reservoir parameters. @hboisgon : what do you think, what are the client expectations?

I thought about it but I think it depends on what you want. 

If what you want is that outflows are accurate for the dowsntream part of the model then I would not want to correct it (also because reservoir volumes may not be well estimated by Wflow). So in practice I don't think you want to force spilling when max level is reached, but you could have it as an option? I think the way we implemented it, that it is only turned on if observed are not NaN and if it is, it goes back to "classic" implementation means we do not really want Wflow to give another result.

Maybe if we start to have a low threshold and high threshold, this could also become just a new simplified storage curve function (5)? So with dynamic demand (that should not have NA) and getting rid of the sigmoid and max release?

Anyway I think for now the current implementation is what we want but I will check with the client what they think about it.

vers-w: Based on this I think it is fine to stay with the current implementation.

JoostBuitink: Agreed!

Wflow/src/routing/reservoir.jl
```diff
@@ -533,4 +544,16 @@ function update!(
...
+"Generate log message for using observed outflow at reservoir locations"
+function log_message_observed_outflow(reservoir::Reservoir)
+    not_nan = findall(x -> !isnan(x), reservoir.variables.outflow_obs)
+    if isempty(not_nan)
+        msg = "Observed outflow is not used for any reservoir location"
+    else
+        ids = reservoir.parameters.id[not_nan]
+        msg = "Observed outflow is used for reservoir location ids $ids"
+    end
+    return msg
```
JoostBuitink, line 558: Nice, good addition!

JoostBuitink: Maybe to add, would it be an idea to only log this debug message when  `"reservoir_water~outgoing~observed__volume_flow_rate"` is defined? Because if it is not specified, there is also no real value in stating that no observed outflow is used for every timestep

JoostBuitink (approved):
Nice! Just one suggestion on the logging message.

Wflow/src/routing/reservoir.jl
vers-w: Good suggestion, see commit https://github.com/Deltares/Wflow.jl/pull/683/commits/7471f9851cfa69f7367a1a63b8c8e661406b57cf.

JoostBuitink: Perfect! Then I think it is good to merge!

PR #687, opened 2025-09-03
Add date time info to error message when forcing/cyclic data is missing.

### LINKED ISSUE #684: MIssing data message when forcings data has a missing elements does not mention timesteps
### Feature type

### Improvement Description

In update_forcing there is a check for missing data which is great. The problem is that we make forcing data assuming nothing is missing and it is very hard to find that missing value in a bit nc file; It would be helpfull if the message includes the date and time of the missing value(s)

_No response_

### Additional Context

_No response_

vers-w:
## Issue addressed
Fixes #684

### DIFF
--- a/Wflow/test/run_sbm.jl
+++ b/Wflow/test/run_sbm.jl
@@ -202,7 +202,7 @@ Wflow.run_timestep!(model)
     land = model.land
     @test land.atmospheric_forcing.precipitation[2] / precip[2] ≈ 2.0
     @test (land.atmospheric_forcing.potential_evaporation[100] - 1.50) / evap[100] ≈ 3.0
-    @test land.vegetation_parameters.leaf_area_index[100] / lai[100] ≈ 1.5999999403953562
+    @test land.vegetation_parameters.leaf_area_index[100] / lai[100] ≈ 1.6 atol=1e-7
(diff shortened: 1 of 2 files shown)


Wflow/test/run_sbm.jl
```diff
@@ -202,7 +202,7 @@ Wflow.run_timestep!(model)
     land = model.land
     @test land.atmospheric_forcing.precipitation[2] / precip[2] ≈ 2.0
     @test (land.atmospheric_forcing.potential_evaporation[100] - 1.50) / evap[100] ≈ 3.0
-    @test land.vegetation_parameters.leaf_area_index[100] / lai[100] ≈ 1.6
+    @test land.vegetation_parameters.leaf_area_index[100] / lai[100] ≈ 1.5999999403953562
```
JoostBuitink: Would something like this be better? Because we do expect the value to be 1.6, but I guess it is slightly different due to floating point

```suggestion
    @test land.vegetation_parameters.leaf_area_index[100] / lai[100] ≈ 1.6 atol=1e-7
```

JoostBuitink (approved):
Looks good to me! Just one comment/suggestion!

Wflow/test/run_sbm.jl
vers-w: Good suggestion, it is indeed related to floating point.

Issue #688, opened 2025-09-05
Rename snow flag TOML setting


vers-w:
`snow_gravitional_transport__flag` to `snow_gravitational_transport__flag`, see also https://github.com/Deltares/hydromt_wflow/issues/542

Issue #243, opened 2023-02-17
consider using Configurations.jl for Config handling


visr:
In https://github.com/Deltares/Ribasim.jl/pull/66 @evetion improved the TOML config handling, which includes some validation. It is worth considersing using a similar approach in Wflow. This could simplify code and catch user errors earlier.

Currently we create our own struct, and overload a bunch of functions to have things like `config.option` work:

https://github.com/Deltares/Wflow.jl/blob/v0.6.2/src/io.jl#L36-L39

That could be turned into something like this:

https://github.com/Deltares/Ribasim.jl/blob/3c8f9c48fae613e1e45d3dfffc97e861a95324a6/src/config.jl#L32-L59

verseve:
Thanks for sharing! Seems indeed quite nice, we will look into this.

evetion:
If it has no priority, I can look into it in a month or two.

verseve:
Hi @evetion, do you have time to look into this? If not, @JoostBuitink has interest to look into this issue (with some guidance/help from your side).

evetion:
@JoostBuitink Feel free to pick it up, let me know if there are questions.

JoostBuitink:
see main issue #351

PR #689, opened 2025-09-08
Parse TOML to a nested struct

SouthEndMusic:
Fixes https://github.com/Deltares/Wflow.jl/issues/243
Fixes https://github.com/Deltares/Wflow.jl/issues/688

The original idea was to use `Configuration.jl` to parse the TOML, but after discussion with @JoostBuitink and @vers-w we decided not to do this. Instead the implementation is custom, using `AbstractConfigSection` and a parse function `init_config_section`. 

An overview of the changes made in this PR:
- The package `PropertyDicts` is now used to access nested dictionaries as if they were nested structs (not sure if this is still actually required though)
- Conversion to the proper type of a config field is done in `init_config_section`, so the types do not have to be specified anymore in the code where the config field is used
- The package `EnumX` is used to make enumerations of config options, see the top of `config_structure.jl`. These enumerations are also used to validate these config options
- Some work was done to write the `Config` object back to `TOML` with the function `to_dict`, which is a requirement for https://github.com/Deltares/Wflow.jl/issues/701
- Basic printing functionality was added for the `Config` object mainly for developers, but this could be added to the log if desirable
- Some constructors were removed that are now automatically generated with the `@with_kw` macro
- Some function names were changed to reflect that they are constructors of a particular type
- Some function signatures were made more specific for clarity


Wflow/src/config.jl
```diff
@@ -0,0 +1,221 @@
...
+    # Routing method
+    land_routing::String = "kinematic-wave"
+    river_routing::String = "kinematic-wave"
+    # Kinematic wave routing
+    pit__flag::Bool = false
+    river_streamorder__min_count::Int = 6
+    land_streamorder__min_count::Int = 5
+    kinematic_wave__adaptive_time_step_flag::Bool = true
+    river_kinematic_wave__time_step::Union{Nothing, Float64} = nothing # Δt
+    land_kinematic_wave__time_step::Union{Nothing, Float64} = nothing # Δt
```
vers-w: This is also not correct in the [docs](https://deltares.github.io/Wflow.jl/dev/model_docs/model_settings.html#kinematic-wave-routing).
```suggestion
    river_kinematic_wave__time_step::Float64 = 900.0
    land_kinematic_wave__time_step::Float64 = 3600.0
```

Wflow/src/config.jl
```diff
@@ -0,0 +1,221 @@
...
+    soil_layer__thickness::Vector{Int} = [100, 300, 800]
+    saturated_hydraulic_conductivity_profile::String = "exponential"
+    # Routing method
+    land_routing::String = "kinematic-wave"
+    river_routing::String = "kinematic-wave"
+    # Kinematic wave routing
+    pit__flag::Bool = false
+    river_streamorder__min_count::Int = 6
+    land_streamorder__min_count::Int = 5
+    kinematic_wave__adaptive_time_step_flag::Bool = true
```
vers-w:
```suggestion
    kinematic_wave__adaptive_time_step_flag::Bool = false
```

Wflow/src/config.jl
```diff
@@ -0,0 +1,221 @@
...
+            new_dict[key] = nested_property_dict(value)
+        else
+            new_dict[key] = value
+        end
+    end
+
+    return PropertyDict(new_dict)
+end
+
+function Config(path::AbstractString; override...)
```
vers-w: Suggest to add a docstring to these `Config` functions (as in the original code). For example the `override...` argument is not very clear unless Configurations.jl is known.

Wflow/src/config.jl
```diff
@@ -0,0 +1,221 @@
...
+    dict = TOML.parsefile(path)
+    Config(dict; path, override...)
+end
+
+function Config(dict::AbstractDict; path::Union{Nothing, String} = nothing, override...)
+    @assert haskey(dict, "model") "The TOML is missing the mandatory [model] section."
+    @assert haskey(dict, "input") "The TOML is missing the mandatory [input] section."
```
vers-w: Probably good to also check `input.forcing` and `input.static` sections? These are required for the `sbm` and `sbm_gwf` model types, and I also think for the `sediment` model type (although it seems you could run the `sediment` model without `input.static` (when river transport is not included)), @hboisgon : not sure, is this the case?

Wflow/src/config.jl
```diff
@@ -0,0 +1,221 @@
...
+# Input configurations
+@option struct InputSection <: AbstractConfigSection
+    # Flow direction and modelling domains
+    path_forcing::String
+    path_static::String
+    basin__local_drain_direction::String
+    basin_pit_location__mask::Union{Nothing, String} = nothing
+    river_location__mask::String
+    reservoir_area__count::String = ""
+    reservoir_location__count::String = ""
```
vers-w: If `reservoir__flag` is `true` these are required, maybe worth adding?

Wflow/src/config.jl
```diff
@@ -0,0 +1,221 @@
...
+    subbasin_location__count::String
+    # Ouput locations: these are not directly part of the model
+    river_gauge__count::String = ""
+    river_gauge_grdc__count::String = ""
```
vers-w: This is user input (flexible), better to remove this.

Wflow/src/network.jl
```diff
@@ -291,11 +289,11 @@ function NetworkReservoir(dataset::NCDataset, config::Config, network::NetworkRi
     logging = false
     # allow reservoir only in river cells
     # note that these locations are only the reservoir outlet pixels
-    lens = lens_input(config, "reservoir_location__count"; optional = false)
+    lens = lens_input(:reservoir_location__count)
     locs = ncread(dataset, config, lens; sel = indices, type = Int, fill = 0, logging)
 
     # this holds the same ids as locs, but covers the entire reservoir
-    lens = lens_input(config, "reservoir_area__count"; optional = false)
+    lens = lens_input(:reservoir_area__count)
```
vers-w: Should `config` not be added as an argument, to check if these keys are included, or is this done somewhere else?

Wflow/src/routing/reservoir.jl
```diff
@@ -61,7 +61,8 @@ function ReservoirParameters(dataset::NCDataset, config::Config, network::Networ
     )
     outflowfunc = ncread(dataset, config, lens; sel = indices_outlet, type = Int, fill = 0)
 
-    lens = lens_input(config, "reservoir~lower_location__count")
+    # TODO: I don't understand what this is supposed to read
+    lens = (; name = "reservoir~lower_location__count", lens = nothing)
```
vers-w: This is reading the reservoir location ID of a lower (downstream) reservoir when reservoirs are linked. This is possible for reservoirs with a rating curve type (`ouflowfunc`) 2.

Wflow/src/routing/reservoir.jl
```diff
@@ -73,7 +74,7 @@ function ReservoirParameters(dataset::NCDataset, config::Config, network::Networ
     )
 
     n_reservoirs = length(area)
-    lens = lens_input(config, "reservoir_location__count"; optional = false)
+    lens = lens_input(:reservoir_location__count)
```
vers-w: Is `config` argument not required here to check if key is included, or is this handled somewhere else in the code?

Wflow/src/routing/surface_kinwave.jl
```diff
@@ -17,14 +17,15 @@ function init_kinematic_wave_timestepping(
...
+    adaptive = config.model.kinematic_wave__adaptive_time_step_flag
     @info "Kinematic wave approach is used for $domain flow, adaptive timestepping = $adaptive."
 
     if adaptive
         stable_timesteps = zeros(n)
         timestepping = TimeStepping(; stable_timesteps, adaptive)
     else
-        dt_fixed = get(config.model, "$(domain)_kinematic_wave__time_step", dt_fixed)
+        dt_config = getfield(config.model, Symbol("$(domain)_kinematic_wave__time_step"))
+        dt_fixed = isnothing(dt_config) ? dt_fixed : dt_config
```
vers-w: This can be set from the default value in `config`.

Wflow/test/bmi.jl
```diff
@@ -176,8 +176,7 @@ end
     model = BMI.initialize(Wflow.Model, tomlpath)
     @test Wflow.get_start_unix_time(model) == 9.466848e8
     satwaterdepth = mean(model.land.soil.variables.satwaterdepth)
-    model.config.model.cold_start__flag = false
-    Wflow.load_state(model)
+    model = BMI.initialize(Wflow.Model, tomlpath; model_cold_start__flag = false)
```
vers-w: Is this also possible with using `Wflow.load_state(model)`? This is for example important for OpenDA coupling (without initializing the `model` again).

vers-w (changes requested):
Nice work @SouthEndMusic ! 

Besides a couple of comments/questions, the code coverage has decreased and is failing, also at the project level, would be good to bump this up. I think it would be good if @JoostBuitink also checks the config settings in `config.jl`.

Wflow/src/config.jl
```diff
@@ -0,0 +1,221 @@
...
+@option struct WaterDemandSubSection <: AbstractConfigSection
+    domestic__flag::Bool = false
+    industry__flag::Bool = false
+    livestock__flag::Bool = false
+    paddy__flag::Bool = false
+    nonpaddy__flag::Bool = false
+end
+
+# Model configurations
+@option struct ModelSection <: AbstractConfigSection
```
vers-w: Maybe worth to have model sections per model type?

SouthEndMusic: That would be nice, but then we have to either:

- Also require this in the TOML, which is breaking
- Do some more processing of the TOML read before passing it to `Configurations.jl`, which sort of defeats the purpose of `Configurations.jl`

Wflow/src/config.jl
SouthEndMusic: You mean as a comment here? I don't think `Configurations.jl` supports validation logic like 'if this condition for this field, then that condition for that field'. I think it's best if such conditional validation is done in the code where the field is parsed.

Wflow/src/config.jl
SouthEndMusic: Good to know. This is never accessed in the Wflow core right? Currently `Configurations.jl` is very strict, in that if a TOML section is parsed with an `@option` like this, every non-hardcoded field yields an error. I made an issue: https://github.com/Roger-luo/Configurations.jl/issues/107. The package isn't actively maintained, the last release was 2 years ago. If there is no response to my issue I can take a stab at it myself.

Wflow/src/network.jl
SouthEndMusic: `haskey` does not work on the fields that are hardcoded in `config.jl`. If the field is (universally) mandatory but omitted in the TOML, this yields an error in `Configurations.jl`.

Wflow/src/config.jl
SouthEndMusic: I have to clarify some things here:
- The reason I explicitly check `model` and `input` here is that these are required for initialing the `HasSection` section (see also the inline docs I just added). 
- If `input.forcing` or `input.static` is missing this is caught by `Configurations.jl`.

Wflow/src/routing/reservoir.jl
SouthEndMusic: There's no test for this, maybe a follow-up issue is required?

Wflow/src/routing/reservoir.jl
SouthEndMusic: See the similar comment

Wflow/src/routing/surface_kinwave.jl
SouthEndMusic: OK, but notice that this changes the logic

SouthEndMusic: Hmm maybe not, I see now where the default values were in the code

Wflow/test/bmi.jl
SouthEndMusic: I'm not completely sure what you mean. I changed this test because (most of) the config is now immutable. That's generally fine because you can just re-initialize the config with overrides. If that's a big problem, then we can make some of the config fields mutable. I already had to do that for some of the `config.time` fields because those are changed in the `Clock` constructor (which I'm not a fan of, but I couldn't come up with a quick refactor to get rid of it).

SouthEndMusic (commented):
/

Wflow/src/config.jl
SouthEndMusic: I found out that there is functionality for this (I had to overload `Configurations.ignore_extra`), now extra fields are ignored for all sections.

SouthEndMusic: I can't actually remove these, these are needed. Does this mean that the whole input section is flexible, with some mandatory fields? I would argue in favor of keeping flexible parts in their own section, even though that might be breaking here

Wflow/src/config.jl
vers-w: OK, yeah, it seems indeed `Configurations.j` does not support this kind of validation. But it would be nice to add this validation to the code, if `reservoir__flag` is `true`, `reservoir_area__count` and `reservoir_location__count` should be set (not default empty string).

SouthEndMusic: OK, I'll set the default to `nothing` and then raise an error when the data is required.

Wflow/src/routing/reservoir.jl
vers-w: There is indeed no integrated/model test for this, there is only a unit test for linked reservoirs. This optional key is read from the TOML file:

```toml
[input]
"reservoir~lower_location__count" = "linked_reservoirs"
```
The current implementation (this PR) now sets this parameter always to its default value, that is not correct.

Wflow/src/config.jl
vers-w: > Good to know. This is never accessed in the Wflow core right? Currently `Configurations.jl` is very strict, in that if a TOML section is parsed with an `@option` like this, every non-hardcoded field yields an error. I made an issue: [Roger-luo/Configurations.jl#107](https://github.com/Roger-luo/Configurations.jl/issues/107). The package isn't actively maintained, the last release was 2 years ago. If there is no response to my issue I can take a stab at it myself.

Access in Wflow core is done here: [io.jl L650](https://github.com/Deltares/Wflow.jl/blob/toml_parse/Wflow/src/io.jl#L650), is that what you mean? These fields are not hard-coded. Hm, if the package is not actively maintained, this seems a bit risky to me. I would say if we could implement something similar (not depending on `Configurations.jl`) without too much extra work, that would be the preferred option I think. Also good to know your thoughts on this @JoostBuitink (as you did try out this package for wflow a while ago)?

Wflow/src/config.jl
SouthEndMusic: @vers-w can you point to the best place in the code to validate this? Probably the earliest relevant point in the initialization

Wflow/src/config.jl
vers-w: > I can't actually remove these, these are needed. Does this mean that the whole input section is flexible, with some mandatory fields? I would argue in favor of keeping flexible parts in their own section, even though that might be breaking here

Yes, good point. It will be breaking, but that's OK as we are working towards v1.0. If we change this, then hydrom_wflow also needs to be updated.

SouthEndMusic: Here's what I can do for the flexible part of the `[input]` section in a non-breaking way: I load the TOML dict, loop over the (key, value) pairs of the input section and put the fields that are not fields in `InputSection` in a dict within `InputSection` from where it can be accessed.

Wflow/src/routing/reservoir.jl
SouthEndMusic: I think I'll handle this as suggested [here](https://github.com/Deltares/Wflow.jl/pull/689#discussion_r2336608541), because tildes are not allowed in field names

Wflow/test/bmi.jl
vers-w: Ah, OK, yes, for this to work (setting states in the model without re-initializing) the `model.cold_start__flag` needs to be mutable.

Wflow/src/config.jl
JoostBuitink: I think it would be best to validate this already when reading the config. Given that Configurations.jl does not support this, and appears to be not too actively maintained, I wonder if it is worth writing our own logic around this, or whether it is better to define our own config object. 

It would also be nice to check if certain layers required specifically to the model type/settings (for example layers like soil moisture, glacier locations, reservoir parameters, floodplain volumes) are present. It is a bit tricky that those can be defined both in the `[input.static]` or `[input.cyclic]` settings.

Also, some fields like `river_routing` only support two options ("kinematic-wave" or "local-inertial"). If a user tries something else (or a typo), it would be best to throw an error when creating the config object (rather than within the computation code/initialization like it is now). I once tried to get this sorted using Configurations, but failed to do so.

vers-w: Yes, I would say as part of the `function Config`?

Wflow/src/config.jl
JoostBuitink: I think that would be nice, as the input section is indeed extremely flexible, where users can define a variable with any name and map it to a layer in the netcdf (staticmaps.nc). But indeed, there are a couple of fields mandatory to be defined. 

Your (@SouthEndMusic) sounds good to me! I just have the slight concern that we are writing quite some logic around a a package that is not actively maintained, and wonder if creating our own code around this is actually easier..

SouthEndMusic: @JoostBuitink I see what you mean. For Ribasim we've been using `Configurations.jl` for a while, so we're used to working with its constraints. The bottom line is that it's hard to fit the very flexible Wflow TOML into a schema. Indeed, if we need more and more workarounds we should make something custom, but I think for what I have now it's still worth using.

vers-w: @SouthEndMusic and @JoostBuitink : suggest to discuss this tomorrow further; what is the added value of using `Configurations.jl` while it is not actively maintained and considering the constraints? How much extra work is needed for a custom implementation?

Wflow/src/config.jl
SouthEndMusic: This hints at a difference we have in 'config philosophy'. In Ribasim the `config` object is a static representation of the TOML which is only used in initialization (where any validation is done) and post-processing. I'm fine with doing this differently in Wflow, as long as the approach is consistent. I'll look into moving as much of the config validation to the `Config` constructor as possible.

vers-w: Not sure if I understand your comment completely, but generating the `config` object is the first step before initializing a model from that `config`. So, I think it makes sense to do the validation here, as @JoostBuitink also indicated.

Wflow/src/io.jl
```diff
@@ -1699,15 +1570,11 @@ end
...
-        ROUTING_OPTIONS,
-        "kinematic-wave",
-    )::String
+    land = config.model.land_routing
+    @assert land ∈ ROUTING_OPTIONS  # Already validated in `validate_config`
+
+    river = config.model.river_routing
+    @assert river ∈ ROUTING_OPTIONS  # Already validated in `validate_config`
 
     subsurface = config.model.type == "sbm" ? "kinematic-wave" : "groundwaterflow"
```
SouthEndMusic: The nothing here used to be `"groundwaterflow"` but that is referenced nowhere else

Wflow/src/config.jl
```diff
@@ -0,0 +1,314 @@
...
+        "366_day",
+        "360_day",
+    ),
+)
+get_option_dict(::TimeSection) = time_section_options
+
+# Logging related configurations
+@option struct LoggingSection <: AbstractConfigSection
+    silent::Bool = false
+    loglevel::String = "info"
```
SouthEndMusic: Is the loglevel input allowed to be an integer?

Wflow/src/io.jl
```diff
@@ -773,16 +663,13 @@ function nc_variables_dims(nc_variables, dataset, config)
...
             push!(
                 ncvars_dims,
                 (
-                    par = par,
-                    var = var,
+                    par,
+                    var,
                     location_dim = nc_var["location"],
                     locations = [nc_var["location"]],
                 ),
```
SouthEndMusic: I feel like there is some validation missing here, or is that done somewhere else? That could already be done in `config.jl`

Wflow/src/io.jl
```diff
@@ -1644,31 +1527,20 @@ end
 "Get `index` for dimension name `layer` based on `config` (TOML file)"
 function get_index_dimension(var, config::Config, dim_value)::Int
     if haskey(var, "layer")
-        v = get(config.model, "soil_layer__thickness", [100, 300, 800])::Vector{Int64}
-        inds = collect(1:(length(v) + 1))
+        inds = collect(1:(length(config.model.soil_layer__thickness) + 1))
```
SouthEndMusic: I don't understand this code. What is the difference between `dim_value` and `index` (apart from the potential out of bounds error)?

Wflow/src/io.jl
vers-w: What `nothing` do you mean here?

Wflow/src/config.jl
vers-w: From checking the code in logging.jl (and the docs) this is allowed.

Wflow/src/io.jl
vers-w: This is not vaildated somewhere else. What kind of validation were you thinking of?

Wflow/src/io.jl
vers-w: Yeah, I agree, this function is not adding much. Probably a remainder of refactoring (another model concept did support also an `extra_dim`).

SouthEndMusic:
After discussing with @evetion I'm motivated to keep `Configurations.jl` as a dependency, because:

- The only part of the configuration that has both standard and flexible fields is the input section. This is handled in a simple wrapper of the default `Configurations.from_dict` for this section.
- Other parts fit nicely in a schema as defined by `Configurations.jl`, including the very flexible output section. This means that a lot of parsing of the entries here can be done (in a simpler way) in `config.jl` than in `io.jl`.
- In general, overloading of `Configurations.from_dict` avoids all needs to work around `Configurations.jl` as I did before. This is within the documented usage of `Configurations.jl`, specifically [this part](https://configurations.rogerluo.dev/stable/convert/).
- A particularly nice conversion that we can let `Configurations.jl` do this way is from string options to enumerator (`EnumX.Enum`) instances. This means that we have a centralized way to define which options a particular part of the configuration has and to throw an informative error if an invalid option was given in the TOML.

My apologies for making the PR so big. I do believe that (most of) this is a good step in making the code easier to understand as a newcomer

docs/model_docs/model_settings.qmd
```diff
@@ -58,7 +58,6 @@ The Tables in this subsection show the available models settings for the model t
 | `land_surface_water_flow_threshold__depth` | Water depth threshold for calculating overland flow between cells | m | 0.001 |
 | `river_water_flow__froude_limit_flag` | Limit river flow to subcritical-critical according to Froude number | - | `true` |
 | `land_surface_water_flow__froude_limit_flag` | Limit overland flow to subcritical-critical according to Froude number | - | `true` |
-| `river_water_flow__froude_limit_flag` | Limit river flow to subcritical-critical according to Froude number | - | `true` |
```
evetion, line 61: I assume it was discussed to drop this parameter? Would be good to mention such changes in the PR description (just like renaming parameters).

Wflow/test/sbm_config.toml
evetion, line 1: Many of these changes are due to formatting. Might be good to introduce a pre-commit that handles this? Or vscode config, so this is the last time this is needed.

Wflow/src/config.jl
```diff
@@ -0,0 +1,556 @@
...
+function Base.show(io::IO, c::AbstractConfigSection)
+    first = true
+    for field in fieldnames(typeof(c))
+        f = getfield(c, field)
+        if f !== nothing
+            first && (first = false; println(io))
+            println(io, "\t\t$field\t= $f")
+        end
+    end
+end
```
evetion: Consider moving this into config_utils.jl or similar, so there's a single file that mirrors the toml file?

Wflow/src/config.jl
```diff
@@ -0,0 +1,556 @@
...
+
+# config value converter to write the Config instance to a TOML file
+# usage: Configurations.to_toml(custom_convert, path, config)
+custom_convert(x::Any) = x
+custom_convert(x::EnumX.Enum) = remove_leading_underscore(String(Symbol(x)))
+custom_convert(x::RoutingType.T) =
+    (x == RoutingType.kinematic_wave) ? "kinematic-wave" : "local-inertial"
+custom_convert(x::LogLevel) = lowercase(string(x))
+custom_convert(var::InputEntry) = to_dict(var)
+custom_convert(::Nothing) = ""
```
evetion: EOF

Wflow/src/domain.jl
```diff
@@ -378,16 +375,10 @@ function reservoir_mask(
...
-    domain::Domain;
-    mask_value = 0,
-)
-    do_reservoirs = get(config.model, "reservoir__flag", false)::Bool
-    if do_reservoirs
+function mask_reservoir_coverage!(mask::AbstractMatrix, config::Config, domain::Domain)
+    if config.model.reservoir__flag
         inds_cov = collect(Iterators.flatten(domain.reservoir.network.indices_coverage))
-        mask[inds_cov] .= mask_value
+        mask[inds_cov] .= 0
```
evetion, line 381: So mask_value was never used in the codebase, and couldn't be set from the config?

Wflow/src/groundwater/aquifer.jl
```diff
@@ -356,16 +355,14 @@ function conductance(
                 (aquifer.parameters.top[j] - aquifer.parameters.bottom[j])
         end
         return saturation * aquifer.variables.conductance[nzi]
-    else
-        error(
-            """An unknown "conductivity_profile" is specified in the TOML file ($conductivity_profile).
-            This should be "uniform" or "exponential".
-            """,
-        )
```
evetion, line 376: Nice to see this logic being moved to the config itself!

docs/model_docs/model_settings.qmd
SouthEndMusic: I removed this because it is an exact duplicate of line 59

vers-w:
> After discussing with @evetion I'm motivated to keep `Configurations.jl` as a dependency, because:

I am a bit puzzled, my impression is that, as we also did discuss last week, that there are not that many structs that we could initialize in a customized way without using `Configurations.jl` as a dependency. With the `@with_kw` macro, default values are possible, and I think initializing these structs from the parsed TOML file (`TOML.parsefile`) can be done quite easily. This would also give us more flexibility, for example related to https://github.com/Deltares/Wflow.jl/pull/689#discussion_r2336058621, and for any future changes we do not have to deal with some of the `Configurations.jl` restrictions. Finally, it is probably also easier to understand and to expand on by a newcomer.

Wflow/src/Wflow.jl
```diff
@@ -314,8 +318,8 @@ function run!(model::Model; close_files = true)
     end
 
     # copy TOML to dir_output, to archive what settings were used
-    if haskey(config, "dir_output")
-        src = normpath(pathof(config))
+    if !isnothing(config.dir_input)
```
JoostBuitink: I think this should still be checking for a dir_output (e.g. copying the toml to that folder only makes sense is all output is written to a different folder)

Wflow/src/bmi.jl
```diff
@@ -369,7 +364,7 @@ function save_state(model::Model)
 end
 
 function get_start_unix_time(model::Model)
-    return datetime2unix(DateTime(model.config.time.starttime))
+    return datetime2unix(model.config.time.starttime)
```
JoostBuitink, line 367: Related to this change: my current toml was not able to be read with the PR (`MethodError: Cannot `convert` an object of type String to an object of type Dates.DateTime`). This is also related to hydromt_wflow, as I believe the dates are put in the toml as a strings. Not sure if that is related to a limitation there, of an actual decision. It would be nice if we can still support both (so strings and DateTime objects). If that is not possible, hydromt_wflow needs to be fixed, together with the model conversion code (v0.x to v1.0).

Example of my current toml (removing the `"` fixes it btw):
```toml
[time]
starttime = "2010-01-01T00:00:00"
endtime = "2010-06-01T00:00:00"
```

Wflow/src/config.jl
```diff
@@ -0,0 +1,581 @@
...
+Output data can be written to netCDF or CSV files.
+For configuration files we use TOML.
+=#
+
+# Option enumerators
+@enumx RoutingType kinematic_wave local_inertial
+@enumx ModelType sbm sbm_gwf sediment
+@enumx CalendarType standard gregorian proleptic_gregorian julian noleap _365_day all_leap _366_day _360_day
+@enumx ConductivityProfileType uniform exponential
+@enumx SHCPType exponential exponential_constant layered layered_exponential
```
JoostBuitink: These names (especially the `SHCPType`) are a bit confusing. Perhaps something like `GwfConductivityProfile` and `VerticalConductivityProfile`?

Wflow/src/config.jl
```diff
@@ -0,0 +1,581 @@
...
+###
+### Input section
+###
+
+@option struct InputEntry <: AbstractConfigSection
+    # Option 1
+    netcdf_variable_name::Union{Nothing, String} = nothing # Comes from dict["netcdf"]["variable"]["name"]
+    scale::Vector{Float64} = [1.0]
+    offset::Vector{Float64} = [0.0]
+    layer::Union{Nothing, Vector{Int}} = nothing
```
JoostBuitink: In 99% of the usecases, people use a float for the scale and offset (so not a vector). I think this vector requirement is only needed for layers which have the layer dimension (such as the `c` parameter, see this [example](https://deltares.github.io/Wflow.jl/dev/user_guide/toml_file.html#modify-parameters) as well). It would be nice if a direct float (like the first couple of examples) is possible as well. If this is not possible (not ideal in my opinion), but then hydromt_wflow should take care of this as well in the toml conversion code.

Wflow/src/config.jl
```diff
@@ -0,0 +1,581 @@
...
+    # Option 2
+    value::Any = nothing
+    # Option 3
+    standard_name::Union{Nothing, String} = nothing
```
JoostBuitink: How is this different from the "netcdf_variable_name" from above? I assume this is used in the case no scale/offset are specified right? In both cases, I think the "netcdf_variable_name" is a better description. The "standard_name" is actually referring to the name before the `=` sign in the toml, and the string describes the name of the layer in the netcdf.

Wflow/src/config.jl
```diff
@@ -0,0 +1,581 @@
...
+    variables::PropertyDictType = PropertyDict(Dict{String, Any}())
+end
+
+###
+### Input section
+###
+
+@option struct InputEntry <: AbstractConfigSection
+    # Option 1
+    netcdf_variable_name::Union{Nothing, String} = nothing # Comes from dict["netcdf"]["variable"]["name"]
```
JoostBuitink: Related to my comments from below (and perhaps for a different PR), but do we actually have a reason to have this nested dict (`dict["netcdf"]["variable"]["name"]`)? Wouldn't it be better to have the toml like below? Also tagging @vers-w to be sure!

Current
```toml
[input.forcing.atmosphere_water__precipitation_volume_flux]
netcdf.variable.name = "P"
scale = 1.5
offset = 0.5

# or when not using scale/offset
[input.forcing]
atmosphere_water__precipitation_volume_flux = "P"
```

New?
```toml
[input.forcing.atmosphere_water__precipitation_volume_flux]
netcdf_variable_name = "P"
scale = 1.5
offset = 0.5

# or when not using scale/offset
[input.forcing]
atmosphere_water__precipitation_volume_flux = "P"
```

Wflow/src/config.jl
```diff
@@ -0,0 +1,581 @@
...
+    basin_pit_location__mask::Union{Nothing, String} = nothing
+    river_location__mask::String
+    reservoir_area__count::Union{Nothing, String} = nothing
+    reservoir_location__count::Union{Nothing, String} = nothing
+    subbasin_location__count::String
+    # Variable name mappings
+    forcing::InputEntries
+    static::InputEntries
+    cyclic::InputEntries = InputEntries()
+    flexible::PropertyDictType
```
JoostBuitink: Maybe rename `flexible` to something like `location_maps` or something?

Wflow/src/config.jl
```diff
@@ -0,0 +1,581 @@
...
+    return invoke(from_dict, Tuple{Type, AbstractDict{String}}, InputSection, dict)
+end
+
+###
+### Output section
+###
+
+@option struct CoordinateSection <: AbstractConfigSection
+    x::Float64
+    y::Float64
```
JoostBuitink: Does this mean that only floating point values are accepted? If users use for example and UTM coordinate system, integers could also be an option (although I guess even then it might be a float to be precise). Just wanted to check!

Wflow/src/config.jl
```diff
@@ -0,0 +1,581 @@
...
+
+@option struct NetCDFScalarVariable <: AbstractConfigSection
+    name::String
+    parameter::String
+    reducer::Union{Nothing, ReducerType.T} = nothing
+    layer::Union{Nothing, Int} = nothing
+    index::Union{Nothing, Int, IndexSection} = nothing
+    map::Union{Nothing, String} = nothing
+    coordinate::Union{Nothing, CoordinateSection} = nothing
+    location::Union{Nothing, String} = nothing
```
JoostBuitink: I just only now learn that we have a `location` argument fot NetCDF scalar output. Actually not sure what it does, I guess it is used to define the name of the location dimension for the scalar series that use either an index or coordinate?

Wflow/src/config.jl
```diff
@@ -0,0 +1,581 @@
...
+"Construct a path relative to both the TOML directory and the optional `dir_output`"
+output_path(config::Config, path::Union{Nothing, AbstractString}) =
+    normpath(dirname(config), config.dir_output, path)
+
+function validate_config(config::Config)::Nothing
+    valid = true
+
+    # check if there is overlap in forcing and cyclic parameters
+    if do_cyclic(config)
+        overlap = intersect(keys(config.input.forcing), keys(config.input.cyclic))
```
JoostBuitink: This can be extended to also cover keys in the config.input.static part

Wflow/src/io.jl
```diff
@@ -184,10 +26,11 @@ function get_at(
...
 
-function get_at(ds::CFDataset, var::NamedTuple, i)
-    data = read_standardized(ds, var.name, (x = :, y = :, time = i))
-    if var.scale != 1.0 || var.offset != 0.0
-        data .= data .* var.scale .+ var.offset
+function get_at(ds::CFDataset, var::InputEntry, i)
+    (; scale, offset) = var
+    data = read_standardized(ds, variable_name(var), (x = :, y = :, time = i))
+    if scale != 1.0 || offset != 0.0
+        data .= data .* scale .+ offset
```
JoostBuitink, line 33: Isn't this line now always triggered given that scale and offsets are defined as vectors?

Wflow/src/sbm_gwf_model.jl
```diff
@@ -13,37 +13,27 @@ function Model(config::Config, type::SbmGwfModel)
...
-        constanthead = get(config.model, "constanthead__flag", false)::Bool,
-        water_demand = haskey(config.model, "water_demand"),
-        pits = false,
-        min_streamorder_river = get(config.model, "river_streamorder__min_count", 6),
-        min_streamorder_land = get(config.model, "land_streamorder__min_count", 5),
-    )
-
-    @info "General model settings" modelsettings[keys(modelsettings)[1:7]]...
+    @info "General model settings" (;
+        model = config.model.snow__flag,
```
JoostBuitink:
```suggestion
        snow = config.model.snow__flag,
```

Wflow/test/io.jl
```diff
@@ -283,19 +273,28 @@ end
...
-    "scale" => 3.0,
-    "offset" => 100.0,
-    "netcdf" => Dict("variable" => Dict("name" => "SoilThickness")),
+config.input.static.dict["snowpack__degree-day_coefficient"] =
+    Wflow.InputEntry(; value = 2.0)
+config.input.static.dict["soil__thickness"] = from_dict(
+    Wflow.InputEntry,
+    Dict(
+        "scale" => 3.0,
+        "offset" => 100.0,
```
JoostBuitink: Ah perhaps my earlier comment(s) on this can be ignored, it looks like single values are still valid, right?

Wflow/test/run.jl
```diff
@@ -5,58 +5,72 @@ using NCDatasets
...
+    model_cold_start__flag = false, # warm start
+    fews_run__flag = true,
+    state_path_input = joinpath(
+        dirname(tomlpath),
+        "data/state-test/outstates-moselle-january-1of2.nc",
+    ),
+    state_path_output = joinpath(
+        dirname(tomlpath),
+        "data/state-test/outstates-moselle-january-2of2-fews_run.nc",
+    ),
```
JoostBuitink: Still a bit in doubt whether I like this new implementation over the old one. To me, it feels a bit more natural to read the toml in one step, and then allow for modifications afterwards. Now these two have be done in the same "step". 

Not a massive change, or maybe just something that I need to get used to. Although changing specific fields (e.g. `time_starttime`) is maybe less intuitive than `config.time.starttime = ...` )

Wflow/test/run_sbm.jl
```diff
@@ -155,25 +155,22 @@ Wflow.close_files(model; delete_output = false)
...
-config = Wflow.Config(tomlpath)
-config.model.snow_gravitional_transport__flag = false
 
 model = Wflow.Model(config)
 Wflow.run_timestep!(model)
 
 @testset "lateral snow transport off" begin
     snow = model.land.snow
-    @test snow.variables.snow_storage[5] ≈ 3.7686103651001375
+    @test snow.variables.snow_storage[5] ≈ 3.768513390588815
```
JoostBuitink: should the toml be read again and the flag explicitly set to false? Because I think that is causing the snow_storage to be different (`snow_gravitional_transport__flag` is set to true for this test I think).

Wflow/test/run_sbm_gwf.jl
```diff
@@ -110,15 +111,20 @@ Wflow.close_files(model; delete_output = false)
...
+    model_river_routing = "local-inertial",
+    model_land_routing = "local-inertial",
+)
+
+config.input.static.dict["river_bank_water__elevation"] =
+    Wflow.InputEntry(; standard_name = "bankfull_elevation")
+config.input.static.dict["river_bank_water__depth"] =
+    Wflow.InputEntry(; standard_name = "bankfull_depth")
+config.input.static.dict["land_surface_water_flow__ground_elevation"] =
+    Wflow.InputEntry(; standard_name = "wflow_dem")
```
JoostBuitink: Related to my earlier comment, here I am also doubting whether this new implementation is actually easier. To me, the previous option is much easier and more straightforward. Similar as before, only expert users will touch this interface, but still a remark I wanted to make

Wflow/test/sbm_gwf_piave_demand_config.toml
```diff
@@ -117,13 +117,13 @@ vegetation__leaf-area_index = "LAI"
 [model]
 type = "sbm_gwf"
 constanthead__flag = true
-gravitational_snow_transport = true
+snow_gravitational_transport__flag = true
```
JoostBuitink, line 120: Good that you spotted this!

docs/model_docs/model_settings.qmd
JoostBuitink: It is actually listed twice in this table ;)

JoostBuitink (commented):
Nice work, and massive PR indeed ;) Thanks a lot for all the other code-improvements/small refactors!

I have added a couple of remarks/questions/suggestions, I hope they are all clear. Overall - but although it might be a bit late given that all this work is already there - I have the same doubts as Willem on whether the (sort of) "rigidness" of Configurations.jl and the "flexibility" of the Wflow.jl toml are a good match. Let's continue the discussion tomorrow!

Wflow/src/Wflow.jl
SouthEndMusic: Good catch!

Wflow/src/bmi.jl
SouthEndMusic: I did a quick check and this works now with the new implementation without `Configurations.jl`. This would have been fixable with `Configurations.jl` though

Wflow/src/config.jl
SouthEndMusic: Fair point, these are quite out of context here. Although, the code base is full of abbreviations that you are used to but (non-hydrological) newcomers are not 😉

Wflow/src/config.jl
SouthEndMusic: A scalar entry here is still possible, this is converted to a vector to have a consistent type. It's worth checking whether this is parsed properly lateron.

Wflow/src/config.jl
SouthEndMusic: I came to this implementation from looking at the different ways in which this part of the config was parsed before. If different field names here are more clear, this can be adjusted. It must still be clear in what way this part of the config was initialized, and that might be lost when `netcdf_variable_name` and `standard_name` are put in the same field. Let's discuss this together.

Wflow/src/config.jl
SouthEndMusic: Yes, I wondered why this was nested so deeply but I figured there would be a reason for it. I agree that the new way is cleaner, but it is breaking.

Wflow/src/config.jl
SouthEndMusic: I don't think I can answer this question directly, let's discuss.

Wflow/src/config.jl
SouthEndMusic: An interesting consequence of making the config structure explicit 😄

Wflow/src/io.jl
SouthEndMusic: That is a good point, I also wondered this. Is the computation expensive enough to warrant this condition? Let's discuss.

Wflow/test/run.jl
SouthEndMusic: The way this works is now a bit different because I don't use `Configurations.jl` anymore: the override of the values in the TOML is passed as a dict. My biased opinion is that config modification should be discouraged because it makes the code harder to understand. Re-initializing the config instead if modifying it is one of the steps to make the tests more self contained which is preferred by the Julia VSCode plugin. Although I don't know whether in-place config modification is needed outside the context of the tests (I know it is done in a few places in the core code, I left the relevant fields mutable).

Wflow/test/run_sbm.jl
SouthEndMusic: I have a few failing tests still with the latest implementation, let's discuss together.

SouthEndMusic:
@JoostBuitink @vers-w I have addressed your comments. Most of the old interface for mutating the config is now intact, type conversions are handled automatically. One thing that I think is worth reviewing is what the info messages now look like. The order has changed, and some things are also formatted differently.

vers-w:
Just did a final check of the logging messages:

- Not sure if it is required to log when a parameters is modified with scale 1.0 and offset 0.0:
  `Info: NetCDF parameter wflow_ldd is modified with scale 1.0 and offset 0.0.`
   I think it is better to remove these, @JoostBuitink ?
- It would be nice to use single quotes for the parameter names and values (this is clearer in the console) and already used. 
- Something goes wrong with formatting this external parameter name?:
  `[ Info: NetCDF parameter wflowgaugesgrdc is modified with scale 1.0 and offset 0.0.`

SouthEndMusic:
@vers-w I resolved all the merge conflicts but there is still one test failing that I don't know how to fix. Can you have a look?

vers-w:
> @vers-w I resolved all the merge conflicts but there is still one test failing that I don't know how to fix. Can you have a look?

@SouthEndMusic : I have resolved the tests and changed the log message for modifying NetCDF parameters (https://github.com/Deltares/Wflow.jl/pull/689/commits/5bcfa41bf1cfd4d2e8617b5041990bbe7a30783e). I think it is fine to merge, @JoostBuitink: maybe you want to do a final check?

PR #704, opened 2025-09-22
Update standard names

SouthEndMusic:
Fixes https://github.com/Deltares/Wflow.jl/issues/697.

Note that this was done by Claude, so check carefully that the task was completed as intended. The name transformation mapping can be found in this file. Note that not all of them are standard names, for instance the routing types.

[standard_name_transformations_final.csv](https://github.com/user-attachments/files/22511815/standard_name_transformations_final.csv)

PR #707, opened 2025-09-25
Fix standard names

vers-w:
A couple of standard names were not updated in https://github.com/Deltares/Wflow.jl/pull/704.


SouthEndMusic (approved):
LGTM

PR #711, opened 2025-09-30
Update manifest files

vers-w:
Build fails on TC.


JoostBuitink (approved):
LGTM! Indeed good note to change version numbers to rc2 if we move the tag (which is preferred I think).  I was a bit surprised that Aqua didn't pick this up, but it looks like it only checks the Project.toml and not the Manifest.

PR #719, opened 2025-10-14
Fix non-optional parameters and allow static input `external_inflow` and `outflow_obs`

### LINKED ISSUE #718: Fix non-optional input parameters
Some non-optional input parameters are missing `optional = false` as argument for the `lens_input_parameter` function (although in this case wflow still fails (correctly) in the `ncread` function). Needs fixing for consistency and cleaner error is thrown from `lens_input_parameter`.

vers-w:
## Issue addressed
Fixes #718


JoostBuitink (approved):
LGTM, I agree that this implementation is much cleaner as it keeps all relevant settings together! Thanks for improving this!

PR #713, opened 2025-10-01
Support water balance computations

vers-w:
## Issue addressed
Fixes #90

- [x] Update changelog, not sure if this should be part of rc2 or a new release candidate.

PR #726, opened 2025-10-27
Fix mass balance subsurface flow at sub-daily model time step

### LINKED ISSUE #724: Mass balance at sub-daily time step for subsurface flow not corrrect
This is the case for `LateralSSF` and `UnconfinedAquifer` subsurface flow concepts. At a sub-daily model time step the units of different variables to compute the water mass balance do not match.

vers-w:
## Issue addressed
Fixes #724 

Also addressed some review comments from https://github.com/Deltares/Wflow.jl/pull/713.

### DIFF
--- a/Wflow/Project.toml
+++ b/Wflow/Project.toml
@@ -1,7 +1,7 @@
 name = "Wflow"
 uuid = "d48b7d99-76e7-47ae-b1d5-ff0c1cf9a818"
 authors = ["Deltares and contributors"]
-version = "1.0.0-rc2"
+version = "1.0.0-dev"
 
 [deps]
 Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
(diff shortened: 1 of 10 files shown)


Wflow/src/mass_balance.jl
```diff
@@ -56,7 +56,8 @@ function HydrologicalMassBalance(domain::Domain, config::Config)
         else
             river_water_balance = MassBalance(; n = n_river)
         end
-        routing = FlowRoutingMassBalance(;
+        args = (river_water_balance, reservoir_water_balance)
+        routing = FlowRoutingMassBalance{typeof.(args)...}(;
```
SouthEndMusic: Why do you need this? I'm pretty sure the default constructor can figure out the type parameters automatically

vers-w: Yeah, I agree, the default constructor can figure this out, removed `args`.

Issue #696, opened 2025-09-18
Add a pre-commit hook


SouthEndMusic:
As suggested [here](https://github.com/Deltares/Wflow.jl/pull/689#discussion_r2352512170). This will automatically run formatting and do certain code quality checks when committing.

PR #734, opened 2025-11-03
Add pre commit

SouthEndMusic:
Fixes https://github.com/Deltares/Wflow.jl/issues/696. For the configuration see https://github.com/Deltares/Wflow.jl/blob/add_pre_commit/.pre-commit-config.yaml.


evetion (approved):
I would also add a Gitub action like https://github.com/Deltares/Ribasim/blob/main/.github/workflows/pre-commit_check.yml, to make sure pre-commit is always run before merging.

Issue #617, opened 2025-05-21
Consider using internal timestep for the vertical part


JoostBuitink:
### Feature type

Adding new functionality

### Improvement Description

Have an internal timestep for the vertical processes

SouthEndMusic:
Refinement conclusions:

- Start working on this after https://github.com/Deltares/Wflow.jl/pull/735 is merged
- Do sub-timestep from `update_land_hydrology_model!` up to and including `update_soil_water_storage!`
- Sub timestep is minimum of sub timestep from TOML and from CFL condition from subsurface routing
- For output it is required to bookkeep cumulative fluxes
- Do the same for `sbm` and `sbm_gwf` model types

PR #735, opened 2025-11-03
Use SI units for internal computations

### LINKED ISSUE #326: Review units of variables
### Feature type

### Improvement Description

Currently, there is a mix between units in different model components, e.g. mm/dt for the vertical processes, and m3/day for the lateral processes. While this isn't a problem directly, it should be good to review all units as an activity before v1.0. Additionally some internal variables have units that are not really user friendly (such as `recharge` of the LateralSSF, which is now in `m2 Δt-1`).

_No response_

### Additional Context

_No response_

### LINKED ISSUE #769: Annotate the units of the sediment computations
While working on https://github.com/Deltares/Wflow.jl/issues/326 I found out that figuring out the units of the sediment computations takes some digging in the literature and some help from @hboisgon, so we decided to make it a follow-up issue.

### LINKED ISSUE #810: Include parameter unit in NetCDF export
### Feature type

Adding new functionality

### Improvement Description

Currently the units of parameters are not written in the NetCDF export file. Including the unit would make the export better meet standards and conventions. For example: FEWS reads the units as well which allows for automatic unit conversions (if needed).

_No response_

### Additional Context

_No response_

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/326
Fixes https://github.com/Deltares/Wflow.jl/issues/769
Fixes https://github.com/Deltares/Wflow.jl/issues/810

## Checklist

## Comments

Please when reviewing the code keep in the back of your mind the question whether everything would still work with a non-default timestep.

Changes in the order in which I think the review makes the most sense:

- A few small sediment documentation fixes
- Have a good look at the unit implementation in `units.jl`
- Have a good look at `io.jl` and the `ncread` function where the unit conversions for IO are done
- For most numerical computations I added the units as I understood them. Please check whether they make sense.

Please see [this](https://github.com/Deltares/Wflow.jl/pull/735#issuecomment-3943665220) comment for some the causes I identified for changes in outcome (other than the change in units)

Also note that for some outputs which are denoted in the TOML by the path in the model object the unit is not available, which will result in warnings like this:

```
┌ Warning: No unit was found for routing.overland_flow.sediment_flux.variables.clay, so the output will be expressed in standard SI units (K, s, m, kg, J).
└ @ Wflow D:\repositories\Deltares\Wflow.jl\Wflow\src\io.jl:703
┌ Warning: No unit was found for routing.overland_flow.transport_capacity.variables.clay, so the output will be expressed in standard SI units (K, s, m, kg, J).
└ @ Wflow D:\repositories\Deltares\Wflow.jl\Wflow\src\io.jl:703
┌ Warning: No unit was found for routing.river_flow.sediment_flux.boundary_conditions.transport_capacity, so the output will be expressed in standard SI units (K, s, m, kg, J).
└ @ Wflow D:\repositories\Deltares\Wflow.jl\Wflow\src\io.jl:703
┌ Warning: No unit was found for routing.river_flow.sediment_flux.boundary_conditions.erosion_land_clay, so the output will be expressed in standard SI units (K, s, m, kg, J).
```


Wflow/src/routing/surface/surface_kinwave.jl
```diff
@@ -1,37 +1,36 @@
...
-    storage::Vector{Float64} = zeros(n)      # Kinematic wave storage [m³] (based on water depth h)
-    h::Vector{Float64} = zeros(n)            # Water depth [m]
+    # Discharge [m³ s⁻¹]
+    q::Vector{Float64} = zeros(n)
+    # Lateral inflow per unit length [m² s⁻¹]
+    qlat::Vector{Float64} = zeros(n)
+    # Inflow from upstream cells [m³ s⁻¹]
+    qin::Vector{Float64} = zeros(n)
+    # Average inflow from upstream cells  [m³ s⁻¹] for model time step dt
+    qin_av::AverageVector = AverageVector(; n)
```
vers-w: Suggest to use here `Vectors` for average and cumulative variables, and not a introduce an new type `AverageVector` for this. I think that is cleaner and also easier to understand, so for example:

```
qin_average::Vector{Float64} = zeros(n)
qin_cumulative:::Vector{Float64} = zeros(n)
```
etc.

SouthEndMusic: I discussed this a while back with @verheem who made a similar change in Ribasim. It's not great that the unit of a vector depends on where the code is in its execution (i.e. before or after averaging), which is why the `AverageVector` object has separate `cumulative_material` and `average` fields with static units. I also added a safeguard in `get_average` to make sure no new material was added to `cumulative_material` when accessing the average.

vers-w: Yes, I agree, it's not ideal that the unit of a vector depends on where the code is in its execution. My recommendation is to add a `cumulative` (e.g. `qin_cumulative` ) field directly to the struct (and not as part of the new introduced object `AverageVector`), to resolve this. I think it is cleaner and easier to understand, and the `cumulative` variable is also a useful addition on it's own (e.g. for water balance computations). Lastly, `AverageVector` is defined as a mutable struct which may have a performance penalty.

SouthEndMusic: To me this solution is cleaner, because the averaging happens in many places (and will only happen more with https://github.com/Deltares/Wflow.jl/issues/617) and the extra field that is needed for each instance is consolidated in the struct.

The `AverageVector` is only defined as a mutable struct because of the mutable `fresh_average` Boolean flag, the other fields are marked as `const`.

vers-w: Hm, yes and I would recommend the more explicit approach, even when it requires more field definitions (only the cumulative part needs to added). What do you think @JoostBuitink ?

Wflow/src/snow/snow.jl
```diff
@@ -131,14 +129,14 @@ end
 
 "Update snow HBV model for a single timestep"
 function update_snow_model!(
-    snow_model::SnowHbvModel,
+    model::SnowHbvModel,
```
JoostBuitink: Is there a reason why you revert this back to model (as this was recently changed to snow_model if I remember correctly)?

Wflow/src/glacier/glacier.jl
```diff
@@ -72,33 +72,36 @@ function GlacierHbvParameters(
...
         "glacier_surface__area_fraction",
         LandHydrologySBM;
         sel = indices,
     )
-    max_snow_to_glacier = 8.0 * (dt / BASETIMESTEP)
-    glacier_hbv_params =
-        GlacierHbvParameters(; g_ttm, g_cfmax, g_sifrac, glacier_frac, max_snow_to_glacier)
-    return glacier_hbv_params
+    # 8 mm dt⁻¹
+    max_snow_to_glacier = 8.0e-3 * tosecond(dt)
```
JoostBuitink: I don't think this is consistent with the old setting, which was 8 mm/day (converted to mm/dt). Now it looks like it is set to be 8 mm/dt, is that right?

JoostBuitink: I am also thinking, should we make this a parameter instead of "hiding" it in the code?

Wflow/src/glacier/glacier_process.jl
```diff
@@ -9,11 +9,11 @@ occurs if the snow storage < 10 mm.
...
-- `temperature` air temperature [°C]
-- `ttm` temperature threshold for ice melting [°C]
-- `cfmax` ice degree-day factor in [mm/(°C/day)]
+- `glacier_store` volume of the glacier [mm => m] w.e.
+- `snow_storage` snow storage on top of glacier [mm => m]
+- `temperature` air temperature [°C => K]
+- `ttm` temperature threshold for ice melting [°C => K]
+- `cfmax` ice degree-day factor in [mm °C⁻¹ day⁻¹ => m K⁻¹ s⁻¹]
 - `g_sifrac` fraction of the snow turned into ice [-]
 - `max_snow_to_glacier` maximum snow to glacier conversion rate
```
JoostBuitink: Good to also add the unit of this variable

Wflow/src/glacier/glacier_process.jl
```diff
@@ -9,11 +9,11 @@ occurs if the snow storage < 10 mm.
...
-- `snow_storage` snow storage on top of glacier [mm]
-- `temperature` air temperature [°C]
-- `ttm` temperature threshold for ice melting [°C]
-- `cfmax` ice degree-day factor in [mm/(°C/day)]
+- `glacier_store` volume of the glacier [mm => m] w.e.
+- `snow_storage` snow storage on top of glacier [mm => m]
+- `temperature` air temperature [°C => K]
+- `ttm` temperature threshold for ice melting [°C => K]
+- `cfmax` ice degree-day factor in [mm °C⁻¹ day⁻¹ => m K⁻¹ s⁻¹]
 - `g_sifrac` fraction of the snow turned into ice [-]
```
JoostBuitink: this variable has [s-1] unit

Wflow/src/vegetation/rainfall_interception.jl
```diff
@@ -74,45 +90,62 @@ function rainfall_interception_modrut(
...
     end
 
     # Drain the canopy_storage again if needed
     if canopy_storage > cmax
-        canopy_drainage = canopy_storage - cmax
+        # [m s⁻¹] = ([m] - [m]) / [s]
+        canopy_drainage = (canopy_storage - cmax) / dt
         canopy_storage = cmax
         throughfall += canopy_drainage
     end
```
JoostBuitink, line 121: This is the same logic as the code a couple of lines above (L114-L122), but has a slightly different formulation. Good to make it consistent (perhaps move to a small function?

Wflow/test/run_sbm_piave.jl
```diff
@@ -231,33 +260,43 @@ end
...
+        @test recharge.variables.rate[1] ≈ to_SI(-0.0002922905062717429, M3_PER_DAY)
+        @test mean(recharge.variables.rate) ≈ to_SI(0.0009271689030318317, M3_PER_DAY)
     end
 
     Wflow.run_timestep!(model)
 
     @testset "Second timestep" begin
         @test subsurface_flow.variables.head[1] ≈ 1.563147470676254
-        @test mean(subsurface_flow.variables.head) ≈ 1106.4867249666993
+        @test mean(subsurface_flow.variables.head) ≈ 1106.4948788348809
```
JoostBuitink: Why did this value change? All other values are (as far as I see) identical to the previous test values, so what would cause this difference?

Wflow/test/soil.jl
```diff
@@ -152,79 +257,132 @@ end
...
         theta_r = [0.17089612782001495],
-        hb = [-10.0],
-        h1 = [0.0],
-        h2 = [-100.0],
-        h3 = [1000.0],
-        h4 = [-16000.0],
+        hb = [to_SI(-10.0, "soil_water__air_entry_pressure_head")],
+        h1 = [to_SI(0.0, "vegetation_root__feddes_critical_pressure_head_h1")],
+        h2 = [to_SI(-100.0, "vegetation_root__feddes_critical_pressure_head_h2")],
+        h3 = [to_SI(-1000.0, "vegetation_root__feddes_critical_pressure_head_h1")],
```
JoostBuitink: Not really wrong, but maybe better to grab the h3_low standard name, to consistent with the something closer to the `h3` value?

```suggestion
        h3 = [to_SI(-1000.0, "vegetation_root__feddes_critical_pressure_head_h3_low")],
```

Wflow/test/soil.jl
```diff
@@ -152,79 +257,132 @@ end
...
         N;
-        runoff = [0.0],
-        zi = [1244.5135404970033],
-        ustorelayerthickness = [SVector((100.0, 300.0, 800.0, 44.513540497003305))],
+        runoff = [to_SI(0.0, MM_PER_DT; dt_val = dt)],
+        zi = [to_SI(1244.5135404970033, "soil_water_saturated_zone_top__depth")],
+        ustorelayerthickness = [
+            to_SI.(
+                SVector((100.0, 300.0, 800.0, 44.513540497003305)),
+                Ref("soil_layer_water_unsaturated_zone__depth"),
```
JoostBuitink: Perhaps a stupid question, but why do you need the `Ref(...)` in the cases where you pass a SVector? In other `to_SI()` calls, you can just pass the standard_name

Wflow/test/reservoir.jl
```diff
@@ -102,25 +114,103 @@ end
...
+    @test reservoir.boundary_conditions.actual_external_abstraction_av.cumulative_material[1] ≈
+          1e3
+    @test reservoir.variables.storage[1] ≈ 4.4998100277777776e7
+    @test reservoir.variables.waterlevel[1] ≈ 0.9683379629629354
+    @test reservoir.variables.outflow[1] ≈ 1.0
 end
 
 @testitem "unit: update_reservoir_model!" begin
     using Graphs: DiGraph, add_edge!
+    using Wflow: AerageVector, get_average, to_SI, MM_PER_DT
```
JoostBuitink: Typo, but can probably be removed as well as it is not used
```suggestion
    using Wflow: AverageVector, get_average, to_SI, MM_PER_DT
```

Wflow/src/routing/surface/surface_kinwave.jl
JoostBuitink: I struggle a bit to form a clear conclusion, as I see the benefits and downsides of both approaches. I leaning more towards the approach from @vers-w (but that might also be since that is closer to the current, and therefore more comfortable ;-), approach). 

Also, I am not fully sure about logic like this (taken from the water_balance.jl script):
```julia
qin_av_average = get_average(qin_av)
q_av_average = get_average(q_av)
```

This is also since the fact that the variables are named `{var}_av`, but the `av` does no longer refer to "average", as the averaging still need to be performed. I cannot oversee the changes required for #617, but can imagine that operations like these need to happen much more frequent, requiring a more elegant solution..

Wflow/src/domain.jl
```diff
@@ -306,7 +306,7 @@ function get_river_fraction(
     dataset::NCDataset,
     config::Config,
     network::NetworkLand,
-    river_location::Vector{Bool},
+    river_location::Union{Vector{Bool},BitVector},
```
JoostBuitink: I might overlook it, but why is this change required?

Wflow/src/io.jl
```diff
@@ -720,161 +713,152 @@ function Writer(config, modelmap, domain, nc_static; extra_dim = nothing)
...
-                dataset[name][:, time_index] .= v
+            if haskey(output_dataset, extra_dim.name)
+                v = to_SI(reducer_(getindex.(A, layer)), unit; dt_val)
+                output_dataset[name][:, time_index] .= v
             else
                 nlayer = length(first(A))
                 for i in 1:nlayer
-                    v = nt.reducer(getindex.(A, i))
-                    dataset[name][:, i, time_index] .= v
+                    v = to_SI(reducer_(getindex.(A, layer)), unit; dt_val)
```
JoostBuitink: Should these two `to_SI` calls not be `from_SI`?

Wflow/src/demand/water_demand.jl
```diff
@@ -151,7 +149,7 @@ get_demand_gross(nonpaddy_model::NonPaddyModel) = nonpaddy_model.variables.deman
 get_demand_gross(irrigation_model::NoIrrigationNonPaddyModel) = Zeros(irrigation_model.n)
 
 """
-    update_demand_gross!(nonpaddy_model::NonPaddyModel, soil_model::SbmSoilModel)
+    update_demand_gross!(nonpaddy_model::NonPaddy, soil::SbmSoilModel, dt::Float64)
```
JoostBuitink: This docstring does not match the function
```suggestion
    update_demand_gross!(nonpaddy_model::NonPaddyModel, soil_model::SbmSoilModel, dt::Float64)
```

Wflow/src/demand/water_demand.jl
```diff
@@ -342,47 +356,53 @@ get_demand_gross(paddy_model::PaddyModel) = paddy_model.variables.demand_gross
 get_demand_gross(paddy_model::NoIrrigationPaddyModel) = Zeros(paddy_model.n)
 
 """
-    evaporation!(paddy_model::PaddyModel, potential_evaporation)
+    evaporation!(paddy_model::Paddy, potential_evaporation, dt::Float64)
```
JoostBuitink, line 345:
```suggestion
    evaporation!(paddy_model::PaddyModel, potential_evaporation, dt::Float64)
```

Wflow/src/demand/water_demand.jl
```diff
@@ -342,47 +356,53 @@ get_demand_gross(paddy_model::PaddyModel) = paddy_model.variables.demand_gross
...
     end
-    return nothing
 end
-update_runoff!(::NoIrrigationPaddyModel, ::Any) = nothing
+
+update_runoff!(paddy_model::NoIrrigationPaddyModel, runoff, dt::Float64) = nothing
 
 """
-    update_demand_gross!(paddy_model::PaddyModel)
+update_demand_gross!(paddy_model::Paddy, dt::Float64)
```
JoostBuitink, line 386:
```suggestion
update_demand_gross!(paddy_model::PaddyModel, dt::Float64)
```

Wflow/src/demand/water_demand.jl
```diff
@@ -1028,27 +1104,35 @@ function update_demand_gross!(demand_model::DemandModel)
     return nothing
 end
 
-update_demand_gross!(demand_model::NoDemandModel) = nothing
+update_demand_gross!(demand_model::NoDemandModel, dt::Float64) = nothing
 
 """
-    update_water_demand_model!(demand_model::DemandModel, soil_model::SbmSoilModel)
+    update_water_demand_model!(demand_model::DemandModel, soil::SbmSoilModel, dt::Float64)
```
JoostBuitink:
```suggestion
    update_water_demand_model!(demand_model::DemandModel, soil_model::SbmSoilModel, dt::Float64)
```

Wflow/src/routing/subsurface/groundwater.jl
```diff
@@ -424,11 +451,17 @@ function conductance(
...
+        # [m] = [m] - [m]
         thickness1 = gwf.parameters.top[i] - gwf.parameters.bottom[i]
+        # [m] = [m] - [m]
         thickness2 = gwf.parameters.top[j] - gwf.parameters.bottom[j]
         # calculate conductivity values corrected for depth of water table
+        # TODO: Only when the parameter f is assumed unitless does this work out, but f has unit [m⁻¹]
+        # [m s⁻¹] = ([m s⁻¹] / [-]) * ([-] - [-])
         k1 =
             (gwf.parameters.k[i] / gwf.parameters.f[i]) *
             (exp(-gwf.parameters.f[i] * zi1) - exp(-gwf.parameters.f[i] * thickness1))
```
JoostBuitink: I think we are calculating kH1 [m3 s-1] here. Then the units are correct, with f in [m-1]

JoostBuitink: also for k2 below

Wflow/src/routing/surface/surface_kinwave.jl
```diff
@@ -643,15 +683,15 @@ end
...
 # is added to the river kinematic wave.
 get_inflow_reservoir(
-    ::KinWaveRiverFlowModel,
+    river_flow_model::KinWaveRiverFlowModel,
     overland_flow_model::KinWaveOverlandFlowModel,
     inds::Vector{Int},
-) = overland_flow_model.variables.q_av[inds]
+) = get_average(overland_flow_model.variables.q_av)[inds]
 get_inflow_reservoir(
     ::KinWaveRiverFlowModel,
```
JoostBuitink: to be consistent with the change made a couple of lines above?
```suggestion
    river_flow_model::KinWaveRiverFlowModel,
```

Wflow/src/routing/surface/surface_local_inertial.jl
```diff
@@ -232,20 +256,53 @@ end
...
+    river_flow_model::LocalInertialRiverFlowModel,
     overland_flow_model::KinWaveOverlandFlowModel,
     inds::Vector{Int},
-) = overland_flow_model.variables.q_av[inds] .+ overland_flow_model.variables.to_river[inds]
+) =
+    get_average(overland_flow_model.variables.q_av)[inds] .+
+    get_average(overland_flow_model.variables.to_river)[inds]
 
 get_inflow_reservoir(
     ::LocalInertialRiverFlowModel,
```
JoostBuitink: to be consistent with other changes?
```suggestion
    river_flow_model::LocalInertialRiverFlowModel,
```

Wflow/src/routing/surface/surface_local_inertial.jl
```diff
@@ -1105,19 +1263,23 @@ Returns tuple: (inflow, abstraction_to_add)
     river_idx::Int,
     dt::Float64,
 )
+    # [m³ s⁻¹] < [m³ s⁻¹]
     if river_flow_model.boundary_conditions.external_inflow[river_idx] < 0.0
+        # TODO: This is quite clearly wrong; in the `if` case a depth is
+        # assigned to a volume
```
JoostBuitink: This is currently being fixed in in #889, so this ToDo can be removed once that is PR is approved and merged

JoostBuitink (changes requested):
Note: incomplete review, as I have not yet had the time to look at the sediment part.

Thanks for all of this work! Changes look good, and I really like your addition of including the units in comments above the computations, that help a lot!

A couple of general comments/questions:
- Did you check the effect on model run time (e.g. running the Moselle with and without these changes)?
- There are a couple of comments open from Copilot, which seem valid, so I have not made additional comments.
- Formatting changes: sometimes `a=b` and other locations `a = b`. I believe there is an issue/PR for this open, but just to be aware that there are some small inconsistencies on this front within this PR

Wflow/src/routing/surface/surface_kinwave.jl
SouthEndMusic: Yeah, `*_av_average` isn't pretty. Ok; I'll get rid of the `AverageVector`

Wflow/src/domain.jl
SouthEndMusic: It's probably not :)

Wflow/src/glacier/glacier.jl
SouthEndMusic: @vers-w

Wflow/src/io.jl
SouthEndMusic: Good catch! This also shows that testing the output files wouldn't be a bad idea

Wflow/src/routing/subsurface/groundwater.jl
SouthEndMusic: I don't understand what you are saying, can you refer to the docs or litarature?

Wflow/src/snow/snow.jl
SouthEndMusic: This was probably due to an incorrect merge

Wflow/test/soil.jl
SouthEndMusic: Here `to_SI` is being applied to an `SVector` using broadcasting, so Julia will try to iterate each argument, which is prevented by the `Ref`.  For vectors which are mutable the in-place `to_SI!` is applied.

SouthEndMusic:
@JoostBuitink  regarding performance: I saw that the unit conversion functions take up a small (< 10%) part of the runtime, I didn't see a significant change in runtime with a quick comparison with the `sbm_river-floodplain-local-inertial` model

Wflow/src/glacier/glacier_process.jl
```diff
@@ -9,13 +9,13 @@ occurs if the snow storage < 10 mm.
...
 # Arguments
 - `glacier_frac` fraction covered by glaciers [-]
-- `glacier_store` volume of the glacier [mm] w.e.
-- `snow_storage` snow storage on top of glacier [mm]
-- `temperature` air temperature [°C]
-- `ttm` temperature threshold for ice melting [°C]
-- `cfmax` ice degree-day factor in [mm/(°C/day)]
-- `g_sifrac` fraction of the snow turned into ice [-]
-- `max_snow_to_glacier` maximum snow to glacier conversion rate
+- `glacier_store` volume of the glacier [mm => m] w.e.
```
vers-w: For clarity I would suggest to only add here the internal units and remove the external units (also for other functions).

Wflow/src/glacier/glacier_process.jl
```diff
@@ -33,27 +33,35 @@ function glacier_hbv(
     cfmax,
     g_sifrac,
     max_snow_to_glacier,
+    dt,
 )
 
     # Fraction of the snow transformed into ice (HBV-light model)
+    # [m s⁻¹]
```
vers-w: I understand that this helps with checking the units while working on this. However, I think this is too verbose as units are already specified in the functions and structs (also for other computations). For internal function variables (that are not computed) of course fine to add unit info!

Wflow/src/glacier/glacier.jl
```diff
@@ -3,9 +3,9 @@ abstract type AbstractGlacierModel end
 "Struct for storing glacier model variables"
 @with_kw struct GlacierVariables
     n::Int
-    # Water within the glacier [mm]
+    # Water within the glacier [mm => m]
```
vers-w: I would suggest for clarity to only report here the internal units (also for other structs). External units are already included in metadata.

Wflow/src/routing/surface/surface_process.jl
```diff
@@ -1,20 +1,28 @@
+const tan80 = 5.67
```
vers-w: I thought we decided to use upper case name for constants?

Wflow/src/routing/utils.jl
```diff
@@ -72,34 +72,57 @@ network is expected to hold a graph and order field, where the graph implements
 interface, and the order is a valid topological ordering such as that returned by
 `Graphs.topological_sort_by_dfs`.
 """
-function accucapacityflux!(flux, material, network, capacity)
+function accucapacityflux!(flux, material, network, capacity, dt; material_is_flux = false)
```
vers-w: I think it is a bit confusing to allow `material` as a flux input (if I understand correctly). It seems this is only used for sediment transport (two function calls). Why not make the input there `material` for clarity?

Wflow/src/utils.jl
```diff
@@ -193,25 +188,15 @@ function set_states!(
...
                 A = nomissing(A)
-                # Convert to desired type if needed
-                if !isnothing(type)
-                    if eltype(A) != type
-                        A = map(type, A)
-                    end
-                end
+                A = apply_unit_and_type_transform!(A, metadata; dt_val)
                 # set state in model object, only set active cells ([1:n]) (ignore boundary conditions/ghost points)
                 lens = get_metadata(state, typeof(land), Routing; model).lens
```
vers-w, line 201: Although this change is not part of this PR, I am wondering why `Routing` is added here as argument?

vers-w (changes requested):
Nice work!
Below a couple of comments.

Wflow/src/glacier/glacier.jl
SouthEndMusic: Fair point, I wanted to do that eventually but leave it for a while for clarity. I'll remove the old  units now.

Wflow/test/soil.jl
vers-w: I also don't understand the use of `Ref` here. If I understand correctly you're mapping to this function in `testing_utils.jl`:

`to_SI(x, Wflow.get_metadata(name).unit; kwargs...)`

Wflow/src/glacier/glacier_process.jl
SouthEndMusic: It's not exactly clear to me what you think is too verbose. My policy here is to be as explicit as possible, as this makes the code much more accessible to people who are not familiar with the hydrological concepts

Wflow/src/routing/surface/surface_process.jl
SouthEndMusic: Fair point 👍

Wflow/src/glacier/glacier_process.jl
vers-w: Yeah, my point is that the units are already clear from the code comments for struct fields and function arguments. So, whether a person is familiar with the hydrological concepts or not, based on that information (and that SI units are used internally) a person should be able to follow this. 

Additionally, I also haven't seen this in other similar software packages (that are most probably developed by people that have and do not have domain expertise) unless there is for example an internal function variable introduced (not computed but given), like `snow_storage_max` in the `lateral_snow_transport!` function.

Wflow/src/routing/subsurface/groundwater.jl
SouthEndMusic: There was a discrepancy between `horizontal conductance` and `conductance`. What is computed here should be called `kH1, kH2`, then it is consistent again with `horizontal_conductance` and what is expected by `harmonicmean_conductance`.

Wflow/src/utils.jl
SouthEndMusic: Fair question; I think this was done here to disambiguate the metadata for standard names that appear in multiple standard name maps because they have different meanings for different model types (which I think is bad practice). These for instance appear both in the routing and sediment standard name maps:

```
"reservoir_surface__area"
"reservoir_location__count"
"land_surface_water__volume_flow_rate"
"river_water__volume_flow_rate"
"river_water__depth"
"land_surface_water__depth"
```

Wflow/src/glacier/glacier_process.jl
SouthEndMusic: I still think it's good practice; I can say from experience that this improves code clarity. The glacier process isn't a great example for this because it is very simple. I would say that anything that lowers the barrier to entry for contribution is a good thing, even if it just something like not having to search the codebase for what the unit of some value is. It gives an immediate insight into what is being simulated physically, just as https://github.com/Deltares/Wflow.jl/pull/886 clarifies where on the domain it happens.

Wflow/test/soil.jl
SouthEndMusic: Note that eventually I want to remove all the `to_SI` calls from the tests and just put in the values. I added all these `to_SI` calls on the old values to verify that the results are the same up to the unit transformation

Wflow/src/utils.jl
vers-w: I don't think these have different meanings (strictly) for different model types, right? These are stored differently within the model, is that what you mean?

Wflow/src/glacier/glacier_process.jl
vers-w: I think all these units in the comments are more distracting than improving code clarity... Besides the glacier process, there are many more instances that are very simple, and I would say for the few bit more complex cases it is fine to search the code base or docs to get a better understanding, you even may run into something new during the search :wink:.

Wflow/src/utils.jl
SouthEndMusic: I did a full search, these are the standard names which are in more than one standard name map:

```julia
--------------------------------------------------
`reservoir_surface__area` is in these standard_name_maps: ["sediment", "routing"]; descriptions:
"Reservoir surface area"
"Area of the reservoir"
--------------------------------------------------
`compacted_soil__area_fraction` is in these standard_name_maps: ["sbm", "sediment"]; descriptions:
"Fraction of compacted soil area"
"Fraction of the soil that is covered (eg paved, snow, etc)"
--------------------------------------------------
`vegetation_canopy__gap_fraction` is in these standard_name_maps: ["sbm", "sediment"]; descriptions:
"Canopy gap fraction"
"Canopy gap fraction"
--------------------------------------------------
`land_surface_water__volume_flow_rate` is in these standard_name_maps: ["sediment", "routing"]; descriptions:
"Overland flow discharge"
"Discharge overland flow"
--------------------------------------------------
`land_surface_water__depth` is in these standard_name_maps: ["sediment", "routing"]; descriptions:
"Overland flow water depth"
"Water depth"
--------------------------------------------------
`land_water_allocation_area__count` is in these standard_name_maps: ["sbm", "domain"]; descriptions:
"Allocation areas"
"Water allocation area ids"
--------------------------------------------------
`atmosphere_water__precipitation_volume_flux` is in these standard_name_maps: ["sbm", "sediment"]; descriptions:
"Precipitation"
"Precipitation"
--------------------------------------------------
`reservoir_location__count` is in these standard_name_maps: ["sediment", "routing"]; descriptions:
"Reservoir location ids"
"Outlet of the reservoirs in which each reservoir has a unique id"
--------------------------------------------------
`vegetation_canopy_water__interception_volume_flux` is in these standard_name_maps: ["sbm", "sediment"]; descriptions:
"Interception"
"Rainfall interception by the vegetation"
--------------------------------------------------
`river_water__volume_flow_rate` is in these standard_name_maps: ["sediment", "routing"]; descriptions:
"River discharge"
"River (+ floodplain for local inertial routing) discharge"
--------------------------------------------------
`river_water__depth` is in these standard_name_maps: ["sediment", "routing"]; descriptions:
"River water depth"
"River water depth"
```

These descriptions are all equivalent I would say. I would happily get rid of the duplicates, that would simplify the lookup logic (i.e. being able to merge all metadata into a single dict). Regarding having different lenses for different model types: with https://github.com/Deltares/Wflow.jl/pull/906 that is no longer an issue, that decouples the lookup from the model data structure

Wflow/src/glacier/glacier_process.jl
JoostBuitink: Reading this discussion made me reconsider my own comment in my own review. Initially I wrote that I like having the units above each line, but I realize now I used it mainly to help with reviewing, to understand if all units are correctly converted. However, since everything is now in SI units, there is not much differentiation between the units of the variables anymore (mostly flux [m/s] or storage [m]), so I am wondering a bit about the added value of these comments now.

docs/model_docs/routing/sediment_flux.qmd
```diff
@@ -174,7 +174,7 @@ $$
 
 where $C_w$ is the sediment concentration by weight, $\SIb{\rho}{g\ m^{-3}}$ and
 $\SIb{\rho_{s}}{g\ m^{-3}}$ are the fluid and sediment density (here respectively equal to
-$\SI{1000}{g\ m^{-3}}$ and $\SI{2650}{g\ m^{-3}}$), $\SIb{u}{m\ s^{-1}}$ is the water mean
+$\SI{1.0 \times 10^6}{g\ m^{-3}}$ and $\SI{2.65 \times 10^6}{g\ m^{-3}}$), $\SIb{u}{m\ s^{-1}}$ is the water mean
```
vers-w, line 177: Is this correct `\times 10^6` ? 

One general question about the docs, should we now also use SI units throughout the docs describing the different concepts including equations (I think this mainly applies to the Model documentation section)?

SouthEndMusic: I'm pretty sure regarding the unit changes in the docs; water has a density of about 1 kg/L and so 10^6 g/m^3.

Regarding the units used in the docs: I think it's good as it is. These docs are meant for users which are probably most familiar with the units used in the literature. The SI units are purely internal.

Wflow/src/glacier/glacier_process.jl
vers-w: I noticed some (but not all) unit annotations were removed in https://github.com/Deltares/Wflow.jl/pull/735/commits/3072af3b0a427c28a4dab2c130debe080785ae59. I would still recommend to remove all as suggested in previous comments.

SouthEndMusic: The remaining unit annotations were removed 👍

Wflow/test/groundwater.jl
```diff
@@ -170,24 +170,19 @@ end
         conductivity_profile = Wflow.GwfConductivityProfileType.uniform
         Wflow.flux!(gwf_model, conductivity_profile, dt)
         # KD is based on upstream saturated thickness, i.e. 7.5 m and 20.0 m (which is capped to 10.0)
-        @test gwf_model.variables.q_net == [562.5, 687.5, -1250.0]
+        @test gwf_model.variables.q_net ≈ [562.5, 687.5, -1250.0]
```
vers-w: These values were in m3/day, suggest to change this test (including `dt`) so these values are now in m3/s.

Wflow/test/groundwater.jl
```diff
@@ -170,24 +170,19 @@ end
...
-            bottom = [1.0, 1.0],
-        )
-        variables = Wflow.GwfRiverVariables(;
-            n,
-            stage = [2.0, 2.0],
-            storage = [20.0, 20.0],
-            flux = [0.0, 0.0],
-            flux_av = [0.0, 0.0],
+            infiltration_conductance = fill(100.0, n),
+            exfiltration_conductance = fill(200.0, n),
```
vers-w: As suggested in previous comment, convert to SI units here so these values are realistic.

Wflow/test/groundwater.jl
```diff
@@ -204,7 +199,7 @@ end
         n = 2
         parameters =
             Wflow.DrainageParameters(; elevation = [2.0, 2.0], conductance = [100.0, 100.0])
```
vers-w: Use SI units for conductance.

Wflow/test/groundwater.jl
```diff
@@ -251,7 +243,8 @@ end
         variables = Wflow.WellVariables(;
             volumetric_rate = [-1000.0],
```
vers-w: Suggest to convert to SI unit.

Wflow/test/groundwater.jl
vers-w: Convert to seconds.

Wflow/test/groundwater.jl
```diff
@@ -233,12 +229,8 @@ end
     @testset "recharge" begin
         dt = 1.0
```
vers-w: Convert to seconds.

Wflow/test/groundwater.jl
```diff
@@ -170,24 +170,19 @@ end
         conductivity_profile = Wflow.GwfConductivityProfileType.uniform
         Wflow.flux!(gwf_model, conductivity_profile, dt)
         # KD is based on upstream saturated thickness, i.e. 7.5 m and 20.0 m (which is capped to 10.0)
-        @test gwf_model.variables.q_net == [562.5, 687.5, -1250.0]
+        @test gwf_model.variables.q_net ≈ [562.5, 687.5, -1250.0]
     end
 
     @testset "river" begin
         dt = 1.0
```
vers-w: convert to seconds.

Wflow/test/groundwater.jl
vers-w: convert to seconds

Wflow/test/routing_process.jl
```diff
@@ -97,45 +96,65 @@ end
...
     include("testing_utils.jl")
+    DAY = Unit(; d = 1)
+    M_PER_DAY = Unit(; m = 1, d = -1)
+    M2_PER_DAY = Unit(; m = 2, d = -1)
+
+    ### Shared values
     n = 1
     N = 4
+    act_thickl = [to_SI.(SVector(100.0, 300.0, 800.0, 800.0), Ref(MM))]
+    sumlayers = [to_SI.(SVector(0.0, 100.0, 400.0, 1200.0, 2000.0), Ref(MM))]
```
vers-w: I guess `to_SI` will be removed (I think you did make a comment about this)?

Wflow/test/routing_process.jl
```diff
@@ -333,21 +766,20 @@ end
...
 
-    Wflow.kinwave_river_update!(river_flow_model, domain, dt, dt_forcing)
+    Wflow.kinwave_river_update!(river_flow_model, domain, dt)
 
     @test dt ≈ 994.6119029285007
     @test river_flow_model.variables.q ≈ [0.37903337592185243, 3.1969698861305855]
     @test river_flow_model.variables.h ≈ [0.01500830539624011, 0.05407828963124342]
     @test river_flow_model.variables.storage ≈ [1506.7893805755937, 4876.828625285123]
-    @test river_flow_model.variables.q_av ≈ [376.9911072990474, 3179.744302049454]
+    @test river_flow_model.variables.q_cumulative ≈ [376.9911072990474, 3179.744302049454]
```
vers-w, line 760: Suggest to test here for `q_av` (also for other tests).

Wflow/test/routing_process.jl
```diff
@@ -333,21 +766,20 @@ end
...
 
     (; reservoir) = river_flow_model.boundary_conditions
     @test reservoir.variables.waterlevel[1] ≈ 29.654579645252387
     @test reservoir.variables.storage[1] ≈ 4.443628667214138e7
     @test reservoir.variables.outflow[1] ≈ 3.0009999145314317
-    @test reservoir.variables.actevap[1] ≈ 0.005295387542208319
+    @test reservoir.variables.actevap_cumulative[1] ≈ 5.295387542208319e-6
```
vers-w, line 766: Suggest to test here for `actevap_average` (also for other tests).

Wflow/test/run_sbm.jl
```diff
@@ -73,75 +79,510 @@
...
     end
 
     @testset "river flow" begin
-        (; q_av) = model.routing.river_flow.variables
-        @test sum(q_av) ≈ 3696.5789619579173
-        @test q_av[1622] ≈ 0.0007502850311515928
-        @test q_av[43] ≈ 11.458528971675033
-        @test q_av[domain.river.network.order[end]] ≈ 0.04397648668235761
+        q = model.routing.river_flow.variables.q_average
+        @test sum(q) ≈ 3696.5789619579173
```
vers-w: This test is failing, the difference is relatively large (~0.6 m3/s). The tests for the vertical part are not failing, right? I think this needs clarification.

Wflow/test/run_sbm.jl
```diff
@@ -73,75 +79,510 @@
...
+        @test q[1622] ≈ 0.0007502850311515928
+        @test q[43] ≈ 11.458528971675033
```
vers-w: Also relatively large difference (see comment above).

Wflow/test/routing_process.jl
```diff
@@ -149,36 +168,92 @@ end
...
         ssfmax,
         kh_profile,
-        soil,
+        soil_model,
+        i,
+    )
+    @test ssf ≈ 0.23130576097772237
+    @test zi ≈ 0.1656875455413981
+    @test exfilt ≈ 0.0
+    @test net_flux ≈ -7.808554363456962e-7
```
vers-w: This test fails with relatively large difference, `net_flux = -3.904277181728481e-7`. Needs clarification.

Wflow/test/routing_process.jl
```diff
@@ -228,38 +303,394 @@ end
...
+    # update_river_cell_storage_and_depth!
+    @test Wflow.compute_river_storage_change(
+        overland_flow_model,
+        river_flow_model,
+        domain,
+        1,
+        dt,
+    ) ≈ 290.0
+    @test Wflow.compute_external_inflow(river_flow_model, overland_flow_model, 1, 1, dt) |>
+          collect ≈ [-8.0e-4, 8.0e-4]
```
vers-w: From the test `Evaluated: [-0.2, 0.2] ≈ [-0.0008, 0.0008]` ?

Wflow/test/routing_process.jl
```diff
@@ -228,38 +303,394 @@ end
...
+
+    Wflow.update_river_cell_storage_and_depth!(
+        overland_flow_model,
+        river_flow_model,
+        domain,
+        1,
+        dt,
+    )
+    @test river_flow_model.variables.h[1] ≈ river_h_expected
+    @test overland_flow_model.variables.h[1] ≈ land_h_expected
```
vers-w: From the test: `Evaluated: 0.001092592592592645 ≈ 0.001461481481481508`?

Wflow/test/run_sbm_gwf.jl
vers-w, line 40: Quite a difference: `Evaluated: 1.9427839461598195 ≈ 1.6466006412729106` ,needs clarification.

Wflow/test/run_sbm_gwf.jl
```diff
@@ -59,18 +63,19 @@
         @test soil.parameters.theta_s[1] ≈ 0.44999998807907104
         @test soil.variables.runoff[1] == 0.0
         @test soil.variables.soilevap[1] == 0.0
```
vers-w, line 62: `Evaluated: 7.447808324668021e-9 == 0.0`, needs clarification.

Wflow/test/routing_process.jl
SouthEndMusic: Yes, that will be removed eventually

Wflow/test/routing_process.jl
SouthEndMusic: That doesn't work; previously what was tested was also the cumulative value before averaging. Averaging happens over the routing sub timesteps, this tests only one sub timestep

Wflow/test/routing_process.jl
SouthEndMusic: See my response to the previous comment

vers-w:
General comment about the tests: there are still quite some large differences for the routing (some tests fail, while the tests for the vertical part succeed), this needs clarification.

Wflow/test/routing_process.jl
SouthEndMusic: Not sure how I came to those values, The values computed here are the same on master

Wflow/test/routing_process.jl
SouthEndMusic: Aligned these with master too

Wflow/test/routing_process.jl
SouthEndMusic: As explained by Claude:

> The test expected values for net_flux were computed before the /its averaging fix was applied. On master, sub-stepping accumulated a total [m] and returned net_flux = net_flux_sum directly. The caller used it as-is. In this PR, net_flux is now a rate [m/s], and the caller multiplies by dt to get the integrated value. With sub-stepping, the correct return is the time-averaged rate: net_flux_sum / its. Since its = 2 for this test case, the expected values were exactly 2× too large. Updated from -7.808554363456962e-7 to -3.904277181728481e-7.

Wflow/test/run_sbm.jl
SouthEndMusic: I'm not sure there is any explanation other than an accumulation of small floating point differences, and some places where there can be a discrete change in behaviour due to e.g. rounding.

Wflow/src/sediment/erosion/erosion_process.jl
```diff
@@ -41,29 +41,31 @@ function rainfall_erosion_eurosem(
...
+    # Total kinetic energy by rainfall
+    E_kin_tot = rainfall_depth_direct * E_kin_direct + rainfall_depth_leaf * E_kin_leaf
+    rainfall_erosion =
+        area * soil_detachability * E_kin_tot * exp(-eurosem_exponent * waterlevel) / dt
 
     # Remove the impervious area
     rainfall_erosion *= 1.0 - soilcover_fraction
-    return rainfall_erosion
+    # TODO: Explain this factor 1e3 needed to pass the tests
+    return rainfall_erosion / 1e3
```
JoostBuitink: regarding the factor 1e3, isn't this related to the 0.001 in L59 in the old code (`ketot = (rddir * kedir + rdleaf * keleaf) * 0.001`) that is no longer done? If I understand it correctly, the computation in these lines (L59 of the old code, L61 in this new code) is working with the same units, since everything before is converted to mm. I think this `*0.001` (or `/1e3`) is to convert from mm to m?

Wflow/src/sediment/sediment_transport/river_transport.jl
```diff
@@ -619,13 +618,15 @@ function update_variables!(
...
 
-    # Leftover sediment for mass balance
-    leftover_clay[v] = input_clay + erosion_clay - deposition_clay - clay_rate[v]
-    leftover_silt[v] = input_silt + erosion_silt - deposition_silt - silt_rate[v]
-    leftover_sand[v] = input_sand + erosion_sand - deposition_sand - sand_rate[v]
-    leftover_sagg[v] = input_sagg + erosion_sagg - deposition_sagg - sagg_rate[v]
-    leftover_lagg[v] = input_lagg + erosion_lagg - deposition_lagg - lagg_rate[v]
-    leftover_gravel[v] = input_gravel + erosion_gravel - deposition_gravel - gravel_rate[v]
+    # Sediment left in the cell
+    # [kg] = ([kg s⁻¹] + [kg s⁻¹] - [kg s⁻¹] - [kg s⁻¹]) * dt
```
JoostBuitink: This line can be removed :)
```suggestion

```

Wflow/src/sediment/sediment_transport/river_transport.jl
```diff
@@ -646,16 +647,18 @@ function update_sediment_river_transport_model!(
 
     # Sediment transport - water balance in the river
     for v in order
-        ### Sediment input in the cell (left from previous timestep + from land + from upstream outflux) ###
-        input_particles = compute_sediment_input(sediment_transport_model, graph, v)
+        ### Sediment input in the cell (left from prevoous time step + from land + from upstream outflux) ###
```
JoostBuitink:
```suggestion
        ### Sediment input in the cell (left from previous time step + from land + from upstream outflux) ###
```

Wflow/test/run_sbm.jl
```diff
@@ -374,17 +818,17 @@ end
...
+        @test q[43] ≈ 11.248731903973153
+        @test q[501] ≈ 3.163313530437266
+        h = model.routing.river_flow.variables.h
         @test h[1622] ≈ 0.00191277920611667
         @test h[43] ≈ 0.447403557220214
         @test h[501] ≈ 0.37810388857945015
-        (; q_channel_av) = model.routing.river_flow.variables
-        @test q_av ≈ q_channel_av
+        q_channel = model.routing.river_flow.variables.q_channel_average
+        @test q ≈ q_channel
```
JoostBuitink, line 805: This test is currently failing, with quite large differences. Please check what is going on here.

Test result (from github logs): 
```
  Expression: q ≈ q_channel
   Evaluated: [1.9270033374406912, 0.42030167476326136, 0.40908309150101096, 2.6002287371080213, 2.3488170821992997, 2.7650154405996865, 2.7471008476720566, 2.6925232818329703, 2.68918869928053, 2.854167329572017  …  0.046905273936870225, 0.029756636483177036, 0.031611965775001236, 0.004619886598613289, 0.010933038999502284, 0.01960880630952823, 0.011591800466742946, 0.00835051400057532, 0.0022145279734597264, 0.005831261738554856] ≈ [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

Wflow/test/run.jl
```diff
@@ -65,13 +63,18 @@
...
             b = endstate_restart[varname][:]
             maxdiff = maximum(abs.(skipmissing(b - a)))
             tol = varname == "volume_reservoir" ? 1e-8 : 1e-9
-            @test maxdiff < tol
+            if maxdiff ≥ tol
+                push!(failed, varname)
+                println("$varname: $maxdiff ≥ $tol")
+            end
         end
+        @test isempty(failed)
```
JoostBuitink, line 77: This test is failing, and it is probably due to floating point errors (and too tight tolerance?), but good to confirm this.

Wflow/test/io.jl
JoostBuitink, line 496: This test is currently failing and erroring, I guess related to the `mktempdir()`?

JoostBuitink (changes requested):
Wrapped up also the sediment review. A couple of small comments/questions

Wflow/src/sediment/sediment_transport/river_transport.jl
vers-w: Would be good to do a final check for these unit annotations, just noticed there are also a few left in surface_process.jl.

Wflow/src/sediment/erosion/erosion_process.jl
SouthEndMusic: Line 61 has units:

```
[J m⁻²] = [mm] * [J m⁻² mm⁻¹] + [mm] * [J m⁻² mm⁻¹]
```

which is the same in the old code, where I don't think the `0.001` factor makes sense either. Could it be that this is a bug in the original code?

Wflow/test/run.jl
SouthEndMusic: I found that for variables with non-trivial unit conversion the trip `endstate -> NetCDF -> warmstate` has small floating point errors because of unit conversion, for this nest I found that `satwaterdepth` and `zi` have errors of magnitude `1e-16`. That could be a seed for further divergence. Is it OK to update the tolerance? @vers-w

Wflow/test/run_sbm.jl
vers-w: For the routing the internal timestepping (adaptive) is not the same anymore. When running this test with fixed internal timestepping it does not fail.

Wflow/test/run.jl
vers-w: I think that's OK, what tolerance update is required?

Wflow/test/run_sbm.jl
SouthEndMusic: This one I was able to fix

Wflow/test/run.jl
SouthEndMusic:
```
h_river: 2.9428215864157892e-8 ≥ 1.0e-9
q_river: 2.848611970307502e-6 ≥ 1.0e-9
```

vers-w: This seems related to a change in internal timestepping (see also https://github.com/Deltares/Wflow.jl/pull/735#discussion_r3303484050). If I run with a fixed internal timestep or use a minimum with adaptive timestepping (not a `quantile`) the test is passing. 

Maybe it is better if we remove the `quantile` from the kinematic wave `stable_timestep` function? This would also be more consistent with the local inertial routing `stable_timestep` function. Also tagging @JoostBuitink for his input on this.

vers-w: Although changing this will have quite some impact on kinematic wave routing performance. Replacing: 
 
`if q[i] >= 0.0` by `if q[i] >= KIN_WAVE_MIN_FLOW` 

in the `stable_timestep` function also results in a passing test.

vers-w:
The logging output is a bit different between input variables/parameters (three examples below):

`[ Info: Set `atmosphere_water__precipitation_volume_flux` [mm Δt⁻¹] using netCDF variable `precip` as forcing parameter.`

```
┌ Info: Set parameter using netCDF variable.
│   parameter = "land_surface__slope"
│   var = Slope
└   unit = m m⁻¹
```

`[ Info: Set `soil_water__air_entry_pressure_head` using default value `-10.0`.`

I would suggest to use the first (more compact) one when there is a netCDF variable available. For the default value, I think it would be good to also include the unit.

Wflow/src/sediment/erosion/erosion_process.jl
JoostBuitink: Yes I am almost certain that the `1e3` is wrong (already in the original code) and is no longer needed. However, that would mean even larger differences between the EUROSEM and ANSWERS options for rainfall erosion (factor 10.000), which is not ideal I think ;)
However, I don't know yet what is causing this difference (most likely either the units of `0.108` or `usle_k` in the equation). As you correctly mention, it is very difficult to track the units, also in literature, os we need to have a good look at these units and how to ensure they are handled correctly. 

I have also asked @vers-w to take a look

Wflow/test/run.jl
JoostBuitink: I think this latter suggestion (KIN_WAVE_MIN_FLOW) makes sense. Do I understand it correctly that changing this (and keeping the `quantile`) results in a passing test?

vers-w: Yes, indeed the change to `KIN_WAVE_MIN_FLOW` and keeping `quantile` results in a passing test.

JoostBuitink: That sounds like a great solution then!

Wflow/src/io.jl
```diff
@@ -540,7 +540,7 @@ function NCReader(config)
             cyclic_nc_times = collect(cyclic_dataset[dimname])
             cyclic_times[par] = timecycles(cyclic_nc_times)
             variable_info(var)
-            @info "Set `$par` [$(get_metadata(par).unit)] using netCDF variable `$ncname` as cyclic parameter, with `$(length(cyclic_nc_times))` timesteps."
+            @info "Set `$par [$(get_metadata(par).unit)]`` using netCDF variable `$ncname` as cyclic parameter, with `$(length(cyclic_nc_times))` timesteps."
```
JoostBuitink: there is one " ` " too many in this line!

Wflow/test/run.jl
SouthEndMusic: OK 👍 That will of course change some of the results in other tests

Wflow/src/sediment/erosion/erosion_process.jl
vers-w: I think the units are OK. The only thing is the unit for `soil_erosion__eurosem_exponent` in standard names, I think that should be `Unit(; mm = -1)` according to the original equation. For ANSWERS we need to know in what unit `usle_k` is expected, I did a bit of searching but so far no clear literature source, the ANSWERS users manual references:

Meyer, L. D. and W. H. Wischmeier. 1969. Mathematical simulation of the processes of soil erosion by water. Transactions of the ASAE 12(6): 754-758.

Wflow/test/io.jl
SouthEndMusic: This was a bit fiddly. I want to run the model with a modified config, namely with the temporary directory to avoid permission problems between tests that use the same model. The easiest way to do this is to load the config first and then modify it in memory. However, we want to test logging, and logging is only done with the method `run(toml_path)`, not the method `run(config)`. My solution now is to create a temporary config on disk with the temporary directory path in it.

Wflow/src/standard_name/standard_name_sediment.jl
vers-w: Based on the equation from literature I think this should be:
```suggestion
        unit = Unit(; mm = -1),
```

SouthEndMusic: There is a discrepancy here: indeed according to the literature you mentioned the default should be `2 mm-1`, which is already there (in value) on master.

However, this parameter is used in `exp(-eurosem_exponent * waterlevel)` where `waterlevel` from `RainfallErosionEurosem` is denoted as having unit `[m]` in the struct (already on master). This `waterlevel` is set from `waterlevel_land` which also has unit `[m]` on master, so that is consistent. So on master already this parameter was interpreted as `2 m-1`, which was wrong.

SouthEndMusic: In case that was a bit vague: your suggestion is correct but will change the results. Since this is going to change the results anyway I will also remove the `/ 1e3` because I don't think it makes sense.

vers-w (approved):

Wflow/src/sediment/erosion/erosion_process.jl
```diff
@@ -79,31 +80,31 @@ end
 Rainfall erosion model based on ANSWERS.
 
 # Arguments
-- `precip` (precipitation [mm Δt⁻¹])
-- `usle_k` (USLE soil erodibility [t ha-1 mm-1])
+- `precip` (precipitation [m s⁻¹])
+- `usle_k` (USLE soil erodibility [-], treated as unitless but is actually [t ha-1 h-1 MJ-1 mm-1])
```
vers-w: I think this should be:

```suggestion
- `usle_k` (USLE soil erodibility [-], treated as unitless but is actually [t ha h ha-1 MJ-1 mm-1])
```

docs/changelog.qmd
```diff
@@ -33,8 +33,19 @@ subsurface_water__instantaneous_volume_flow_rate = "ssf"
 ```
 
 ### Fixed
+- Fixed a bug in EUROSEM rainfall erosion calculation where the exponent was incorrectly
+  specified in units of mm⁻¹ instead of m⁻¹, causing the protective effect of water depth
+  to be overestimated by a factor of 1000. This significantly changes EUROSEM rainfall
+  erosion results.
```
JoostBuitink: Good to mention this! Also, I think it is worth explicitly mentioning the `/1e3` that was wrongly applied to the total rainfall_erosion, as those are two separate fixes (right?)

Wflow/test/io.jl
JoostBuitink: Ah yeah that "makes sense". Maybe good to add a comment to the test why it is done that way? I think we should fix the logging to always occur, so that once that is fixed, we remove this workaround in the test

.gitignore
```diff
@@ -16,6 +16,7 @@
 /Wflow/test/data/
 /tmp/
 Wflow/test/sbm_piave_config-debug.toml
+Wflow/test/sbm_simple_temp.toml
```
JoostBuitink, line 19: Do we want/need to explicitly ignore these tomls? I know the first one was already there, but still ;)

JoostBuitink (approved):
Just a couple of very small nitpicks, but otherwise good to go! Thanks for all this work!

docs/changelog.qmd
vers-w: Actually, these fixes should be mentioned as part of bug fixing for patch release 1.0.3...

server/test/client.jl
```diff
@@ -97,17 +98,17 @@
             name = "soil_water_saturated_zone_top__depth",
             dest = fill(0.0, zi_size),
         )
-        @test mean(request(msg)["value"]) ≈ 283.17139935351275
+        @test mean(request(msg)["value"]) ≈ to_SI(283.17139935351275, MM)
```
vers-w: I thought these `to_SI` as part of tests would be removed eventually?

.gitignore
SouthEndMusic: I think so; they appear only temporarily when running certain tests, but can stick around when those tests fail.

server/test/client.jl
SouthEndMusic: Fair point, forgot to do that here

.gitignore
JoostBuitink: Ah right, very true, I didn't make that connection to the toml being generated in the test. So feel free to ignore this comment :P

PR #756, opened 2025-11-17
Fix reading `rootfraction` parameter

vers-w:


JoostBuitink (approved):
LGTM! Sorry for also not spotting this in my previous review ;)

Issue #764, opened 2025-11-25
Test all lenses in `standard_name_*.jl`


SouthEndMusic:
It would be nice if we can make a test which doesn't initialize a model object, but for each lens walks trough the model structure using `subtypes` for abstract fields, to check whether there exists a model structure which contains the lens path.

Issue #765, opened 2025-11-25
Format code via pre-commit


SouthEndMusic:
Now it depends on the version of the VSCode Julia plugin the developer uses how the code gets formatted. It would be nicer if pre-commit is the be-all-end-all of formatting, with a pinned formatter version.

visr:
> Since todate Wflow model only works with Base.VERSION == v"1.11.3"

Perhaps I missed something, but how do you come to this conclusion? It is true that 1.11.3 is the version we currently develop and run CI on, but it should work with other versions as well. By running `juliaup override set 1.11.3` in the repo we tell `juliaup` to use that version in this repo. Unfortunately that is not respected by VSCode, so for Ribasim we additionally set this:

https://github.com/Deltares/Ribasim/blob/9a513b51c985947fd62a6e7f10c7b88fbd74266a/.vscode/settings.json#L66

That could be set in the Wflow repo as well if there is interest.

PR #730, opened 2025-10-29
Use drainable and dynamic fillable porosity for water table change

vers-w:
## Issue addressed
Fixes #180

## Explanation
Drainable porosity/specific yield defined as the difference between `theta_s` (saturated water content) and `theta_fc` (water content at field capacity, new parameter for the `sbm` soil model) is now used to compute the falling water table of subsurface flow concepts `LateralSSF` and `UnconfinedAquifer`. For the rising water table of these concepts a dynamic fillable porosity (based on `theta_s` and the water content of unsaturated soil layers) is used. This results in a better integration of the `sbm` soil model (unsaturated zone) with these subsurface flow concepts as specific yield is typically represented by drainable porosity for these concepts and not by the effective porosity (difference between saturated and residual water content) in the current implementation.

## Checklist
- [x] Updated tests or added new tests
- [x] Branch is up to date with `master`
- [x] Tests & pre-commit hooks pass
- [x] Updated documentation if needed
- [x] Updated changelog.qmd if needed

## Additional Notes (optional)
Add any additional notes or information that may be helpful.

PR #783, opened 2025-12-09
Test lenses

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/764

## Explanation
For each lens, check whether there is a test model for which the lens works. I originally did this in a 'static analysis' way, where I checked the lenses without initializing models. To make this work I had to add some type parameter bounds. For the current way of testing with initialized models that isn't required anymore, but I still prefer to have the type parameter bounds for clarity of the code.

## Checklist

### DIFF
--- a/Wflow/test/utils.jl
+++ b/Wflow/test/utils.jl
@@ -41,8 +41,15 @@ end
     models = Wflow.Model[]
     # Initialize the first model with mass balance
     do_mass_balance = true
-    for file_name in readdir()
-        !endswith(file_name, ".toml") && continue
+    for file_name in [
+        "sbm_config.toml",
+        "sbm_gwf_config.toml",
+        "sbm_river-floodplain-local-inertial_config.toml",
+        "sbm_river-land-local-inertial_config.toml",
+        "sbm_gwf_piave_demand_config.toml",
+        "sediment_config.toml",
+        "sediment_eurosem_engelund_config.toml",
+    ]
         config = Wflow.Config(normpath(@__DIR__, file_name))
         if do_mass_balance
             config.model.water_mass_balance__flag = true


Wflow/src/groundwater/aquifer.jl
```diff
@@ -469,37 +469,53 @@ maximum_head(aquifer::ConfinedAquifer) = aquifer.variables.head
 maximum_head(aquifer::UnconfinedAquifer) =
     min.(aquifer.variables.head, aquifer.parameters.top)
 
-@kwdef struct GroundwaterFlow{A} <: AbstractSubsurfaceFlowModel
+@kwdef struct AquiferBoundaries{
```
vers-w, line 472: OK, I did make similar changes in PR https://github.com/Deltares/Wflow.jl/pull/730 without knowing about these changes. My suggestion in general would be to try to stick to the issue as much as possible (including of course fixing bugs that were found while working on the PR).

Another comment that also applies to other suggested struct code changes: do we really need to specify these types? I don't think it impacts the performance when these types are left out, and also may restrict flexibility in the end.

Wflow/src/groundwater/aquifer.jl
```diff
@@ -143,27 +143,27 @@ function UnconfinedAquiferParameters(
         dataset,
         config,
         "subsurface_surface_water__horizontal_saturated_hydraulic_conductivity";
-        optional = false,
-        sel = indices,
-        type = Float64,
+        optional=false,
```
vers-w, line 146: Why did the formatting change here?

Wflow/src/groundwater/aquifer.jl
```diff
@@ -469,37 +469,53 @@ maximum_head(aquifer::ConfinedAquifer) = aquifer.variables.head
...
+    (boundaries.recharge, boundaries.river, boundaries.drain, boundaries.well)
+
+@kwdef struct GroundwaterFlow{A<:Aquifer,B<:AquiferBoundaries} <:
+              AbstractSubsurfaceFlowModel
     timestepping::TimeStepping
     aquifer::A
     connectivity::Connectivity
     constanthead::ConstantHead
-    boundaries::NamedTuple
+    boundaries::B = AquiferBoundaries()
```
vers-w, line 493: Why not declare `AquiferBoundaries` here and leave out `B` here?

Wflow/test/utils.jl
```diff
@@ -36,3 +36,42 @@ end
     @test Wflow.bounded_divide(1.0, 0.5; max = 0.75) == 0.75
     @test Wflow.bounded_divide(1.0, 2.0) == 0.5
 end
+
+@testitem "Lenses" begin
+    models = Wflow.Model[]
+    # Initialize the first model with mass balance
+    do_mass_balance = true
+    for file_name in readdir()
```
vers-w: My recommendation is to restrict this test to a subset of model/TOML files:
- sbm_config.toml
- sbm_river-floodplain-local-inertial_config.toml
- sbm_river-land-local-inertial_config.toml
- sbm_gwf_piave_demand_config.toml
- sediment_config.toml
- sediment_eurosem_engelund_config.toml

This will cover the different routing types and water demand of model type `sbm`.

Wflow/src/groundwater/aquifer.jl
SouthEndMusic: > try to stick to the issue as much as possible

Yes, I learned my lesson with https://github.com/Deltares/Wflow.jl/pull/735 that it is a bad idea to do a lot of things in one PR :) I explain in the PR description where the type changes come from within the context of the linked issue.

> do we really need to specify these types?

I am in favor; it makes the structure of the code more clear, both to humans and LLMs. A potential future benefit could also be static compilability. If more flexibility turns out to be required the type parameter restrictions can be loosened.

Wflow/src/groundwater/aquifer.jl
SouthEndMusic: Good question, I would like to get to the bottom of this at some point: https://github.com/Deltares/Wflow.jl/issues/765

Wflow/src/groundwater/aquifer.jl
SouthEndMusic: The `AquiferBoundaries` struct itself has multiple type parameters, so instead of propagating all of those type parameters individually, it's more economical to use a single type parameter for the `boundaries` field.

Wflow/test/utils.jl
SouthEndMusic: Thanks for this, I wondered whether there was a subset of the test models that would suffice here. `"land_drain_water__to_subsurface_volume_flow_rate"` was still missing, I added `sbm_gwf_config.toml`.

Wflow/src/groundwater/aquifer.jl
vers-w: OK, let's stick to this then. I quickly did check Ribasim core and it is also seems to be using type parameter bounds quite a lot, so good to be on the same page here. In the end I would say it is a matter of taste, @visr and I did not make use of this a lot when working together on the code.

Wflow/src/groundwater/aquifer.jl
vers-w: I don't think it is required to propagate all these parameter individually, e.g.:
```
model.routing.subsurface_flow.boundaries isa Wflow.AquiferBoundaries
```
Or am I missing something here?

SouthEndMusic: `boundaries::AquiferBoundaries` would not be type stable, then you would have to do `boundaries::AquiferBoundaries{<Type parameters>}` and propagate those type parameters

vers-w (approved):

Issue #745, opened 2025-11-06
Performance: avoid unnecessary heap allocations


SouthEndMusic:
I have found, for instance looking at the sediment code, that a lot of computations are done by allocating arrays in `update!` instead of using pre-allocated caches. I haven't tested the performance myself, but it's good practice to avoid heap allocations in simulation code. This can be investigated rigorously using `@profview allocs` (see the documentation [here](https://www.julia-vscode.org/docs/stable/userguide/profiler/).

PR #786, opened 2025-12-10
Sediment code cleanup

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/745, split out from https://github.com/Deltares/Wflow.jl/pull/735.

## Explanation
Some of the changes include:
- Defining global constants for code clarity
- Avoiding allocations
- Simplifying logic
- Using optimized functions, e.g. `cbrt(x)` instead of `x^0.33` (which leads to some small changes in the results)
- Renaming `amount` fields to more explanatory names
- Missing `@test` statements
- Using the same gravitational acceleration for sediment and local inertial flow (which leads to some small changes in the results)

## Checklist

### DIFF
--- a/Wflow/src/sediment/sediment_transport/river_transport.jl
+++ b/Wflow/src/sediment/sediment_transport/river_transport.jl
@@ -4,13 +4,13 @@ abstract type AbstractSedimentRiverTransportModel end
 @with_kw struct SedimentRiverTransportVariables
     n::Int
     # Sediment flux [t dt-1]
-    sediment_flux::Vector{Float64} = fill(MISSING_VALUE, n)
-    clay_flux::Vector{Float64} = zeros(n)
-    silt_flux::Vector{Float64} = zeros(n)
-    sand_flux::Vector{Float64} = zeros(n)
-    sagg_flux::Vector{Float64} = zeros(n)
-    lagg_flux::Vector{Float64} = zeros(n)
-    gravel_flux::Vector{Float64} = zeros(n)
+    sediment_rate::Vector{Float64} = fill(MISSING_VALUE, n)
+    clay_rate::Vector{Float64} = zeros(n)
+    silt_rate::Vector{Float64} = zeros(n)
+    sand_rate::Vector{Float64} = zeros(n)
+    sagg_rate::Vector{Float64} = zeros(n)
+    lagg_rate::Vector{Float64} = zeros(n)
+    gravel_rate::Vector{Float64} = zeros(n)
     # Total Sediment deposition rate [t dt-1]
     deposition::Vector{Float64} = fill(MISSING_VALUE, n)
     # Total sediment erosion rate (from store + direct river bed/bank) [t dt-1]
@@ -318,13 +318,13 @@ function update!(model::SedimentRiverTransportModel, domain::DomainRiver, dt::Fl
         reservoir_trapping_efficiency,
     ) = model.parameters
     (;
-        sediment_flux,
-        clay_flux,
-        silt_flux,
-        sand_flux,
-        sagg_flux,
-        lagg_flux,
-        gravel_flux,
+        sediment_rate,
+        clay_rate,
+        silt_rate,
+        sand_rate,
+        sagg_rate,
+        lagg_rate,
+        gravel_rate,
         deposition,
         erosion,
         leftover_clay,
@@ -358,13 +358,13 @@ function update!(model::SedimentRiverTransportModel, domain::DomainRiver, dt::Fl
         upstream_nodes = inneighbors(graph, v)
         if !isempty(upstream_nodes)
             for i in upstream_nodes
-                if clay_flux[i] >= 0.0 # avoid NaN from upstream non-river cells
-                    input_clay += clay_flux[i]
-                    input_silt += silt_flux[i]
-                    input_sand += sand_flux[i]
-                    input_sagg += sagg_flux[i]
-                    input_lagg += lagg_flux[i]
-                    input_gravel += gravel_flux[i]
+                if clay_rate[i] >= 0.0 # avoid NaN from upstream non-river cells
+                    input_clay += clay_rate[i]
+                    input_silt += silt_rate[i]
+                    input_sand += sand_rate[i]
+                    input_sagg += sagg_rate[i]
+                    input_lagg += lagg_rate[i]
+                    input_gravel += gravel_rate[i]
                 end
             end
         end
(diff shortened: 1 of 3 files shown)


Wflow/src/Wflow.jl
```diff
@@ -73,6 +73,8 @@ const CFVariable_MF = Union{NCDatasets.CFVariable, NCDatasets.MFCFVariable}
 const VERSION =
     VersionNumber(TOML.parsefile(joinpath(@__DIR__, "..", "Project.toml"))["version"])
 
+g_gravity = 9.81 # m s⁻²
```
visr: It will produce better code if this is marked as const. Not aware of plans to run Wflow on other planets, otherwise `g_gravity::Float64 = 9.81` will also get most benefits.

SouthEndMusic: Gravity_anomalies_on_Earth

visr: Ah, I see you are turning into a proper earth scientist.

Wflow/src/Wflow.jl
```diff
@@ -73,6 +73,8 @@ const CFVariable_MF = Union{NCDatasets.CFVariable, NCDatasets.MFCFVariable}
 const VERSION =
     VersionNumber(TOML.parsefile(joinpath(@__DIR__, "..", "Project.toml"))["version"])
 
+const g_gravity = 9.80665 # m s⁻²
```
vers-w:
```suggestion
const GRAVITY = 9.80665 # m s⁻²
```

Wflow/src/sediment/erosion/erosion_process.jl
```diff
@@ -229,19 +236,26 @@ Repartition of the effective shear stress between the bank and the bed from Knig
...
+    width::Float64,
+    length::Float64,
+    slope::Float64,
+    dt::Float64,
+)
     if waterlevel > 0.0
         # Bed and Bank from Shields diagram, Da Silva & Yalin (2017)
-        E_ = (2.65 - 1) * 9.81
-        E = (E_ * (d50 * 1e-3)^3 / 1e-12)^0.33
+        E_ = (2.65 - 1) * g_gravity
```
vers-w: Is it not better to use here `1.0` instead of `1` (also other parts of the code)?

Wflow/src/sediment/erosion/soil_erosion.jl
```diff
@@ -4,7 +4,7 @@ abstract type AbstractSoilErosionModel end
 @with_kw struct SoilErosionModelVariables
     n::Int
     # Total soil erosion rate [t dt-1]
-    amount::Vector{Float64} = fill(MISSING_VALUE, n)
+    soil_erosion_rate::Vector{Float64} = fill(MISSING_VALUE, n)
     # Total clay erosion rate [t dt-1]
     clay::Vector{Float64} = fill(MISSING_VALUE, n)
```
vers-w: Suggest to also rename the other variables, e.g. `clay_erosion_rate` to be consistent.

Wflow/src/sediment/sediment_transport/deposition.jl
```diff
@@ -1,3 +1,7 @@
+# From Eq. 7:2.2.24 in https://swat.tamu.edu/media/99192/swat2009-theory.pdf
+# [m s⁻¹ mm⁻²]
+const STOKES_FACTOR = 411 / 3600
```
vers-w, line 3: Fine to define as a `const` here, but if this parameter can be part a `parameters` struct I think that's preferred.

Wflow/src/sediment/sediment_transport/land_to_river.jl
```diff
@@ -40,17 +40,17 @@ end
...
 end
 
 "Struct to store differentiated sediment reaching the river model variables"
 @with_kw struct SedimentToRiverDifferentiationVariables
     n::Int
     # Total sediment rate [t dt-1]
-    amount::Vector{Float64} = fill(MISSING_VALUE, n)
+    sediment_rate::Vector{Float64} = fill(MISSING_VALUE, n)
     # Clay rate [t dt-1]
     clay::Vector{Float64} = fill(MISSING_VALUE, n)
```
vers-w: Suggest to also rename the other variables for consistency.

Wflow/src/sediment/sediment_transport/river_transport.jl
```diff
@@ -4,7 +4,7 @@ abstract type AbstractSedimentRiverTransportModel end
 @with_kw struct SedimentRiverTransportVariables
     n::Int
     # Sediment flux [t dt-1]
-    amount::Vector{Float64} = fill(MISSING_VALUE, n)
+    sediment_flux::Vector{Float64} = fill(MISSING_VALUE, n)
```
vers-w:
```suggestion
    sediment_rate::Vector{Float64} = fill(MISSING_VALUE, n)
```

vers-w: Suggest to also rename other variables for consistency.

Wflow/test/run_sediment.jl
```diff
@@ -157,9 +162,10 @@ end
 
         @test river.transport_capacity.parameters.d50[1] == 0.05000000074505806
         @test mean(river.transport_capacity.boundary_conditions.q) ≈ 0.6975180562953642
-        @test mean(river.transport_capacity.variables.amount) ≈ 0.14184859055736687
+        @test mean(river.transport_capacity.variables.sediment_transport_capacity) ≈
+              0.1418728167951134
 
-        @test mean(river.concentrations.variables.suspended) ≈ 0.2478754835662198
+        @test mean(river.concentrations.variables.suspended) ≈ 0.24791810261189964
```
vers-w, line 168: Any idea what is causing the difference? The test would have failed.

Wflow/test/run_sediment.jl
```diff
@@ -184,10 +190,11 @@ end
         eros = model.land
         land = model.routing.overland_flow
 
-        @test mean(eros.soil_erosion.variables.amount) ≈ 0.0776983847440198
+        @test mean(eros.soil_erosion.variables.soil_erosion_rate) ≈ 0.0776983847440198
         @test mean(land.transport_capacity.parameters.c_govers) ≈ 0.16393911236592437
-        @test mean(land.transport_capacity.variables.amount) ≈ 1.0988158364353527f6
-        @test mean(land.to_river.variables.amount) ≈ 0.07708434959918917
+        @test mean(land.transport_capacity.variables.sediment_transport_capacity) ≈
+              1.0988158364353527f6
```
vers-w: Suggest to remove float32.

Wflow/test/run_sediment.jl
```diff
@@ -240,7 +248,8 @@ end
 
         @test river.transport_capacity.parameters.d50[1] == 0.05000000074505806
         @test mean(river.transport_capacity.boundary_conditions.q) ≈ 0.6975180562953642
-        @test mean(river.transport_capacity.variables.amount) ≈ 39.959093179632234
+        @test mean(river.transport_capacity.variables.sediment_transport_capacity) ≈
+              39.955369987101946
```
vers-w, line 252: Good to check the difference here, test would have failed.

Wflow/test/run_sediment.jl
```diff
@@ -292,9 +302,10 @@ end
 
         @test river.transport_capacity.parameters.d50[1] == 0.05000000074505806
         @test mean(river.transport_capacity.boundary_conditions.q) ≈ 0.6975180562953642
-        @test mean(river.transport_capacity.variables.amount) ≈ 350.6483600591209
+        @test mean(river.transport_capacity.variables.sediment_transport_capacity) ≈
+              350.7538564169241
 
-        @test mean(river.concentrations.variables.suspended) ≈ 884.4794354416674
+        @test mean(river.concentrations.variables.suspended) ≈ 884.749798262824
```
vers-w, line 308: Good to check difference here, test would have failed.

vers-w (commented):
LGTM. 

Please check carefully why the results of some tests have changed, in some cases tests would have failed.

Wflow/src/Wflow.jl
SouthEndMusic: Fair point, I at some point advised to denote global constants with capital letters. However, I would like to retain the `g` as it is the recognized symbol  for gravitational acceleration, so would `g_GRAVITY` be a good compromise?

SouthEndMusic: Or maybe `GRAVITATIONAL_ACCELERATION`

Wflow/src/sediment/erosion/erosion_process.jl
SouthEndMusic: I think generally it is indeed better to avoid type conversions if they aren't necessary, but I think `(2.65 - 1)` gets compiled away here so it doesn't matter much. @visr thoughts?

Wflow/src/sediment/sediment_transport/deposition.jl
SouthEndMusic: Can you motivate why, maybe to allow for customizability? The advantage of global constants is that you don't have to pass them around to the functions where they are required.

Wflow/test/run_sediment.jl
SouthEndMusic: From the PR description:

> Using optimized functions, e.g. cbrt(x) instead of x^0.33 (which leads to some small changes in the results)

Wflow/src/sediment/erosion/erosion_process.jl
evetion: For variables, [certainly](https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-changing-the-type-of-a-variable). Here it gets compiled away though (you can check yourself with `@code_llvm` or `@code_native`).

Wflow/src/Wflow.jl
vers-w: Although a bit long, I think `GRAVITATIONAL_ACCELERATION` is preferred.

Wflow/src/sediment/sediment_transport/river_transport.jl
SouthEndMusic: Are you sure about rate instead of flux here? I followed the comments. I suppose it's flux here to differentiate between erosion and transport.

vers-w: The standard name of this variable has quantity `mass_flow_rate` and unit `t dt-1` and sediment standard names were reviewed by @hboisgon , so I think rate is correct. There was indeed some confusion about sediment units because of differences between code, metadata and docs when working on the standard names.

PR #787, opened 2025-12-11
Post release

vers-w:


JoostBuitink (commented):
Don't know what the common approach here is, `1.1.0-dev` or `1.0.1-dev`? I actually see that hydromt uses `1.2.0.dev0` where `1.2.0` is their latest release

vers-w:
> Don't know what the common approach here is, `1.1.0-dev` or `1.0.1-dev`? I actually see that hydromt uses `1.2.0.dev0` where `1.2.0` is their latest release

I don't think the approach of hydromt is common. It seems to me the general approach is to bump the version, and `dev` indicates it is in development. The approach `1.1.0-dev` is also used by Julia itself, and I think makes sense (we're working in the Master branch to the next `MINOR` version).

PR #788, opened 2025-12-11
update dockerfile

JoostBuitink:
The dockerfile was not yet updated to reflect the new download_test_data.jl script. Also updated the Julia version to match the other versions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,14 @@
-FROM julia:1.9.2
+FROM julia:1.11.7
 LABEL maintainer="Maarten Pronk <maarten.pronk@deltares.nl>"
 
 RUN apt-get update && apt-get install -y \
-    g++ git gcc \
-    && rm -rf /var/lib/apt/lists/*
+    g++ git gcc && \
+    rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,13 +1,14 @@
...
 RUN apt-get update && apt-get install -y \
-    g++ git gcc \
-    && rm -rf /var/lib/apt/lists/*
+    g++ git gcc && \
+    rm -rf /var/lib/apt/lists/*
 ADD . /app
 WORKDIR /app/build/create_binaries/
-RUN julia --project -e "using Pkg; Pkg.instantiate();" && \
-    julia --project download_test_data.jl && \
+RUN julia --project=/app/Wflow -e "using Pkg; Pkg.instantiate();" && \
```
vers-w, line 9: Is this line of code required? It seems to me this is handled by the Manifest.toml file in \create_binaries\?

JoostBuitink: Yes, this line is required to install all the packages inside the docker environment. This indeed uses the Manifest, but the packages still need to be downloaded and installed.

vers-w (approved):

PR #789, opened 2025-12-11
Small improvements

SouthEndMusic:
## Explanation
Split out from https://github.com/Deltares/Wflow.jl/pull/735.
Changes:

+ Add enumerators for reservoir options: `ReservoirProfileType`, `ReservoirOutflowType`
+ More readible and efficient rewrites of `snowpack_hbv`, `glacier_hbv`, `feddes_h3`, `rwu_reduction_feddes`
+ Remove unnecessary `@reset` calls in `ReservoirParameters` constructor
+ `kinematic_wave`, `ssf_celerity`, `local_inertial_flow`, `sum_at` efficiency refactor
+ Add `to_table` function for pretty printing info messages using `PrettyTables.jl`

## Checklist

### DIFF
--- a/Wflow/src/routing/routing_process.jl
+++ b/Wflow/src/routing/routing_process.jl
@@ -23,7 +23,7 @@ function flowgraph(ldd::AbstractVector, indices::AbstractVector, PCR_DIR::Abstra
     return graph
 end
 
-const KIN_WAVE_MIN_VOLUME = 1e-30 # [m]
+const KIN_WAVE_MIN_FLOW = 1e-30 # [m³ s⁻¹]
 
 "Kinematic wave surface flow rate for a single cell and timestep"
 function kinematic_wave(q_in, q_prev, q_lat, alpha, beta, dt, dx)
@@ -39,7 +39,7 @@ function kinematic_wave(q_in, q_prev, q_lat, alpha, beta, dt, dx)
         if isnan(q)
             q = 0.0
         end
-        q = max(q, KIN_WAVE_MIN_VOLUME)
+        q = max(q, KIN_WAVE_MIN_FLOW)
         # newton-raphson
         max_iters = 3000
         epsilon = 1.0e-12
@@ -52,7 +52,7 @@ function kinematic_wave(q_in, q_prev, q_lat, alpha, beta, dt, dx)
             if isnan(q)
                 q = 0.0
             end
-            q = max(q, KIN_WAVE_MIN_VOLUME)
+            q = max(q, KIN_WAVE_MIN_FLOW)
             if (abs(f_q) <= epsilon) || (count >= max_iters)
                 break
             end
@@ -105,7 +105,7 @@ function kw_ssf_newton_raphson(ssf, constant_term, celerity, dt, dx)
         if isnan(ssf)
             ssf = 0.0
         end
-        ssf = max(ssf, KIN_WAVE_MIN_VOLUME)
+        ssf = max(ssf, KIN_WAVE_MIN_FLOW)
         if (abs(f) <= epsilon) || (count >= max_iters)
             break
         end
@@ -154,7 +154,7 @@ function kinematic_wave_ssf(
         # lower boundary ssf
         zi = zi_prev - (ssfin * dt + r * dt * dx - ssf * dt) / (dw * dx) / theta_e
         if zi > d
-            ssf = max(ssf - (dw * dx) * theta_e * (zi - d), KIN_WAVE_MIN_VOLUME)
+            ssf = max(ssf - (dw * dx) * theta_e * (zi - d), KIN_WAVE_MIN_FLOW)
         end
         exfilt = min(zi, 0.0) * -theta_e
         zi = clamp(zi, 0.0, d)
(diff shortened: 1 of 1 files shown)


Wflow/src/Wflow.jl
```diff
@@ -307,7 +308,13 @@ function run!(model::Model; close_files = true)
     endtime = cftime(config.time.endtime, config.time.calendar)
     times = range(starttime + dt, endtime; step = dt)
 
-    @info "Run information" model_type = String(Symbol(model_type)) starttime dt endtime nthreads()
+    @info "Run information." * to_table(;
```
vers-w: I think the logging output now gets too verbose/large with these tables and would prefer the previous implementation with most logging messages covering one line. @JoostBuitink : good to also hear your opinion on this.

Wflow/src/glacier/glacier_process.jl
```diff
@@ -36,21 +36,24 @@ function glacier_hbv(
 )
 
     # Fraction of the snow transformed into ice (HBV-light model)
-    snow_to_glacier = g_sifrac * snow
-    snow_to_glacier = glacierfrac > 0.0 ? snow_to_glacier : 0.0
+    snow_to_glacier = if glacier_frac > 0.0
+        g_sifrac * snow_storage
+    else
+        0.0
+    end
```
vers-w, line 43: These kind of changes are fine with me, although in my view whether this is an improvement is mostly a matter of taste/personal coding style.

Wflow/src/routing/reservoir.jl
```diff
@@ -1,13 +1,17 @@
+# Explicit integers added because the enumerators are 0-based by default
+@enumx ReservoirProfileType linear = 1 interpolation = 2
+@enumx ReservoirOutflowType rating_curve = 1 free_weir = 2 modified_puls = 3 simple = 4
```
vers-w, line 3: Nice addition!

Wflow/src/routing/routing_process.jl
```diff
@@ -23,31 +23,36 @@ function flowgraph(ldd::AbstractVector, indices::AbstractVector, PCR_DIR::Abstra
     return graph
 end
 
+const MIN_FLOW = 1e-30 # [m s⁻¹]
```
vers-w: The unit is wrong, it should be $\mathrm{m}^3$. 

I think `MIN_FLOW` is a bit too general as it is a specific kinematic wave (solver) setting.

Wflow/src/routing/routing_process.jl
```diff
@@ -91,14 +96,16 @@ function kw_ssf_newton_raphson(ssf, constant_term, celerity, dt, dx)
     epsilon = 1.0e-12
     max_iters = 3000
     count = 0
+    C = dt / dx
```
vers-w: Preference is to have lower case, e.g. (I think this is also clearer):
```suggestion
    dt_dx = dt / dx
```

Wflow/test/reservoir.jl
```diff
@@ -60,6 +61,7 @@
 end
 
 @testitem "Update reservoir Modified Puls approach (outflowfunc = 3)" begin
+    using Wflow: to_SI, Unit, ReservoirProfileType, ReservoirOutflowType
```
vers-w: I think `to_SI` and `Unit` can be removed?

Wflow/src/Wflow.jl
SouthEndMusic: That's fair, I'll revert this change

Wflow/src/glacier/glacier_process.jl
SouthEndMusic: I agree with you, I had an interesting discussion with @evetion about refactors like this. Maarten can you recall what you think the motivation was for the current implementation style?

I am motivated by minimizing the number of operations and code clarity, which admittedly aren't significantly improved here, so what remains is consistency. I think the snow refactor is a better example; splitting the two temperature cases in separate code blocks.

Wflow/src/routing/routing_process.jl
SouthEndMusic: Thanks!

Wflow/test/reservoir.jl
SouthEndMusic: Yep, I never understood why this doesn't throw an error

Wflow/src/glacier/glacier_process.jl
evetion: I think our argument was mostly about being able to test different cases, and you suggested that code coverage would pick up these changes better. In this case, the else case should have its own test, whereas that might've been unclear from the previous code.

While we're here, I find it confusing to have this check, because we use it later anyway, and if it's zero we could skip most of this method anyway (all becomes zero, even melt from my quick read here).

Wflow/src/glacier/glacier_process.jl
```diff
@@ -36,21 +36,24 @@ function glacier_hbv(
...
 
     # Restrict snow_to_glacier conversion
     snow_to_glacier = min(snow_to_glacier, max_snow_to_glacier)
 
-    snow -= snow_to_glacier * glacierfrac
-    glacierstore += snow_to_glacier
+    snow_storage -= snow_to_glacier * glacier_frac
+    glacier_store += snow_to_glacier
```
evetion, line 49: Should this also be multiplied by the glacier_frac?

vers-w: No, `glacier_store` has units [mm] and is defined for a fraction of a cell. `snow_storage` [mm] is defined for the complete cell.

Wflow/src/routing/routing_process.jl
```diff
@@ -23,36 +23,36 @@ function flowgraph(ldd::AbstractVector, indices::AbstractVector, PCR_DIR::Abstra
     return graph
 end
 
-const MIN_FLOW = 1e-30 # [m s⁻¹]
+const KIN_WAVE_MIN_VOLUME = 1e-30 # [m]
```
vers-w: Sorry, I did provide the wrong unit, it should be m3/s. And then `KIN_WAVE_MIN_FLOW` is a better name.

Issue #758, opened 2025-11-19
Add (unit) tests simulating isolated processes


SouthEndMusic:
It would be very nice if there was an unit test per process. My idea of this is to write a simple simulation loop within the unit test that only calls the tested process, and updates all necessary boundary data. This small simulation would then test all relevant cases of the process in an explainable way.

vers-w:
Note that we do already have unit tests in:
- land_process.jl
- routing_process.jl
- reservoir.jl
- groundwater.jl

For the integrated tests we also test vertical processes or specific parts (for example water balance) so not only a downstream flow.

Anyway, we can of course discuss if more tests would be helpful.

PR #793, opened 2025-12-15
Add unit tests

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/758

## Explanation

Unit tests of the form "some numbers in, some numbers out" are not very informative. I created/refactored unit tests where the variable names of all inputs are used, and where relevant I specify different cases which yield different code paths within the tested function.

Notes: 
- Refactor of `rainfall_interception_gash` yields changes of no more than 1e-14 in the outcome, but yields larger differences downstream
- There seems to be no integration test with the modified Rutter interception model

## Checklist

### DIFF
--- a/Wflow/src/vegetation/parameters.jl
+++ b/Wflow/src/vegetation/parameters.jl
@@ -3,19 +3,19 @@
     # Leaf area index [m² m⁻²]
     leaf_area_index::Union{Vector{Float64}, Nothing}
     # Storage woody part of vegetation [mm]
-    storage_wood::Union{Vector{Float64}, Nothing} = nothing
+    storage_wood::Union{Vector{Float64}, Nothing}
     # Extinction coefficient [-] (to calculate canopy gap fraction)
-    kext::Union{Vector{Float64}, Nothing} = nothing
+    kext::Union{Vector{Float64}, Nothing}
     # Specific leaf storage [mm]
-    storage_specific_leaf::Union{Vector{Float64}, Nothing} = nothing
+    storage_specific_leaf::Union{Vector{Float64}, Nothing}
     # Canopy gap fraction [-]
-    canopygapfraction::Vector{Float64} = []
+    canopygapfraction::Vector{Float64}
     # Maximum canopy storage [mm]
-    cmax::Vector{Float64} = []
+    cmax::Vector{Float64}
     # Rooting depth [mm]
-    rootingdepth::Vector{Float64} = []
+    rootingdepth::Vector{Float64}
     # Crop coefficient Kc [-]
-    kc::Vector{Float64} = []
+    kc::Vector{Float64}
 end
 
 "Initialize (shared) vegetation parameters"
--- a/Wflow/test/utils.jl
+++ b/Wflow/test/utils.jl
@@ -67,7 +67,9 @@ end
 end
 
 @testitem "Lenses" begin
-    models = Wflow.Model[]
+    configs_sbm = Wflow.Config[]
+    configs_sediment = Wflow.Config[]
+
     # Initialize the first model with mass balance
     do_mass_balance = true
     for file_name in [
@@ -76,8 +78,6 @@ end
         "sbm_river-floodplain-local-inertial_config.toml",
         "sbm_river-land-local-inertial_config.toml",
         "sbm_gwf_piave_demand_config.toml",
-        "sediment_config.toml",
-        "sediment_eurosem_engelund_config.toml",
     ]
         config = Wflow.Config(normpath(@__DIR__, file_name))
         config.dir_output = mktempdir()
@@ -85,19 +85,25 @@ end
             config.model.water_mass_balance__flag = true
             global do_mass_balance = false
         end
-        push!(models, Wflow.Model(config))
+        push!(configs_sbm, config)
     end
 
-    for (map_name, standard_name_map) in (
-        ("sbm", Wflow.sbm_standard_name_map),
-        ("sediment", Wflow.sediment_standard_name_map),
+    for file_name in ["sediment_config.toml", "sediment_eurosem_engelund_config.toml"]
+        config = Wflow.Config(normpath(@__DIR__, file_name))
+        config.dir_output = mktempdir()
+        push!(configs_sediment, config)
+    end
+
+    models_sbm = Wflow.Model.(configs_sbm)
+    models_sediment = Wflow.Model.(configs_sediment)
+
+    for (map_name, standard_name_map, models) in (
+        ("sbm", Wflow.sbm_standard_name_map, models_sbm),
+        ("sediment", Wflow.sediment_standard_name_map, models_sediment),
     )
         @testset "Test lenses: $map_name" begin
-            dict_data = collect(standard_name_map)
-            n = length(dict_data)
-            invalids = zeros(Bool, n)
-            Wflow.threaded_foreach(1:n; basesize = 25) do i
-                (name, data) = dict_data[i]
+            invalids = String[]
+            for (name, data) in standard_name_map
                 (; lens) = data
                 invalid = true
                 for model in models
@@ -109,10 +115,9 @@ end
                         nothing
                     end
                 end
-                invalids[i] = invalid
+                invalid && push!(invalids, name)
             end
-            invalid = [dict_data[i][1] for i in findall(invalids)]
-            @test isempty(invalid)
+            @test isempty(invalids)
         end
     end
 end
--- a/Wflow/test/water_demand.jl
+++ b/Wflow/test/water_demand.jl
@@ -36,6 +36,12 @@
         vegetation_parameter_set = Wflow.VegetationParameters(;
             rootingdepth = [150.0],
             leaf_area_index = nothing,
+            storage_wood = nothing,
+            kext = nothing,
+            storage_specific_leaf = nothing,
+            canopygapfraction = [0.5],
+            cmax = [0.0],
+            kc = [1.0],
         ),
     )
 


Wflow/src/demand/water_demand.jl
```diff
@@ -186,55 +186,91 @@ function update_demand_gross!(model::NonPaddy, soil::SbmSoilModel)
...
         else
-            irri_dem_gross = 0.0
+            0.0
         end
-        model.variables.demand_gross[i] = irri_dem_gross
     end
     return nothing
 end
+
+function update_demand_gross(
```
vers-w: Ideally, we would limit the number of function arguments, so I think the previous implementation is preferred. My recommendation is to apply the unit test on the previous implementation, also for the other changes in water_demand.jl. As an option, creating the required objects for these functions could be simplified by removing some of the type restrictions. 

@JoostBuitink : good to also hear your opinion on this.

JoostBuitink: I agree that this list is very long, which also increases the risk that an argument is placed at two wrong location (which is now easier to traceback due to the unit test, so that is good). This does however raise the question on which number of arguments we deem too many: for example, the snow functions only have a couple of arguments, which is perfectly fine.

vers-w: Yes, my main point here (water demand) is: let's not increase the number of function arguments when it is not really necessary. Additionally, for this case the arguments belong to two different objects which makes it also harder to read.

SouthEndMusic: I see your point about having simple function signatures. On the other hand, making the code more unit-testable is also valuable (I think @deltamarnix will agree with me there). The new unit tests that these refactors allowed me to make can for instance be really helpful in debugging https://github.com/Deltares/Wflow.jl/pull/735. These functions that I split out are easier to understand and debug by someone who isn't intimately familiar with the code base because of the simple input and output types, and do not require any specialized initialization machinery to generate inputs for (although I see your point in loosening the type restrictions, that's I think a matter of taste). It also looks like for some of these loops that I refactored, different iterations can affect the same value, which can also make debugging a test harder.

deltamarnix: I agree that the number of arguments for this function is too high. I also agree that splitting out functions that can be tested separately is nice.

Instead, try to see if you can split the function even further, so that you can call those pieces one after another.
For example, keep the for loop in the original function, and create some smaller functions inside of that for loop.

```jl
# original function (not update_demand_gross)
for k in 1:n_unsatlayers
    raw = water_demand_root_zone(some, args, here)
    irri_dem_gross = check_depletion(some, more, args, here)
end
irri_dem_griss = limit_irrigation_demand(more, args)
# ... you get the drill
```

Or consider other ways to pass the arguments. For example, by creating an options-struct that contains all the information that you want to pass. The struct can be filled with named arguments, reducing the chance that the order of the arguments is wrong.

```jl
struct DemandGrossOptions
    n_unsatlayers
    ...
end

function update_demand_gross(demand_gross_options)
```

It is also possible to make keyword arguments mandatory with `;`. https://docs.julialang.org/en/v1/manual/functions/#Keyword-Arguments

```jl
function update_demand_gross(
    ; # see this semi-colon
    n_unsatlayers,
    rootingdepth,
    sumlayers,
    ustorelayerthickness,
    ustorelayerdepth,
    hb,
    theta_s,
    theta_r,
    c,
    h3,
    maximum_irrigation_rate,
    f_infiltration_reduction,
    pathfrac,
    infiltcapsoil,
    irrigation_efficiency,
    demand_gross,
)
```

vers-w: I agree that adding more unit tests is useful.

However, I don't agree that these refactors (including @deltamarnix suggestions using an options-struct or keyword arguments) make the code easier to understand. Actually, I think it generally makes the code harder to read. Additionally, I would say initializing the objects as part of the unit test is a better way to understand the function instead of splitting it out into a function with (many) arguments from two different objects (also for someone that is not that familiar with the code, although in my opinion this is not a very important aspect of a unit test). Also, note, that we already make use of initializing objects as part of unit tests for other parts of the code. I have also done this for PR https://github.com/Deltares/Wflow.jl/pull/730 and did remove the type restriction for `SbmSoilModel` and think it is very reasonable to do this also for the water demand functions.

As per @deltamarnix suggestion, splitting the function even further may help, but even then I think you end up (at least partly) with passing objects to avoid many function arguments. Maybe something to look into further as part of a separate issue/PR?

SouthEndMusic: I discussed this with @deltamarnix and came to the following conclusion:

Being able to do finer unit tests is good, but it shouldn't be achieved at a high cost for the kernel code quality. Having only a few function arguments is good for readability and reducing errors as mentioned, so I accept having to initialize objects for unit tests. @vers-w I don't see passing objects down a nested function chain as a bad thing as long as the nesting is not too deep. I realize now that unpacking structs only at the level where it is needed (as low as possible) makes quite a lot of sense.

Wflow/src/demand/water_demand.jl
```diff
@@ -169,72 +169,80 @@ by the infiltration capacity, taking into account limited irrigation efficiency
...
+        (ustorelayerdepth[i][k] + theta_r[i] * ustorelayerthickness[i][k])
+    depletion *= rootfrac
+    readily_available_water = (vwc_fc - vwc_h3) * ustorelayerthickness[i][k]
+    readily_available_water *= rootfrac
+
+    return depletion, readily_available_water
+end
+
+"Limit irrigation demand to infiltration capacity"
+function apply_infiltration_capacity(
```
vers-w: Suggest to rename function to `compute_demand_gross` and also change doc string accordingly. Indeed, infiltration capacity is applied to limit irrigation demand, but also `irrigation_efficiency` is used to compute `irri_dem_gross`. Maybe the function could be split in two, not sure if that's worth it.

Wflow/src/demand/water_demand.jl
```diff
@@ -397,36 +405,44 @@ is `true` (`on`) and when the paddy water depth `h` reaches below the minimum wa
...
         else
-            irri_dem_gross = 0.0
+            demand_gross[i] = 0.0
         end
-        demand_gross[i] = irri_dem_gross
     end
-    return nothing
+end
+
+function calc_demand_gross(model::Paddy, i::Int)
```
vers-w: Suggest to rename function to `compute_irrigation_depth`, gross water demand is computed after this function based on `irrigation_efficiency` and `maximum_irrigation_rate`.

Wflow/test/testing_utils.jl
```diff
@@ -111,3 +111,70 @@ function homogenous_aquifer(nrow, ncol)
     unconf_aqf = Wflow.UnconfinedAquifer(; parameters, variables)
     return (connectivity, conf_aqf, unconf_aqf)
 end
+
+function simple_soil(n, N; kwargs...)
```
vers-w: Suggest to rename this function, I think `simple_soil` is a bit too general, maybe something like `init_sbm_soil_model`?

Wflow/test/water_demand.jl
```diff
@@ -0,0 +1,245 @@
...
+@testitem "unit: update_demand_gross! (Paddy)" begin
+    variables = Wflow.PaddyVariables(; n = 1, h = [15.0])
+    parameters = Wflow.PaddyParameters(;
+        irrigation_efficiency = [1.0],
+        maximum_irrigation_rate = [25.0],
+        irrigation_areas = [true],
+        irrigation_trigger = [true],
+        h_min = [20.0],
+        h_opt = [50.0],
+        h_max = [NaN],
```
vers-w: Why a `NaN` value here? Suggest to use the default value 80.0.

Wflow/test/water_demand.jl
```diff
@@ -0,0 +1,245 @@
...
+end
+
+@testitem "unit: surface_water_allocation_local!" begin
+    include("testing_utils.jl")
+    n = 1
+
+    model = Wflow.AllocationLand(;
+        n,
+        parameters = Wflow.AllocationLandParameters(;
+            frac_sw_used = [NaN],
```
vers-w: Why a `NaN` value here? Suggest to use the default value 1.0.

Wflow/test/testing_utils.jl
```diff
@@ -111,3 +111,70 @@ function homogenous_aquifer(nrow, ncol)
...
+    variables = Wflow.SbmSoilVariables(; kwargs_variables...)
+
+    kwargs_parameters =
+        filter(pair -> pair.first ∈ fieldnames(Wflow.SbmSoilParameters), kwargs)
+    parameters = Wflow.SbmSoilParameters(; kwargs_parameters...)
+
+    return Wflow.SbmSoilModel(; n, variables, parameters)
+end
+
+@kwdef struct DummyRiver{A, B, V} <: Wflow.AbstractRiverFlowModel
```
vers-w, line 180: I guess this struct is only used for unit testing surface water allocation? Suggest to add a comment for clarity and maybe also change the struct name (e.g. `RiverWaterAllocation`?).

Wflow/test/water_demand.jl
```diff
@@ -0,0 +1,245 @@
...
+            irrigation_areas = [true],
+            irrigation_trigger = [true],
+        ),
+        variables = Wflow.NonPaddyVariables(; n, demand_gross = [0.8604076280853505]),
+    )
+
+    soil = simple_soil(
+        n,
+        N;
+        # Variables
```
vers-w, line 20: The variables/parameters (quantities) are not consistent. Did you generate these values from a wflow test model? For example with `n_unsatlayers  = 3` you would also expect three values (not `NaN`) for `ustorelayerthickness`. The number of soil layers `nlayers` is 7, but for `ustorelayerthickness` and `ustorelayerdepth` there are 6 values.

Suggest to make the quantities consistent. Also for other unit tests if this is not the case.

Wflow/test/water_demand.jl
```diff
@@ -0,0 +1,245 @@
...
+    @test model.variables.surfacewater_alloc |> only ≈ 0.02
+end
+
+@testitem "unit: surface_water_allocation_area!" begin
+    include("testing_utils.jl")
+
+    n = 3
+    model = Wflow.AllocationLand(;
+        n,
+        parameters = Wflow.AllocationLandParameters(; frac_sw_used = [], areas = []),
```
vers-w: Why empty here? Suggest to use default values.

Wflow/test/water_demand.jl
SouthEndMusic: That is a fair point, I adapted this from a Wflow test model but I didn't know how to create consistent input. Can you fix it?

vers-w: Because you already adapted this from a Wflow test model, I think it is better (see it also a learning experience :smiley:) if you fix it. You can just use the values from one cell of the Wflow test model!

Wflow/test/water_demand.jl
```diff
@@ -0,0 +1,239 @@
...
+        ustorelayerthickness = [SVector(2.334, 2.334, 2.334)],
+        ustorelayerdepth = [SVector(0.0, 0.0, 0.0)],
+        n_unsatlayers = [3],
+        h3 = [-934.9109542889025],
+        f_infiltration_reduction = [0.8],
+        # Parameters
+        maxlayers = 3,
+        sumlayers = [SVector(0.0, 50.0, 150.0, 200.0)],
+        c = [SVector(9.195682525634766, 9.297739028930664, 9.597416877746582)],
+        nlayers = [7],
```
vers-w: Did you make use here of the actual values of a Wflow test model? `nlayers`  should be `[3]`, `ustorelayerthickness` is not correct, suggest to use `[SVector(50.0, 100.0, 50.0)]` and suggest to make `rootingdepth` smaller, for example 150.0.

Wflow/src/vegetation/parameters.jl
```diff
@@ -1,21 +1,21 @@
 "Struct to store (shared) vegetation parameters"
 @with_kw struct VegetationParameters
     # Leaf area index [m² m⁻²]
-    leaf_area_index::Union{Vector{Float64}, Nothing}
+    leaf_area_index::Union{Vector{Float64}, Nothing} = nothing
```
vers-w: As discussed for the `SbmSoilModel`, better to remove the trivial defaults and add these as part of unit testing.

Wflow/test/water_demand.jl
SouthEndMusic: Thanks!

Wflow/src/vegetation/parameters.jl
```diff
@@ -3,19 +3,19 @@
...
+    canopygapfraction::Vector{Float64} = []
     # Maximum canopy storage [mm]
-    cmax::Vector{Float64}
+    cmax::Vector{Float64} = []
     # Rooting depth [mm]
-    rootingdepth::Vector{Float64}
+    rootingdepth::Vector{Float64} = []
     # Crop coefficient Kc [-]
-    kc::Vector{Float64}
+    kc::Vector{Float64} = []
```
vers-w: There are still some trivial defaults left here.

Wflow/test/utils.jl
```diff
@@ -64,21 +93,25 @@ end
         ("sediment", Wflow.sediment_standard_name_map),
     )
         @testset "Test lenses: $map_name" begin
-            invalid = String[]
-            for (name, data) in standard_name_map
+            dict_data = collect(standard_name_map)
+            n = length(dict_data)
+            invalids = zeros(Bool, n)
+            Wflow.threaded_foreach(1:n; basesize = 25) do i
```
vers-w: Do we gain a lot by running this part in parallel for all tests? If not, I am not much in favor of parallelizing this (also note that we consider the multi-threading functionality as experimental, also the reason that we run our tests on 1 and 4 threads).

vers-w (changes requested):

Wflow/test/utils.jl
SouthEndMusic: I did this earlier because I found the test quite slow. I removed the parallelization and restructured the test a bit so that lenses are tried on model types that are expected to have the variable.

PR #795, opened 2025-12-16
Introduce unit system and use standard name data as single source of truth

### LINKED ISSUE #492: Automatically generate tables in docs from data in source code
### Wflow version checks

- [x] I have checked that the issue still exists on the latest versions of the docs on `master` [here](https://github.com/Deltares/Wflow.jl)

### Kind of issue

Docs are wrong

### Location of the documentation

Multiple

### Documentation problem

(Follow-up of https://github.com/Deltares/Wflow.jl/pull/448)

Multiple tables are now hardcoded in the documentation, e.g. [here](https://deltares.github.io/Wflow.jl/quarto/model_docs/parameters_vertical.html). If the data for the table can be obtained from the `Wflow` package, then the table can be automatically generated:

```julia
using Wflow
using DataFrames: DataFrame
using MarkdownTables

df = DataFrame(parameter = collect(fieldnames(Wflow.SBM)))
df.unit = Wflow.get_units.(Wflow.SBM, df.parameter)

markdown_table(df)
```

This yields the following result:

Image.

However, not all data that we want to have in this table is easily accessible, i.e. variable descriptions and default values. That would require some refactoring in the core to achieve. Default values can be added to the struct definitions in the way that is allowed by `@with_kw`, for the metadata I am not quite sure (in principle using `@metadata` as used in the example above is possible).

### Suggested fix for documentation

_No response_

### LINKED ISSUE #653: parameter (time) units in docs
Some thoughts about the input time units in docs (btw really nice work in refactoring docs, congrats!). 
I ll take the soil_layer_water__vertical_saturated_hydraulic_conductivity as an example. 

The input unit in the model docs appears to be in mm per model timestep. This is correct, but it's kinda from an 'internal' point of view. The input of KsatVer (e.g if I want to build my own layers) is always expected in mm/day (and is produced by HydroMT in the staticmaps, which is what the user 'always' sees), and then the unit conversion is handled internally. I had similar discussions/questions with colleagues from the groundwater department, where I was handling infiltration and recharge time series from an hourly model while they were trying to inspect KsatVer and the effect of KsatHorFrac (of course they didn’t read the docs).

The default column units are indicating this and in the introduction it is mentioned "_The time unit of input flux parameters is day (the model base time step size), hence default values for these parameters are expressed per day. During model initialization these input parameters are converted to the user-defined model time step size Δt_.", but could we somehow improve the tables? 

An example could be that we include an additional column and restructuring the table to something like this

| Standard Name                                                   | Description                                                                                           | **Unit (Input Expected)** | **Unit (Used Internally)** | Default |
|--------------------------------------------------------|-------------------------------------------------------------------------------------------------------|----------------------------|-----------------------------|------------------------|
| `soil_layer_water__vertical_saturated_hydraulic_conductivity` | Vertical saturated hydraulic conductivity per soil layer, when a `layered` or `layered_exponential` saturated hydraulic conductivity profile is used | mm day⁻¹                   | mm Δt⁻¹                     | –                      |

or maybe copy the introduction section to the tables as well.

Or do we complicate things more? Curious to hear your thoughts. This “issue” is meant for discussion.

### LINKED ISSUE #749: Have a single source of truth for input and output metadata
As a follow-up of https://github.com/Deltares/Wflow.jl/pull/735 I would like to move the default values (and also the descriptions for completeness) of the variables from hardcoded in the docs to part of the `standard_name` structure. That way:
- Data only has to be added/updated in one place reducing the risk of errors creeping in
- The parameter tables in the docs can be generated automatically (I see that there's already an instance of running Julia code during rendering of the docs). We already do this for Ribasim: https://ribasim.org/reference/validation.html
- Default values are no longer distributed over the code base but in one place

I would say use something like

```julia
struct VariableMetadata{L, D}
   lens::L
   unit::Unit
   default::D
   description::String
   flag::Symbol
end
```

where `flag` can be used to filter parameters during table generation.

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/492
Fixes https://github.com/Deltares/Wflow.jl/issues/749
Fixes https://github.com/Deltares/Wflow.jl/issues/653

## Explanation
The unit system was split out from https://github.com/Deltares/Wflow.jl/pull/735

## Checklist

Changes in the order in which I think the review makes the most sense:

- The standard name data: Because there is now a lot more standard name data, I split it up in several categories, see the folder `standard_name`. The file `standard_name_utils.jl` defines the central `ParameterMetadata` struct used as the values in the standard name maps.
- All the tables in the docs are now automatically generated from the standard name maps, where a table is defined by all entries with a certain tag. For how the tables are generated see `docs_utils.jl`. I took quite some care to make the tables complete (and have a similar look), but a check of the tables is probably warranted.
- The next significant change is in the function `ncread`, where the metadata is now retrieved from the standard name maps.
- The unit system is introduced in this PR in the file `units.jl` since the units are part of the metadata, but most of it is more relevant in https://github.com/Deltares/Wflow.jl/pull/735


AquaPore (approved):

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        lens = @optic(_.land.vegetation_parameters.cmax),
+        unit = Unit(; mm = 1),
+        default = 1.0,
+        description = "Maximum canopy storage",
+        tags = [:vegetation_input],
+    ),
+    "vegetation__leaf_area_index" => ParameterMetadata(;
+        lens = @optic(_.land.vegetation_parameters.leaf_area_index),
+        unit = Unit(; m = (2, 2)),
+        default = 3.0,
```
vers-w: There is no default value for `leaf_area_index`.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+        unit = Unit(; m = 1),
+        description = "Reservoir water level",
+        tags = [
+            :reservoir_state,
+            :reservoir_output,
+            :kinematic_wave_river_static_cyclic_forcing_input,
+        ],
+    ),
+    "reservoir_water__outgoing_volume_flow_rate" => ParameterMetadata(;
+        lens = @optic(_.routing.river_flow.boundary_conditions.reservoir.variables.outflow),
```
vers-w:
```suggestion
        lens = @optic(_.routing.river_flow.boundary_conditions.reservoir.variables.outflow_av),
```

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+        type = Int,
+        description = "Type of reservoir rating curve",
+        tags = [:reservoir_input],
+    ),
+    "reservoir_water_surface__initial_elevation" => ParameterMetadata(;
+        unit = Unit(; m = 1),
+        description = "Water level of reservoir (used for initialization)",
+        tags = [:reservoir_input],
+    ),
+    #### Static or cyclic/forcing input
```
vers-w, line 78: In the previous implementation an input parameter had to be part of the standard name mapping if it was allowed as cyclic/forcing input (implicit check whether this was allowed). Now both static input parameters and allowed cyclic/forcing input parameters are part of the standard name mapping. I think it would be good to explicitly add this to the metadata, this way we could also check if this is allowed. 

Probably also need to add to the docs somewhere that input parameters can also be written as output (also it was asked recently if this is possible as part of a wflow project) , though we may not want to support writing all input parameters (for example floodplain profile data), also tagging @JoostBuitink here for his input.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+        tags = [:reservoir_generic_input],
+    ),
+    #### Input
+    "reservoir_surface__area" => ParameterMetadata(;
+        lens = @optic(_.routing.river_flow.boundary_conditions.reservoir.parameters.area),
+        unit = Unit(; m = 2),
+        description = "Area of the reservoir",
+        tags = [:reservoir_input],
+    ),
+    "reservoir_water__max_volume" => ParameterMetadata(;
```
vers-w, line 28: The `lens` is missing here:
`_.routing.river_flow.boundary_conditions.reservoir.parameters.maxstorage`

Also good to check for other parameters, see also below.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+    "reservoir_water_surface__elevation" => ParameterMetadata(;
+        lens = @optic(
+            _.routing.river_flow.boundary_conditions.reservoir.variables.waterlevel
+        ),
+        unit = Unit(; m = 1),
+        description = "Reservoir water level",
+        tags = [
+            :reservoir_state,
+            :reservoir_output,
+            :kinematic_wave_river_static_cyclic_forcing_input,
```
vers-w: I don't think this tag is correct and should be removed?

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+        description = "Area of the reservoir",
+        tags = [:reservoir_input],
+    ),
+    "reservoir_water__max_volume" => ParameterMetadata(;
+        unit = Unit(; m = 3),
+        description = "Maximum volume (above which water is spilled)",
+        tags = [:reservoir_input],
+    ),
+    "reservoir_water__rating_curve_coefficient" => ParameterMetadata(;
+        description = "Rating curve coefficient",
```
vers-w, line 38: `lens` is missing.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+        tags = [:reservoir_input],
+    ),
+    "reservoir_water__rating_curve_exponent" => ParameterMetadata(;
+        description = "Rating curve exponent",
```
vers-w, line 43: `lens` is missing.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+        unit = Unit(; mm = 1, dt = -1),
+        description = "Average potential evaporation over the reservoir area",
+        tags = [:reservoir_output],
+    ),
+
+    ## Kinematic wave
+    ### River flow
+    #### Input
+    "river__length" => ParameterMetadata(;
+        unit = Unit(; m = 1),
```
vers-w, line 195: `lens` is missing here.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+        fill = 0.0,
+        description = "River length",
+        tags = [:kinematic_wave_river_flow_input, :local_inertial_river_input],
+    ),
+    "river__width" => ParameterMetadata(;
```
vers-w, line 200: `lens` is missing here.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+        tags = [:kinematic_wave_river_flow_input, :local_inertial_river_input],
+    ),
+    "river_water__external_inflow_volume_flow_rate" => ParameterMetadata(;
+        lens = @optic(_.routing.river_flow.boundary_conditions.external_inflow),
+        unit = Unit(; m = 3, s = -1),
+        default = 0.0,
+        description = "External inflow into the river (negative for abstractions)",
+        tags = [
+            :kinematic_wave_river_static_cyclic_forcing_input,
+            :local_inertial_river_input,
```
vers-w, line 231: As for kinematic wave, this can also be cyclic/forcing for local inertial routing.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+        ],
+    ),
+    "river__slope" => ParameterMetadata(;
+        unit = Unit(; m = 1 // 1),
```
vers-w, line 236: `lens` is missing.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+            :kinematic_wave_river_output,
+            :local_inertial_river_state,
+            :local_inertial_river_output,
+        ],
+    ),
+    #### Output
+    "river_water__volume_flow_rate" => ParameterMetadata(;
+        lens = @optic(_.routing.river_flow.variables.q_av),
+        unit = Unit(; m = 3, s = -1),
+        description = "River (+ floodplain) discharge",
```
vers-w: Note that kinematic wave concept does not have a floodplain option.

vers-w (changes requested):
Finished the review, nice to have this info at one location!

Besides lots of comments, couple of general comments:

1. With these standard name maps we can also allow to write input parameters as output. Some input parameters are changed during model initialization, for example input parameter `land_water_covered__area_fraction` is split into fraction open water (excluding rivers) and river water fraction, and I guess we should exclude these parameters. Also for example not sure if we can already write floodplain profile data, we should check that. Finally, these variables are now also exposed through BMI, for example floodplain profile data (extra dimension) this not supported now (only support for layered soil variables).
2. Would be good to explicitly set in the standard name maps if variables are allowed as cyclic/forcing input (besides static), see also https://github.com/Deltares/Wflow.jl/pull/795#discussion_r2878401819.
3. The time unit for input flux parameters (day) is now missing (in the current docs included in the `Default` column of model parameter tables).

Wflow/src/network.jl
vers-w, line 1: This https://github.com/Deltares/Wflow.jl/blob/expand_on_standard_name_data/Wflow/src/io.jl#L855 works only for `land` variables or with one to one mapping (model data structure). I guess the `get_metadata` function should be added here as well.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,531 @@
...
+    "reservoir_water__rating_curve_type_count" => ParameterMetadata(;
+        lens = @optic(
+            _.routing.river_flow.boundary_conditions.reservoir.parameters.outflowfunc
+        ),
+        type = Int,
+        description = "Type of reservoir rating curve",
+        tags = [:reservoir_input],
+    ),
+    "reservoir_water_surface__initial_elevation" => ParameterMetadata(;
+        unit = Unit(; m = 1),
```
vers-w, line 74: `lens` is missing?

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    "vegetation__crop_factor" => ParameterMetadata(;
+        lens = @optic(_.land.vegetation_parameters.kc),
+        default = 1.0,
+        description = "Crop coefficient",
+        tags = [:vegetation_input],
+    ),
+    "vegetation__specific_leaf_storage" => ParameterMetadata(;
+        lens = @optic(_.land.vegetation_parameters.storage_specific_leaf),
+        unit = Unit(; mm = 1),
+        default = 0.5,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        description = "Specific leaf storage",
+        tags = [:vegetation_input],
+    ),
+    "vegetation_wood_water__storage_capacity" => ParameterMetadata(;
+        lens = @optic(_.land.vegetation_parameters.storage_wood),
+        unit = Unit(; mm = 1),
+        default = 0.0,
```
vers-w, line 213: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        description = "Storage woody part of vegetation",
+        tags = [:vegetation_input],
+    ),
+    "vegetation_canopy__light_extinction_coefficient" => ParameterMetadata(;
+        lens = @optic(_.land.vegetation_parameters.kext),
+        default = 0.5,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    "vegetation_canopy_water__mean_evaporation_to_mean_precipitation_ratio" =>
+        ParameterMetadata(;
+            lens = @optic(_.land.interception.parameters.e_r),
+            default = 0.1,
+            description = "Gash interception model parameter",
+            tags = [:vegetation_input],
+        ),
+    "vegetation_canopy__gap_fraction" => ParameterMetadata(;
+        lens = @optic(_.land.vegetation_parameters.canopygapfraction),
+        default = 0.1,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    "atmosphere_air__temperature" => ParameterMetadata(;
+        lens = @optic(_.land.atmospheric_forcing.temperature),
+        unit = Unit(; degC = 1, absolute_temperature = true),
+        description = "Temperature",
+        tags = [:atmospheric_forcing],
+    ),
+    "vegetation_root__depth" => ParameterMetadata(;
+        lens = @optic(_.land.vegetation_parameters.rootingdepth),
+        unit = Unit(; mm = 1),
+        default = 750.0,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        lens = @optic(_.mass_balance.routing.river_water_balance.error),
+        unit = Unit(; m = 3, s = -1),
+        description = "River water mass balance error",
+        tags = [:water_mass_balance_river_flow],
+    ),
+    "river_water_mass_balance_relative_error__number" => ParameterMetadata(;
+        lens = @optic(_.mass_balance.routing.river_water_balance.relative_error),
+        description = "River water mass balance relative error",
+        tags = [:water_mass_balance_river_flow],
+    ),
```
vers-w: Wrong location?

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    "land_surface_water__abstraction_volume_flux" => ParameterMetadata(;
+        lens = @optic(_.routing.river_flow.allocation.variables.act_surfacewater_abst),
+        unit = Unit(; mm = 1, dt = -1),
+        description = "Actual surface water abstraction",
+        tags = [:demand_allocation_output],
+    ),
+    "reservoir_water__outgoing_observed_volume_flow_rate" => ParameterMetadata(;
```
vers-w: Wrong location?

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        lens = @optic(
+            _.routing.river_flow.boundary_conditions.reservoir.variables.outflow_obs
+        ),
+        unit = Unit(; m = 3, s = -1),
+        default = MISSING_VALUE,
+        fill = MISSING_VALUE,
+        description = "Observed outflow reservoir",
+        tags = [:reservoir_static_cyclic_forcing_input],
+    ),
+    "reservoir_water__external_inflow_volume_flow_rate" => ParameterMetadata(;
```
vers-w: Wrong location?

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        lens = @optic(
+            _.routing.river_flow.boundary_conditions.reservoir.boundary_conditions.external_inflow
+        ),
+        unit = Unit(; m = 3, s = -1),
+        default = 0.0,
+        description = "External inflow reservoir (negative for abstractions)",
+        tags = [:reservoir_static_cyclic_forcing_input],
+    ),
+    "reservoir_water_mass_balance_error__volume_flow_rate" => ParameterMetadata(;
```
vers-w: Wrong location?

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        lens = @optic(_.mass_balance.routing.reservoir_water_balance.error),
+        unit = Unit(; m = 3, s = -1),
+        description = "Reservoir water mass balance error",
+        tags = [:water_mass_balance_reservoir],
+    ),
+    "reservoir_water_mass_balance_relative_error__number" => ParameterMetadata(;
```
vers-w: Wrong location?

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        lens = @optic(_.mass_balance.routing.reservoir_water_balance.relative_error),
+        description = "Reservoir water mass balance relative error",
+        tags = [:water_mass_balance_reservoir],
+    ),
+    "soil_exponential_vertical_saturated_hydraulic_conductivity_profile_below_surface__depth" =>
+        ParameterMetadata(;
+            lens = nothing,
```
vers-w: This parameter has a `lens`.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+            unit = Unit(; mm = 1),
+            description = "Depth from soil surface for which exponential decline of the vertical saturated hydraulic conductivity at the soil surface is valid, when an `exponential_constant` saturated hydraulic conductivity profile is used",
+            tags = [:soil_input],
+        ),
+    "soil_layer_water__vertical_saturated_hydraulic_conductivity" =>
+        ParameterMetadata(;
+            lens = nothing,
```
vers-w: This parameter has a `lens`.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+            unit = Unit(; mm = 1, dt = -1),
+            dimname = :layer,
+            description = "Vertical saturated hydraulic conductivity per soil layer, when a `layered` or `layered_exponential` saturated hydraulic conductivity profile is used",
+            tags = [:soil_input],
+        ),
+    "soil_layered_vertical_saturated_hydraulic_conductivity_profile_below_surface__depth" =>
+        ParameterMetadata(;
+            lens = nothing,
```
vers-w: Correct (`lens = nothing`) as this data is not read in directly!

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    ),
+    "soil_surface_water__infiltration_reduction_parameter" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.cf_soil),
+        default = 0.038,
+        description = "Controls soil infiltration reduction factor when soil is frozen",
+        tags = [:soil_input],
+    ),
+    "soil_water__saturated_volume_fraction" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.theta_s),
+        default = 0.6,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        description = "Saturated water content",
+        tags = [:soil_input],
+    ),
+    "soil_water__residual_volume_fraction" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.theta_r),
+        default = 0.01,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        description = "Residual water content",
+        tags = [:soil_input],
+    ),
+    "soil_surface_water__vertical_saturated_hydraulic_conductivity" =>
+        ParameterMetadata(;
+            lens = @optic(_.land.soil.parameters.kv_profile.kv_0),
+            unit = Unit(; mm = 1, dt = -1),
+            default = 3000.0,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+            description = "Vertical saturated hydraulic conductivity at soil surface",
+            tags = [:soil_input],
+        ),
+    "soil_water__vertical_saturated_hydraulic_conductivity_scale_parameter" =>
+        ParameterMetadata(;
+            lens = nothing,
```
vers-w: Add `lens` for this parameter.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        ParameterMetadata(;
+            lens = @optic(_.land.soil.parameters.alpha_h1),
+            default = 1.0,
+            description = "Root water uptake reduction at soil water pressure head h1 (0.0 or 1.0)",
+            tags = [:soil_input],
+        ),
+    "soil__thickness" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.soilthickness),
+        unit = Unit(; mm = 1),
+        default = 2000.0,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    "soil_water_saturated_zone_bottom__max_leakage_volume_flux" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.maxleakage),
+        unit = Unit(; mm = 1, dt = -1),
+        default = 0.0,
+        description = "Maximum leakage from saturated zone",
+        tags = [:soil_input],
+    ),
+    "soil_layer_water__brooks_corey_exponent" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.c),
+        default = 10.0,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        ParameterMetadata(;
+            lens = @optic(_.land.soil.parameters.kvfrac),
+            default = 1.0,
+            description = "Multiplication factor applied to vertical saturated hydraulic conductivity per soil layer",
+            dimname = :layer,
+            tags = [:soil_input],
+        ),
+    "compacted_soil__area_fraction" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.pathfrac),
+        default = 0.01,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    "soil_layer_water__brooks_corey_exponent" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.c),
+        default = 10.0,
+        description = "Brooks-Corey power coefficient per soil layer",
+        dimname = :layer,
+        tags = [:soil_input],
+    ),
+    "land_water_covered__area_fraction" => ParameterMetadata(;
+        default = 0.0,
+        description = "Fraction of open water (excluding rivers)",
```
vers-w: I think we should remove `(excluding rivers)`, as the river fraction is subtracted from this in the code. It is correct that this parameter has no `lens` because of the subtraction.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    ),
+    "soil_wet_root__sigmoid_function_shape_parameter" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.rootdistpar),
+        default = -500.0,
+        description = "Controls sharpness of transition between fully wet and fully dry roots",
+        tags = [:soil_input],
+    ),
+    "soil_root__length_density_fraction" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.rootfraction),
+        default = 1.0,
```
vers-w: This parameter has no default value.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+            description = "Averianov exponent controlling capillary rise",
+            tags = [:soil_input],
+        ),
+    "subsurface_water_mass_balance_error__volume_flow_rate" => ParameterMetadata(;
+        lens = @optic(_.mass_balance.routing.subsurface_water_balance.error),
+        unit = Unit(; m = 3, d = -1),
+        description = "Subsurface water mass balance error",
+        tags = [:water_mass_balance_subsurface_flow],
+    ),
+    "subsurface_water_mass_balance_relative_error__number" => ParameterMetadata(;
```
vers-w: Wrong location?

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+            tags = [:soil_input],
+        ),
+    "soil_water_saturated_zone_top__capillary_rise_averianov_exponent" =>
+        ParameterMetadata(;
+            lens = @optic(_.land.soil.parameters.cap_n),
+            default = 2.0,
+            description = "Averianov exponent controlling capillary rise",
+            tags = [:soil_input],
+        ),
+    "subsurface_water_mass_balance_error__volume_flow_rate" => ParameterMetadata(;
```
vers-w: Wrong location?

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+            tags = [:glacier_input],
+        ),
+    "glacier_surface__area_fraction" => ParameterMetadata(;
+        lens = @optic(_.land.glacier.parameters.glacier_frac),
+        default = 0.0,
+        fill = 0.0,
+        description = "Fraction covered by a glacier",
+        tags = [:glacier_input],
+    ),
+    "land_surface_water_mass_balance_error__volume_flow_rate" => ParameterMetadata(;
```
vers-w: Wrong location?

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        lens = @optic(_.mass_balance.routing.overland_water_balance.error),
+        unit = Unit(; m = 3, s = -1),
+        description = "Overland flow mass balance error",
+        tags = [:water_mass_balance_overland_flow],
+    ),
+    "land_surface_water_mass_balance_relative_error__number" => ParameterMetadata(;
```
vers-w: Wrong location?

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    "land__allocated_irrigation_water_volume_flux" => ParameterMetadata(;
+        lens = @optic(_.land.allocation.variables.irri_alloc),
+        unit = Unit(; mm = 1, dt = -1),
+        description = "Allocated water for irrigation",
+        tags = [:demand_allocation_output],
+    ),
+    "land__allocated_non_irrigation_water_volume_flux" => ParameterMetadata(;
+        lens = @optic(_.land.allocation.variables.nonirri_alloc),
+        unit = Unit(; mm = 1, dt = -1),
+        default = 25.0,
```
vers-w: Remove default.

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+        lens = @optic(_.routing.river_flow.concentrations.parameters.dm_lagg),
+        unit = Unit(; μm = 1),
+        default = 500.0,
+        description = "Large aggregates mean diameter",
+        tags = [:sediment_transport_cap_overland_input, :sediment_river_transport_input],
+    ),
+    "land_surface_water_sediment__govers_transport_capacity_coefficient" =>
+        ParameterMetadata(;
+            lens = @optic(_.routing.overland_flow.transport_capacity.parameters.c_govers),
+            default = 5.0e-7,
```
vers-w: This was the default value in the code:  0.000505.

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+            description = "Govers transport capacity coefficient",
+            tags = [:sediment_transport_cap_overland_input],
+        ),
+    "land_surface_water_sediment__govers_transport_capacity_exponent" =>
+        ParameterMetadata(;
+            lens = @optic(_.routing.overland_flow.transport_capacity.parameters.n_govers),
+            default = 1.5,
```
vers-w: This was the default value in the code: 4.27.

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+            description = "Govers transport capacity exponent",
+            tags = [:sediment_transport_cap_overland_input],
+        ),
+    ### Transport capacity river
+    #### Input
+    "river_sediment__median_diameter" => ParameterMetadata(;
+        lens = @optic(_.routing.river_flow.transport_capacity.parameters.d50),
+        unit = Unit(; mm = 1),
+        default = 1.0,
```
vers-w: In the code the default value is 0.1.

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+        description = "Particle mean diameter",
+        tags = [:sediment_transport_cap_river_input],
+    ),
+    "river_water_sediment__bagnold_transport_capacity_coefficient" =>
+        ParameterMetadata(;
+            lens = @optic(_.routing.river_flow.transport_capacity.parameters.c_bagnold),
+            default = 0.0017,
```
vers-w: In the code/docs there is no default value for this parameter.

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+            description = "Bagnold transport capacity coefficient",
+            tags = [:sediment_transport_cap_river_input],
+        ),
+    "river_water_sediment__bagnold_transport_capacity_exponent" => ParameterMetadata(;
+        lens = @optic(_.routing.river_flow.transport_capacity.parameters.e_bagnold),
+        default = 1.8,
```
vers-w: In the code/docs there is no default value for this parameter.

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+        description = "Bagnold transport capacity exponent",
+        tags = [:sediment_transport_cap_river_input],
+    ),
+    "river_water_sediment__kodatie_transport_capacity_a_coefficient" =>
+        ParameterMetadata(;
+            lens = @optic(_.routing.river_flow.transport_capacity.parameters.a_kodatie),
+            default = 5.0,
```
vers-w: In the code/docs there is no default value for this parameter.

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+            description = "Kodatie transport capacity coefficient a",
+            tags = [:sediment_transport_cap_river_input],
+        ),
+    "river_water_sediment__kodatie_transport_capacity_b_coefficient" =>
+        ParameterMetadata(;
+            lens = @optic(_.routing.river_flow.transport_capacity.parameters.b_kodatie),
+            default = 0.6,
```
vers-w: In the code/docs there is no default value for this parameter.

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+            description = "Kodatie transport capacity coefficient b",
+            tags = [:sediment_transport_cap_river_input],
+        ),
+    "river_water_sediment__kodatie_transport_capacity_c_coefficient" =>
+        ParameterMetadata(;
+            lens = @optic(_.routing.river_flow.transport_capacity.parameters.c_kodatie),
+            default = 0.05,
```
vers-w: In the code/docs there is no default value for this parameter.

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+            description = "Kodatie transport capacity coefficient c",
+            tags = [:sediment_transport_cap_river_input],
+        ),
+    "river_water_sediment__kodatie_transport_capacity_d_coefficient" =>
+        ParameterMetadata(;
+            lens = @optic(_.routing.river_flow.transport_capacity.parameters.d_kodatie),
+            default = 2.0,
```
vers-w: In the code/docs there is no default value for this parameter.

Wflow/src/standard_name/standard_name_utils.jl
```diff
@@ -0,0 +1,109 @@
...
+
+"""
+Metadata associated with parameters and variables.
+
+# Arguments
+- `lens`: The path in the model data structure to the parameter/variable if it exists
+- `unit`: The unit of the parameter/variable in the Wflow input
+- `default`: The default (initial) value of the parameter/variable if it exists
+- `fill`: Missing input values are replaced by this value if allow_missing == false
+- `type`: The output type of the data. Assumed to be `Float64` if it is not provided and cannot be derived
```
vers-w, line 18: formatting

Wflow/src/Wflow.jl
```diff
@@ -224,7 +228,19 @@ include("sediment_flux.jl")
...
 include("routing/initialize_routing.jl")
 include("sbm_gwf_model.jl")
-include("standard_name.jl")
+include("standard_name/standard_name_utils.jl")
+include("standard_name/standard_name_domain.jl")
+include("standard_name/standard_name_routing.jl")
+include("standard_name/standard_name_sbm.jl")
+include("standard_name/standard_name_sediment.jl")
+
+const standard_name_maps = (
```
vers-w: For consistency:

```suggestion
const STANDARD_NAME_MAPS = (
```

Wflow/src/standard_name/standard_name_domain.jl
```diff
@@ -0,0 +1,30 @@
+# NOTE: The order of the entries determines the order in the docs tables
```
vers-w, line 1: `sbm` and `sediment` standard name mapping have a docstring and `domain` and `routing` are missing one. Suggest to include also a docstring for `domain` and `routing` .

Wflow/src/io.jl
vers-w, line 1: This https://github.com/Deltares/Wflow.jl/blob/expand_on_standard_name_data/Wflow/src/io.jl#L286 also works only for `land` variables.

Wflow/src/standard_name/standard_name_utils.jl
```diff
@@ -0,0 +1,109 @@
...
+            type,
+            description,
+            allow_missing,
+            dimname,
+            flags,
+        )
+    end
+end
+
+get_metadata(name::AbstractString, ::Type{<:Writer}) = ParameterMetadata()
```
vers-w: Do we need this? Where is this called?

Wflow/src/units.jl
```diff
@@ -0,0 +1,343 @@
...
+    (factor = 1e-3, unit_SI = Unit(; m = 3)), # L
+    (factor = 1.0, unit_SI = Unit(; kg = 1)), # kg
+    (factor = 1e-3, unit_SI = Unit(; kg = 1)), # g
+    (factor = 1e3, unit_SI = Unit(; kg = 1)), # t
+    (factor = 1.0, unit_SI = Unit(; kg = 1, m = 2, s = -2)), # J
+    (factor = 1e-2, unit_SI = Unit()), # percentage
+    (factor = 1e-6, unit_SI = Unit()), # ppm
+]
+
+# Predefined units used within the code
```
vers-w, line 123: It seems not all units are used within the code, suggest to only keep the units that are used.

Wflow/src/units.jl
```diff
@@ -0,0 +1,343 @@
...
+    unit::Unit;
+    dt_val::Union{Nothing, Number} = nothing,
+)
+    unit_ref = Ref(unit)
+    @. x = from_SI(x, unit_ref; dt_val)
+end
+
+"""
+Get the unit by checking all `*_standard_name_map` dictionaries
+and make sure that there is no ambiguity.
```
vers-w: formatting.

Wflow/src/units.jl
```diff
@@ -0,0 +1,343 @@
...
+end
+
+# Fallback method for non-Floats, e.g. nothing, missing, Bool, Int
+from_SI(x, unit::Unit; kwargs...) = x
+
+"""
+Convert the given values of the SI equivalent of the given unit to the values in the given unit in-place
+"""
+function from_SI!(
+    x::Union{AbstractArray},
```
vers-w: Why `Union`?

Wflow/src/units.jl
```diff
@@ -0,0 +1,343 @@
...
+    else
+        x / to_SI_factor(unit; dt_val)
+    end
+end
+
+# Fallback method for non-Floats, e.g. nothing, missing, Bool, Int
+from_SI(x, unit::Unit; kwargs...) = x
+
+"""
+Convert the given values of the SI equivalent of the given unit to the values in the given unit in-place
```
vers-w:
```suggestion
In-place conversion of values in SI unit to the given unit.
```

Wflow/src/units.jl
```diff
@@ -0,0 +1,343 @@
...
+Convert an array of values to the values in the corresponding SI unit in-place
+"""
+function to_SI!(x::AbstractArray, unit::Unit; dt_val::Union{Nothing, Number} = nothing)
+    unit_ref = Ref(unit)
+    @. x = to_SI(x, unit_ref; dt_val)
+    return x
+end
+
+"""
+Convert the given value of the SI equivalent of the given unit to the value in the given unit
```
vers-w:
```suggestion
Convert a value in SI unit to the given unit.
```

Wflow/src/units.jl
```diff
@@ -0,0 +1,343 @@
...
+    else
+        x * to_SI_factor(unit; dt_val)
+    end
+end
+
+# Fallback method for non-Floats, e.g. nothing, missing, Bool, Int
+to_SI(x, unit::Unit; kwargs...) = x
+
+"""
+Convert an array of values to the values in the corresponding SI unit in-place
```
vers-w:
```suggestion
In-place conversion of an array of values to the corresponding SI unit.
```

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    "snowpack__melting_temperature_threshold" => ParameterMetadata(;
+        lens = @optic(_.land.snow.parameters.ttm),
+        unit = Unit(; degC = 1, absolute_temperature = true),
+        default = 0.0,
+        description = "Threshold temperature for snowmelt",
+        tags = [:snow_input],
+    ),
+    "snowpack__degree_day_coefficient" => ParameterMetadata(;
+        lens = @optic(_.land.snow.parameters.cfmax),
+        unit = Unit(; mm = 1, degC = -1, d = -1),
```
vers-w, line 327: Is this the external unit (from the input data) or the internal unit (not in SI)? If it is the external unit then the unit (and default value) is correct. However. for a similar glacier input parameter (L. 594) the unit is `Unit(; mm = 1, degC = -1, dt = -1)`. This is also important for the docs (units).

docs/model_docs/parameters_landhydrology_sbm.qmd
```diff
@@ -1,268 +1,159 @@
...
+```{julia}
+# | code-fold: true
+generate_table(Wflow.LandHydrologySBM, :vegetation_output; relative_widths = [4, 3, 1])
+```
 ## Soil
 ### Input
-Note that the input parameter water content at field capacity
-(`soil_water__field_capacity_volume_fraction`) is optional. When not provided the water
-content is computed at a pressure head of -100 cm using the Brooks and Corey water retention
-model.
```
vers-w, line 116: This part should not be removed.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+    "irrigated_paddy_area__count" => ParameterMetadata(;
+        lens = @optic(_.land.demand.paddy.parameters.irrigation_areas),
+        type = Int,
+        description = "Irrigation areas (paddy)",
+        tags = [:demand_paddy_irrigation_input],
+    ),
+    "irrigated_paddy__irrigation_trigger_flag" => ParameterMetadata(;
+        lens = @optic(_.land.demand.paddy.parameters.irrigation_trigger),
+        type = Bool,
+        description = "Irrigation trigger (off or on, 0 or 1)",
```
vers-w:
```suggestion
        description = "Paddy irrigation trigger flag",
```

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+        lens = @optic(_.land.rainfall_erosion.parameters.soilcover_fraction),
+        default = 0.01,
+        description = "Fraction of the soil that is covered (eg paved, snow, etc)",
+        tags = [:rainfall_erosion_input],
+    ),
+    "soil_erosion__usle_k_factor" => ParameterMetadata(;
+        lens = @optic(_.land.overland_flow_erosion.parameters.usle_k),
+        default = 0.1,
+        description = "USLE soil erodibility factor",
+        tags = [:overland_flow_erosion_input],
```
vers-w: This is also input for rainfall erosion.

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
...
+    ),
+    "soil_erosion__usle_c_factor" => ParameterMetadata(;
+        lens = @optic(_.land.overland_flow_erosion.parameters.usle_c),
+        default = 0.01,
+        description = "USLE crop management factor",
+        tags = [:overland_flow_erosion_input],
```
vers-w: This is also input for rainfall erosion.

Wflow/src/standard_name/standard_name_routing.jl
vers-w: This is not input for kinematic wave river.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
...
+        lens = @optic(_.land.soil.variables.zi),
+        unit = Unit(; mm = 1),
+        description = "Pseudo-water table depth (top of the saturated zone)",
+        tags = [:soil_output],
+    ),
+    "land_water_mass_balance_error__volume_flux" => ParameterMetadata(;
+        lens = @optic(_.mass_balance.land_water_balance.error),
+        unit = Unit(; mm = 1, dt = -1),
+        description = "Land water mass balance error",
+        tags = [:water_mass_balance_land_hydrology, :water_mass_balance_overland_flow],
```
vers-w: This variable is not part of overland flow.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,799 @@
+# NOTE: The order of the entries determines the order in the docs tables
+"""
+Mapping of (CSDMS) standard names to model variables and units for models with a land model
+of type `LandHydrologySBM`. The `lens` allows access to a nested model variable.
+"""
+const sbm_standard_name_map = OrderedDict{String, ParameterMetadata}(
```
vers-w, line 7: As for `routing` and `domain` would be nice to order the variables also here (input etc.).

Wflow/src/standard_name/standard_name_sediment.jl
```diff
@@ -0,0 +1,554 @@
+# NOTE: The order of the entries determines the order in the docs tables
+"""
+Mapping of (CSDMS) standard names to model variables and units for models with a land model
+of type `SoilLoss`. The `lens` allows access to a nested model variable.
+"""
+const sediment_standard_name_map = OrderedDict{String, ParameterMetadata}(
```
vers-w, line 6: As for `routing` and `domain` would be nice to order the variables also here (input etc.).

docs/model_docs/parameters_sediment.qmd
```diff
@@ -1,86 +1,61 @@
...
-  word-wrap: break-word;
-}
-</style>
-
 ## Atmospheric forcing
-| Standard Name | Description | Unit | Default |
-| ------ | ---- | -- | -- |
-| `atmosphere_water__precipitation_volume_flux` | Precipitation | mm Δt⁻¹ | - |
-: {.striped .hover}
+```{julia}
```
vers-w: Locally I got this output here in the doc page, below Atmospheric forcing:

```
Status `\.julia\dev\Wflow\docs\Project.toml`
⌃ [13f3f980] CairoMakie v0.13.10
  [a93c6f00] DataFrames v1.8.1
⌃ [7073ff75] IJulia v1.34.2
⌃ [54e234d5] MakieExtra v0.1.69
  [1862ce21] MarkdownTables v1.1.0
⌃ [08abe8d2] PrettyTables v3.1.2
  [d48b7d99] Wflow v1.1.0-dev `Wflow`
Info Packages marked with ⌃ have new versions available and may be upgradable.
```

Wflow/test/utils.jl
vers-w, line 1: Noticed that test coverage went down a bit, would be good to check why and if it can be increased.

Wflow/src/standard_name/standard_name_sbm.jl
SouthEndMusic: I don't understand exactly what should be removed

Wflow/src/standard_name/standard_name_routing.jl
SouthEndMusic: Same as my comment elsewhere, I don't think it makes sense to add a lens for this because this is probably overwritten each timestep.

Wflow/src/units.jl
SouthEndMusic: That's ok, they are used in the code in https://github.com/Deltares/Wflow.jl/pull/735

Wflow/src/standard_name/standard_name_utils.jl
SouthEndMusic: Not yet used, will be in https://github.com/Deltares/Wflow.jl/pull/735.

Wflow/src/units.jl
SouthEndMusic: How should the formatting be improved?

docs/model_docs/parameters_sediment.qmd
SouthEndMusic: Ah right, there was some unnessecary code there

Wflow/src/standard_name/standard_name_routing.jl
SouthEndMusic: There is a table for 'Static or cyclic/forcing input' under Kinematic wave -> River flow -> Input; there is no such table under Local inertial at `parameters_routing.qmd`. Should it be added?

Wflow/src/units.jl
SouthEndMusic: nvm, got rid of the function entirely

Wflow/src/standard_name/standard_name_routing.jl
vers-w: Yes, correct. We decided to not support writing (output) static input parameters using the standard_name. So, only input parameters that can be provided as cyclic or forcing need a `lens`. This also needs to be checked for other static input parameters (did notice that some other static input parameters do have a `lens`).

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,735 @@
...
+        unit = Unit(; mm = 1, dt = -1),
+        default = 0.0,
+        description = "Net domestic water demand",
+        allow_dynamic_input = true,
+        tags = [:demand_non_irrigation_static_cyclic_forcing_input],
+    ),
+    "glacier_firn_accumulation__snowpack_dry_snow_leq_depth_fraction" =>
+        ParameterMetadata(;
+            lens = @optic(_.land.glacier.parameters.g_sifrac),
+            unit = Unit(; dt = -1),
```
JoostBuitink: Input unit should be `day-1`

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,735 @@
...
+    "atmosphere_water__precipitation_volume_flux" => ParameterMetadata(;
+        lens = @optic(_.land.atmospheric_forcing.precipitation),
+        unit = Unit(; mm = 1, dt = -1),
+        description = "Precipitation",
+        allow_dynamic_input = true,
+        tags = [:atmospheric_forcing],
+    ),
+    "compacted_soil_surface_water__infiltration_capacity" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.infiltcappath),
+        unit = Unit(; mm = 1, dt = -1),
```
JoostBuitink: Input unit is mm/day

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,735 @@
...
+        ParameterMetadata(;
+            lens = @optic(_.land.soil.parameters.kv_profile.z_exp),
+            unit = Unit(; mm = 1),
+            description = "Depth from soil surface for which exponential decline of the vertical saturated hydraulic conductivity at the soil surface is valid, when an `exponential_constant` saturated hydraulic conductivity profile is used",
+            tags = [:soil_input],
+        ),
+    "soil_layer_water__vertical_saturated_hydraulic_conductivity" =>
+        ParameterMetadata(;
+            lens = @optic(_.land.soil.parameters.kv_profile.kv),
+            unit = Unit(; mm = 1, dt = -1),
```
JoostBuitink: Input unit is mm/day

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,735 @@
...
+    "soil_surface_water__infiltration_reduction_parameter" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.cf_soil),
+        default = 0.038,
+        description = "Controls soil infiltration reduction factor when soil is frozen",
+        tags = [:soil_input],
+    ),
+    "soil_surface_water__vertical_saturated_hydraulic_conductivity" =>
+        ParameterMetadata(;
+            lens = @optic(_.land.soil.parameters.kv_profile.kv_0),
+            unit = Unit(; mm = 1, dt = -1),
```
JoostBuitink: Input unit is mm/day

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,735 @@
...
+        tags = [:soil_input],
+    ),
+    "soil_water__saturated_volume_fraction" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.theta_s),
+        description = "Saturated water content",
+        tags = [:soil_input],
+    ),
+    "soil_water_saturated_zone_bottom__max_leakage_volume_flux" => ParameterMetadata(;
+        lens = @optic(_.land.soil.parameters.maxleakage),
+        unit = Unit(; mm = 1, dt = -1),
```
JoostBuitink: Input unit is mm/day

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,735 @@
...
+    ),
+    "compacted_soil_surface_water__excess_volume_flux" => ParameterMetadata(;
+        lens = @optic(_.land.soil.variables.excesswatersoil),
+        unit = Unit(; mm = 1, dt = -1),
+        description = "Excess water for compacted soil fraction",
+        tags = [:soil_output],
+    ),
+    "domestic__gross_water_demand_volume_flux" => ParameterMetadata(;
+        lens = @optic(_.land.demand.domestic.demand.demand_gross),
+        unit = Unit(; mm = 1, dt = -1),
```
JoostBuitink: Input unit is mm/day. 

Note: This holds for all `"$(sector)__gross_water_demand_volume_flux"` and `"$(sector)__net_water_demand_volume_flux"` fields

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,735 @@
...
+    "irrigated_non_paddy__irrigation_trigger_flag" => ParameterMetadata(;
+        lens = @optic(_.land.demand.nonpaddy.parameters.irrigation_trigger),
+        type = Bool,
+        description = "Non-paddy irrigation trigger flag",
+        allow_dynamic_input = true,
+        tags = [:demand_non_paddy_irrigation_static_cyclic_forcing_input],
+    ),
+    "irrigated_non_paddy__max_irrigation_rate" => ParameterMetadata(;
+        lens = @optic(_.land.demand.nonpaddy.parameters.maximum_irrigation_rate),
+        unit = Unit(; mm = 1, dt = -1),
```
JoostBuitink: Input units are mm/day

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -0,0 +1,735 @@
...
+    "irrigated_paddy__max_depth" => ParameterMetadata(;
+        lens = @optic(_.land.demand.paddy.parameters.h_max),
+        unit = Unit(; mm = 1),
+        default = 80.0,
+        description = "Water depth when paddy field starts spilling water (overflow)",
+        tags = [:demand_paddy_irrigation_input],
+    ),
+    "irrigated_paddy__max_irrigation_rate" => ParameterMetadata(;
+        lens = @optic(_.land.demand.paddy.parameters.maximum_irrigation_rate),
+        unit = Unit(; mm = 1, dt = -1),
```
JoostBuitink: Input units are in mm/day

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,639 @@
...
+            description = "A multiplication factor applied to vertical hydraulic conductivity",
+            tags = [:kinematic_lateral_subsurface_input],
+        ),
+    #### States
+    "subsurface_water__volume_flow_rate" => ParameterMetadata(;
+        lens = @optic(_.routing.subsurface_flow.variables.q_av),
+        unit = Unit(; m = 3, d = -1),
+        description = "Subsurface flow",
+        tags = [:kinematic_lateral_subsurface_state, :kinematic_lateral_subsurface_output],
+    ),
```
JoostBuitink: I think this is no longer a state, as it is replace by the instantaneous one. Pinging @vers-w to be sure!

Wflow/src/standard_name/standard_name_routing.jl
JoostBuitink: Yes, this parameter should also be labelled as cyclic/forcing for local inertial routing!

Wflow/src/standard_name/standard_name_sbm.jl
JoostBuitink, line 1: General comment on the tables: good to make a distinction between input units and internal units. Input units are often mm/day, which are internally corrected to mm/dt. Issue #653 also mentions this, and would be good to improve here as well. Perhaps also add a couple of sentences that explain this in the parameters_intro.qmd file

JoostBuitink (changes requested):
Nice work, I really like the automatic parameter tables! A couple of comments, mostly on the units

Wflow/src/standard_name/standard_name_sbm.jl
vers-w: The `description` is not correct, it should be: "Fraction of open water".

Wflow/src/standard_name/standard_name_routing.jl
vers-w: We decided to not allow writing static input parameters, so these parameters do not need a `lens`. And to the metadata the field `allow_dynamic_input` has been added, nice.

SouthEndMusic: I'd say that having the lens in general for all parameters and variables is a good thing, e.g. for if you want to request them via BMI. If certain things should not be allowed, that should be added to the metadata explicitly.

Wflow/src/standard_name/standard_name_routing.jl
vers-w: Yes, correct the state should be `subsurface_water__instantaneous_volume_flow_rate`.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,639 @@
...
+        unit = Unit(; m = 1),
+        description = "Water level of reservoir (used for initialization)",
+        tags = [:reservoir_input],
+    ),
+    #### Static or cyclic/forcing input
+    "reservoir_water_demand__required_downstream_volume_flow_rate" =>
+        ParameterMetadata(;
+            unit = Unit(; m = 3, s = -1),
+            description = "Minimum (environmental) flow released from reservoir",
+            allow_dynamic_input = true,
```
vers-w, line 86: If `allow_dynamic_input` is `true` a `lens` needs to be added. See also couple of reservoir variables below.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,639 @@
...
+        tags = [:local_inertial_river_input],
+    ),
+    "model_boundary_condition_river_bank_water__depth" => ParameterMetadata(;
+        default = 0,
+        unit = Unit(; m = 1),
+        description = "Boundary condition bankfull depth downstream river outlets",
+        tags = [:local_inertial_river_input],
+    ),
+    "river_bank_water__elevation" => ParameterMetadata(;
+        unit = Unit(; m = 1),
```
vers-w, line 392: Suggest to add a `lens` here.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,639 @@
...
+    ),
+    ### 1D floodplain flow
+    #### Input
+    "floodplain_water__sum_of_volume_per_depth" => ParameterMetadata(;
+        unit = Unit(; m = 3),
+        dimname = :flood_depth,
+        description = "Floodplain profile (cumulative volume per flood depth)",
+        tags = [:local_inertial_floodplain_1D_flow_input],
+    ),
+    "floodplain_water_flow__manning_n_parameter" => ParameterMetadata(;
```
vers-w, line 405: Suggest to add a `lens` here.

Wflow/src/standard_name/standard_name_routing.jl
```diff
@@ -0,0 +1,639 @@
...
+    ),
+    "floodplain_water__volume_flow_rate" => ParameterMetadata(;
+        lens = @optic(_.routing.river_flow.floodplain.variables.q_av),
+        unit = Unit(; m = 3, s = -1),
+        description = "Floodplain discharge",
+        tags = [:local_inertial_floodplain_1D_flow_output],
+    ),
+    ### Overland flow
+    #### Input
+    "land_surface_water_flow__ground_elevation" => ParameterMetadata(;
```
vers-w, line 446: Suggest to add `lens` here.

vers-w (approved):
LGTM. 

A couple of small comments, one about `allow_dynamic_input` needing a `lens` (noticed this for reservoir parameters) and would be good to add a `lens` for input parameters when possible (checked this for a couple of parameters, maybe there are more). 

Finally, test coverage is still a bit lower, would be could to check if this can be increased.

Wflow/src/standard_name/standard_name_sbm.jl
```diff
@@ -216,7 +216,7 @@ const sbm_standard_name_map = OrderedDict{String, ParameterMetadata}(
     ),
     "irrigated_paddy__gross_water_demand_volume_flux" => ParameterMetadata(;
         lens = @optic(_.land.demand.paddy.variables.demand_gross),
-        unit = Unit(; mm = 1, dt = -1),
+        unit = Unit(; mm = 1, d = -1),
```
JoostBuitink, line 219: Given that this is an output variable, the unit is (I assume) mm/dt, and not mm/day (as is the case for the input variables). Can you check that? Also, I see these also have a default value (comment from @vers-w above), which does not make sense for output variables

Wflow/src/standard_name/standard_name_routing.jl
SouthEndMusic: That doesn't seem to be possible since this data doesn't directly end up in the model:

https://github.com/Deltares/Wflow.jl/blob/15e9d269f0ec667b62ac18b92cbf1524945327b6/Wflow/src/routing/surface/surface_local_inertial.jl#L44-L47

Wflow/src/standard_name/standard_name_routing.jl
SouthEndMusic: Some selection is being made:

https://github.com/Deltares/Wflow.jl/blob/15e9d269f0ec667b62ac18b92cbf1524945327b6/Wflow/src/routing/surface/surface_local_inertial.jl#L591-L593

Wflow/src/standard_name/standard_name_routing.jl
vers-w: Ah, yes, correct, no `lens` required here.

SouthEndMusic:
@vers-w I got the following list of input parameters without a lens, and for all of the ones I checked there was a reason there was no lens to define:

```
glacier_ice__initial_leq_depth
land_water_covered__area_fraction
soil_layered_vertical_saturated_hydraulic_conductivity_profile_below_surface__depth
subbasin_location__count
basin__local_drain_direction
basin_pit_location__mask
river_location__mask
land_water_allocation_area__count
reservoir_area__count
reservoir_water_flow_threshold_level__elevation
reservoir_lower_location__count
reservoir_water_surface__initial_elevation
reservoir_water_demand__required_downstream_volume_flow_rate
reservoir_water_release_below_spillway__max_volume_flow_rate
reservoir_water__target_full_volume_fraction
reservoir_water__target_min_volume_fraction
model_boundary_condition_river__length
model_boundary_condition_river_bank_water__depth
river_bank_water__elevation
floodplain_water__sum_of_volume_per_depth
floodplain_water_flow__manning_n_parameter
land_surface_water_flow__ground_elevation
land_surface__elevation
land_drain_location__mask
```

Wflow/src/standard_name/standard_name_routing.jl
vers-w: Such selection (catchment/land `indices`) is always made, so I think it is fine to add a `lens` here.

vers-w:
> @vers-w I got the following list of input parameters without a lens, and for all of the ones I checked there was a reason there was no lens to define:

Thanks for this list, I think it is fine that these do not have a `lens`, also tagging @JoostBuitink to check.

JoostBuitink (commented):
i know it is already merged, but just one comment I had still open

atsiokanos:
Nice!

Issue #808, opened 2026-01-12
Update link to Wflow executable in the docs


hboisgon:
### Wflow version checks

- [x] I have checked that the issue still exists on the latest versions of the docs on `master` [here](https://github.com/Deltares/Wflow.jl)

### Kind of issue

Docs are wrong

### Location of the documentation

https://deltares.github.io/Wflow.jl/dev/getting_started/install.html#installing-the-compiled-executable

### Documentation problem

The download link to the Deltares page is wrong. 

### Suggested fix for documentation

Should be https://download.deltares.nl/wflow

PR #817, opened 2026-01-15
Prevent reservoir `storage_input` to become negative

### LINKED ISSUE #814: check fluxes for reservoirs
### Wflow version checks

### Reproducible Example

When a reservoir is empty, negative outflows can appear. This is not correct and causes the kinematic wave to fail. It should also not happen, and is most likely related to a floating point error. This should be prevented 

### Current behaviour

Relevant lines of code: 
https://github.com/Deltares/Wflow.jl/blob/2ad283652ac478cd4fc571280147ae95b8d73541/Wflow/src/routing/reservoir.jl#L566-L570

https://github.com/Deltares/Wflow.jl/blob/2ad283652ac478cd4fc571280147ae95b8d73541/Wflow/src/routing/reservoir.jl#L459

### Desired behaviour

Only positive or zero outflow

### Additional Context

_No response_

JoostBuitink:
## Issue addressed
Fixes #814 and #808

## Explanation
Limit the `storage_input` flux to 0, as it could become negative due to floating point errors (when the reservoir is empty and there is some evaporation. Also updated the link to the download portal

## Checklist

### DIFF
--- a/docs/changelog.qmd
+++ b/docs/changelog.qmd
@@ -10,9 +10,9 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## Unreleased
 
 ### Fixed
-- Fixed issue with lake rating curve type 2 (H-Q equation) where the outflow could become
-  negative for very low lake storage values, due to floating point precision errors. The
-  `storage_input` variable is now limited to a minimum of zero.
+- Fixed issue with where the reservoir outflow could become negative for very low reservoir
+  storage values, due to floating point precision errors. The `storage_input` variable is
+  now limited to a minimum of zero.
 
 ## v1.0.1 - 2026-01-07
 


SouthEndMusic:
This PR could use a more descriptive name 😄

docs/changelog.qmd
```diff
@@ -7,6 +7,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this
 project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+- Fixed issue with lake rating curve type 2 (H-Q equation) where the outflow could become
```
vers-w: The fix is not only for rating curve type 2, right? Maybe better to use the term `reservoir` instead of `lake`, consistent with code and docs?

JoostBuitink: Good point, wrote it a bit too fast :P Is improved now!

PR #825, opened 2026-01-19
Scale and offset transform bug fix

### LINKED ISSUE #821: Wflow throws error when scale or offset parameters not entered
### Wflow version checks

### Reproducible Example

A static map can be specified in the toml file like this:

```
[input.static.some_variable]
netcdf_variable_name = "some_variable"
```

Wflow throws an error when scale and/or offset are not specified. It tries to multiply the netcdf variable with an empty float (scale).  

Ofcourse it is easily solved by entering a scale of 1.0, or by entering the variable without the separate header. However I feel the above should also be allowed?

### Current behaviour

A static map can be specified in the toml file like this:

```
[input.static.some_variable]
netcdf_variable_name = "some_variable"
```

Wflow throws an error when scale and/or offset are not specified. It tries to multiply the netcdf variable with an empty float (scale).  

### Desired behaviour

Ofcourse it is easily solved by entering a scale of 1.0, or by entering the variable without the separate header. However I feel the above should also be allowed?

### Additional Context

_No response_

SouthEndMusic:
## Issue addressed
Fixes #821

## Explanation
Scale and offset are converted to arrays in the TOML parse. For both operations it is now checked separately whether they have to be applied, and when they have length 1 they are treated as scalars.

### DIFF
--- a/docs/changelog.qmd
+++ b/docs/changelog.qmd
@@ -19,6 +19,8 @@ project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Limiting kinematic wave subsurface flow that flows partly below the soil bottom while
   solving the kinematic wave has been fixed. The computation to limit subsurface flow was
   not correct resulting in an subsurface flow overestimation.
+- A bug was fixed that caused Wflow to error when parameters in the TOML didn't have a specified
+  `scale` and/or `offset` value.
 
 ## v1.0.1 - 2026-01-07
 


Wflow/src/io.jl
```diff
@@ -27,11 +27,8 @@ function get_at(
 end
 
 function get_at(ds::CFDataset, var::InputEntry, i)
-    (; scale, offset) = var
     data = read_standardized(ds, variable_name(var), (x = :, y = :, time = i))
-    if scale != 1.0 || offset != 0.0
-        data .= data .* scale .+ offset
-    end
+    apply_affine_transform!(data, var)
```
vers-w, line 31: Does this change have any impact on performance? 

Note that for forcing time series we do expect only one value for `scale` and `offset`. The exception (multiple values) are `SVectors` (layered data) for static input and these are handled in the function `ncread`, see also https://github.com/Deltares/Wflow.jl/blob/affine_transform_bug_fix/Wflow/src/utils.jl#L346.

Wflow/src/utils.jl
```diff
@@ -233,6 +233,25 @@ function get_var(config::Config, parameter::AbstractString; optional = true)
     return var
 end
 
+function apply_affine_transform!(v::Union{AbstractArray, Number}, var::InputEntry)
```
vers-w: Suggest to add a docstring.

SouthEndMusic: @jurianbeunk 😄

Wflow/src/io.jl
SouthEndMusic: I don't have a complete overview over all possible cases here. Note that the function `apply_affine_transform!` itself contains the checks whether the scaling is 1 and the offset is 0. 
[Here](https://github.com/Deltares/Wflow.jl/blob/4debdc5a84b5bb8d1a1bad577114f622716c830c/Wflow/src/config_init.jl#L123) scalars are converted to vectors.
I do wonder though whether checking each entry being 0 or 1 can be expensive. A possible workaround would be to make the default values `[NaN]` or `[]` (see also https://github.com/Deltares/Wflow.jl/issues/821#issuecomment-3771604232) and not perform the scaling/translation with that value.

vers-w: I also don't expect that the function `apply_affine_transform!` (that also includes a `length` check compared to the previous implementation) is more expensive than the previous implementation. Anyway, would be good to check/confirm this.

Wflow/src/utils.jl
```diff
@@ -233,6 +233,30 @@ function get_var(config::Config, parameter::AbstractString; optional = true)
     return var
 end
 
+"""
+Apply the affine transform in `var` to the incoming array `A` in place element-wise.
+The affine transform consists of a scaling by `scale` and a translation by `offset`.
+These operations are only applied when non-trivial.
+"""
+function apply_affine_transform!(A::Union{AbstractArray, Number}, var::InputEntry)
```
vers-w: Do we expect a `Number` type for `A`? I would expect only vectors?

Wflow/src/config_init.jl
```diff
@@ -135,11 +135,6 @@ function init_config_section(::Type{InputEntry}, dict::AbstractDict{String})
             end
         end
 
-        # Add a proper amount of the default values
-        len = coalesce(len, 0)
```
SouthEndMusic, line 139: This was probably the source of the empty list that caused the error in the stack trace provided in the issue

PR #840, opened 2026-02-10
Add missing update `total_storage` for model type `sbm_gwf`

### LINKED ISSUE #839: Total water storage not updated model type `sbm_gwf`
The variable `total_storage` as part of the `SBM` soil concept is not updated for model type `sbm_gwf`, it is only implemented for model type `sbm`. This can be fixed as part of patch release v1.0.2.

vers-w:
## Issue addressed
Fixes #839

### DIFF
--- a/Wflow/test/run_sbm_gwf.jl
+++ b/Wflow/test/run_sbm_gwf.jl
@@ -43,6 +43,7 @@
         @test sbm.soil.parameters.theta_s[1] ≈ 0.44999998807907104
         @test sbm.soil.variables.runoff[1] == 0.0
         @test sbm.soil.variables.soilevap[1] == 0.0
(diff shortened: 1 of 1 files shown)


Wflow/test/run_sbm_gwf.jl
```diff
@@ -43,7 +43,8 @@
         @test sbm.soil.parameters.theta_s[1] ≈ 0.44999998807907104
         @test sbm.soil.variables.runoff[1] == 0.0
         @test sbm.soil.variables.soilevap[1] == 0.0
-        @test sbm.soil.variables.transpiration[1] ≈ 0.30587632831650247
```
JoostBuitink, line 46: This test on `transpiration` was probably removed by accident, is that correct?

PR #848, opened 2026-02-19
Use PackageCompiler branch to avoid bundling workspace artifacts

visr:
This switches to https://github.com/JuliaLang/PackageCompiler.jl/pull/1090. Ribasim has been using this for a while, and I expect this to get merged.


visr:
I noticed the build CLI job on 1.12 was failing so I added https://github.com/Deltares/Wflow.jl/pull/848/commits/7f02c8d96fe0c6e6bddba98ce0e5c08634f48cf6 here, and kicked off a new TeamCity job on this branch to check if it is enough to fix it.

PR #849, opened 2026-02-26
Switch from pre-commit to prek

### LINKED ISSUE #765: Format code via pre-commit

SouthEndMusic:
Fixes https://github.com/Deltares/Wflow.jl/issues/765

We used to have large diffs for pull requests due to different people using different versions of `JuliaFormatter`. In this PR:
- We change from `JuliaFormatter` to `Runic` which is more stable
- We now have a pre-commit hook for formatting. Since @visr experienced problems calling Runic via `pre-commit`, we converted from `pre-commit` to `prek`, a rewrite of `pre-commit` in Rust
- Since the Julia VSCode plugin does not yet support formatters other than `JuliaFormatter`, `Runic` is called via `JETLS`.

To make this work you need to add `.julia/bin` to your `PATH`. See also the updated developer guide.


docs/developments/guide.qmd
```diff
@@ -10,7 +10,7 @@ example so we are able to reproduce the issue. Furthermore, we welcome contribut
 the [ColPrac guide for collaborative practices](https://github.com/SciML/ColPrac). New
 contributors should make sure to read that guide.
 
-To get started with local development, install Julia and pre-commit by running the command below.
+To get started with local development, install Julia and prek (pre-commit) by running the command below.
```
visr:
```suggestion
To get started with local development, install Julia and prek (for pre-commit hooks) by running the command below.
```

docs/developments/guide.qmd
```diff
@@ -10,7 +10,7 @@ example so we are able to reproduce the issue. Furthermore, we welcome contribut
...
```
visr, line 14: You should also document that developers should put `.julia\bin` in their path. This is where runic and jetls will be. Runic can be installed globally in Pkg mode with `app add Runic`.

You can possibly include that in the install task with

```
julia --eval 'using Pkg; Pkg.Apps.add("Runic")'
or as a task
install-runic = { cmd = "julia --eval='using Pkg; Pkg.Apps.add(\"Runic\")'" }
```

I'll add this to the Ribasim install step as well.

.vscode/extensions.json
visr, line 1: In settings.json you need:
```
    "[julia]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "aviatesk.jetls-client"
    },
```

Since Julia-VSCode doesn't support other formatters at the moment, we delegate this task to the JETLS plugin.

PR #870, opened 2026-03-18
Support kinematic wave floodplain routing

### LINKED ISSUE #235: Add 1D floodplain schematization to kinematic wave routing
For local inertial river routing a 1D floodplain schematization has been implemented, and it makes sense to add this functionality also to the kinematic wave routing. The current kinematic wave solution assumes a fixed `alpha` value. This will not work properly with a 1D floodplain schematization, so some research work (alternative solver) is required.

vers-w:
## Issue addressed
Fixes #235

## Explanation
This PR is about:

- Support of 1D floodplain schematization for kinematic wave routing, making use of the same 1D floodplain schematization that is already in place for local inertial river flow routing. Routing is done separately for the river channel and floodplain. Floodplain water is routed down slope using the `accucapacityflux` function with a transport capacity based on Manning's equation.
- Added for river and floodplain flow on a staggered grid the option to use the Manning's equation, in addition to the local inertial equation, for applications where water flow is mostly controlled by the topography.

Besides these implementations river and overland flow routing (except local inertial overland flow) has been refactored, the type of routing (`routing_method`) is now used to distinguish between different routing types for  `RiverFlowModel` and `OverlandFlowModel`.

## Checklist


Wflow/src/routing/surface/floodplain.jl
```diff
@@ -0,0 +1,364 @@
...
+abstract type AbstractFloodPlainVariables end
+
+"""
+    FloodPlainProfile
+
+Floodplain `storage` is a function of `depth` (flood depth intervals). Based on the
+cumulative floodplain `storage` a floodplain profile as a function of `flood_depth` is
+derived with floodplain area `a` (cumulative) and wetted perimeter radius `p` (cumulative).
+"""
+@with_kw struct FloodPlainProfile
```
SouthEndMusic, line 36: It's important to indicate for non-hydrologists on which part of the domain this concept lives. I've made an attempt at that in that in this PR: https://github.com/Deltares/Wflow.jl/pull/886

Wflow/src/routing/surface/floodplain.jl
```diff
@@ -0,0 +1,364 @@
...
+    depth::Vector{Float64}     # Flood depth [m]
+    storage::Matrix{Float64}   # Flood storage (cumulative) [m³]
+    width::Matrix{Float64}     # Flood width [m]
+    a::Matrix{Float64}         # Flow area (cumulative) [m²]
+    p::Matrix{Float64}         # Wetted perimeter (cumulative) [m]
```
SouthEndMusic: One character variable names are pretty bad, I've made an attempt at improving that in this PR: https://github.com/Deltares/Wflow.jl/pull/904

Wflow/src/routing/surface/floodplain.jl
```diff
@@ -0,0 +1,364 @@
...
+    segment_storage = zeros(n_depths, n)
+    width = zeros(n_depths, n)
+    width[1, :] = flow_width[1:n]
+
+    # determine flow area (a), width and wetted perimeter (p) FloodPlain
+    h = diff(flood_depths)
+    incorrect_vol = 0
+    riv_cells = 0
+    error_vol = 0
+    for i in 1:n
```
SouthEndMusic, line 86: Please use descriptive indices like in https://github.com/Deltares/Wflow.jl/pull/886

Wflow/src/routing/surface/floodplain.jl
```diff
@@ -0,0 +1,364 @@
...
+    )
+    return parameters
+end
+
+"Struct to store floodplain flow model variables on a staggered grid"
+@with_kw struct FloodPlainStaggeredVariables <: AbstractFloodPlainVariables
+    n::Int
+    n_edges::Int
+    # edge variables
+    a::Vector{Float64} = zeros(n_edges)    # flow area at edge [m²]
```
SouthEndMusic: Again please use descriptive field names

Wflow/src/routing/surface/floodplain.jl
```diff
@@ -0,0 +1,364 @@
...
+    for i in 1:nriv
+        i1, i2 = interpolation_indices(floodplain.variables.h[i], profile.depth)
+        a = compute_floodplain_flow_area(profile, floodplain.variables.h[i], i, i1, i2)
+        floodplain.variables.storage[i] = flow_length[i] * a
+    end
+    return nothing
+end
+
+"helper function to get interpolation indices"
+function interpolation_indices(x, v::AbstractVector)
```
SouthEndMusic, line 287: Interpolation is one of those things where I think Wflow should not re-invent the wheel. DataInteprolations.jl, as I have introduced in https://github.com/Deltares/Wflow.jl/pull/900, is highly optimized as it comes to e.g. interpolation interval finding. This counts for many of the functions below

Wflow/src/routing/surface/floodplain.jl
```diff
@@ -0,0 +1,364 @@
...
+function FloodPlainModel(
+    dataset::NCDataset,
+    config::Config,
+    domain::DomainRiver,
+    zb_floodplain::Vector{Float64},
+)
+    (; indices, local_drain_direction, graph) = domain.network
+    (; river_routing) = config.model
+    n = length(indices)
+    index_pit = findall(x -> x == 5, local_drain_direction)
```
SouthEndMusic: This is one of those things where defining a global constant like `PIT_DIRECTION = 5` would clarify things

Wflow/src/routing/surface/surface_flow.jl
```diff
@@ -0,0 +1,61 @@
...
+abstract type AbstractRiverFlowVariables end
+
+abstract type AbstractOverlandFlowBC end
+abstract type AbstractOverlandFlowParameters end
+abstract type AbstractOverlandFlowVariables end
+
+"Struct for storing river flow model boundary conditions"
+@with_kw struct RiverFlowBC{R <: Union{ReservoirModel, Nothing}}
+    n::Int
+    inwater::Vector{Float64} = zeros(n)                         # Lateral inflow [m³ s⁻¹]
```
SouthEndMusic: Minor nit: I think putting the docstrings above the fields is better, because it makes the text less wide and doesn't need fiddly alignment

Wflow/src/routing/surface/surface_kinwave.jl
```diff
@@ -1,49 +1,46 @@
...
-        mannings_n,
-        alpha_pow = Float64((2.0 / 3.0) * 0.6),
-    )
+function ManningFlowParameters(
+    mannings_n::Vector{Float64},
+    slope::Vector{Float64},
+    wetted_perimeter::Vector{Float64},
+)
+    beta = Float64(0.6)
+    hydraulic_radius_pow = Float64(2.0 / 3.0)
```
SouthEndMusic: Why add `Float64`?

Wflow/src/routing/surface/surface_kinwave.jl
```diff
@@ -351,10 +337,48 @@ function update_overland_flow_model!(
...
+            hydraulic_radius = flow_area/wetted_perimeter
+            flow_capacity[i] =
+                manning_flow(mannings_n[i], hydraulic_radius, slope[i], flow_area) * dt
+        else
+            flow_capacity[i] = 0.0
+        end
+    end
+    q .= accucapacityflux(storage, domain.network, flow_capacity) # q/dt
+    q_av .+= q
+    q ./= dt # m3/s
```
SouthEndMusic: Looks like this is going to need some work for https://github.com/Deltares/Wflow.jl/pull/735

Wflow/src/routing/surface/surface_process.jl
```diff
@@ -145,3 +145,8 @@ function local_inertial_flow(
 
     return q
 end
+
+function manning_flow(mannings_n, hydraulic_radius, slope, area)
+    q = 1.0 / mannings_n * pow(hydraulic_radius, 2.0 / 3.0) * sqrt(slope) * area
```
SouthEndMusic:
```suggestion
    q = cbrt(hydraulic_radius^2) * sqrt(slope) * area / mannings_n
```

Wflow/src/routing/surface/surface_staggered_scheme.jl
```diff
@@ -1,136 +1,166 @@
...
+    (; river_routing) = config.model
+    (; graph, indices, local_drain_direction, nodes_at_edge) = domain.network
+    (; reservoir_outlet) = domain.parameters
+
+    waterdepth_threshold, froude_limit = log_message_staggered_flow(config)
+
+    n = length(indices)
+    n_edges = ne(graph)
     active_index = findall(x -> x == 0, reservoir_outlet)
+    index_pit = findall(x -> x == 5, local_drain_direction)
```
SouthEndMusic: As before with global constant (no 'magic numbers' in the code, @evetion), can also be abbreviated to `==(PIT_DIRECTION)`

Wflow/src/routing/surface/surface_staggered_scheme.jl
```diff
@@ -1,136 +1,166 @@
...
-    error::Vector{Float64} = zeros(n)          # error storage [m³]
+    # edge variables
+    q::Vector{Float64} = zeros(n_edges)         # river discharge at edge (subgrid channel) [m³ s⁻¹]
+    q0::Vector{Float64} = zeros(n_edges)        # river discharge at edge (subgrid channel) at previous time step [m³ s⁻¹]
+    q_av::Vector{Float64}                       # average river channel (+ floodplain) discharge at edge [m³ s⁻¹] (model timestep Δt)
+    q_channel_av::Vector{Float64}               # average river channel discharge at edge [m³ s⁻¹] (for model timestep Δt)
+    zs_max::Vector{Float64} = zeros(n_edges)    # maximum water elevation at edge [m]
+    zs_src::Vector{Float64} = zeros(n_edges)    # water elevation of source node of edge [m]
+    zs_dst::Vector{Float64} = zeros(n_edges)    # water elevation of downstream node of edge [m]
+    hf::Vector{Float64} = zeros(n_edges)        # water depth at edge [m]
```
SouthEndMusic: Again, descriptive names

Wflow/src/routing/surface/surface_staggered_scheme.jl
```diff
@@ -816,12 +917,33 @@ function stable_timestep(
...
+    (; alpha_coefficient) = river_flow_model.timestepping
+    (; n, mannings_n) = river_flow_model.parameters
+    (; h) = river_flow_model.variables
+    (; flow_length, flow_width, slope) = parameters
+
+    beta = 5.0/3.0
+    @batch per = thread reduction = ((min, dt_min),) for i in 1:(n)
+        @fastmath @inbounds h_r =
+            (flow_width[i] * h[i]) / wetted_perimeter_channel(h[i], flow_width[i])
+        celerity = beta * 1.0/mannings_n[i] * pow(h_r, 2.0/3.0) * sqrt(slope[i])
```
SouthEndMusic: Looks like this does the same computation as the suggestion I did elsewhere, generally avoid `pow` for performance reasons.

Wflow/src/routing/surface/surface_staggered_scheme.jl
```diff
@@ -816,12 +917,33 @@ function stable_timestep(
...
+    river_flow_model::RiverFlowModel{<:ManningStaggered},
+    parameters::RiverParameters,
+)
+    dt_min = Inf
+    (; alpha_coefficient) = river_flow_model.timestepping
+    (; n, mannings_n) = river_flow_model.parameters
+    (; h) = river_flow_model.variables
+    (; flow_length, flow_width, slope) = parameters
+
+    beta = 5.0/3.0
```
SouthEndMusic: Make global constant?

Wflow/src/routing/surface/surface_staggered_scheme.jl
```diff
@@ -816,12 +917,33 @@ function stable_timestep(
...
+    @batch per = thread reduction = ((min, dt_min),) for i in 1:(n)
+        @fastmath @inbounds h_r =
+            (flow_width[i] * h[i]) / wetted_perimeter_channel(h[i], flow_width[i])
+        celerity = beta * 1.0/mannings_n[i] * pow(h_r, 2.0/3.0) * sqrt(slope[i])
+        dt = alpha_coefficient * flow_length[i] / celerity
+        dt_min = min(dt, dt_min)
+    end
+    dt_min = isinf(dt_min) ? 600.0 : dt_min
```
SouthEndMusic: magic number?

Wflow/src/routing/utils.jl
```diff
@@ -122,3 +122,37 @@ function flux_in!(flux_in, flux, network)
     end
     return nothing
 end
+
+function compute_value_at_edge(v, nodes_at_edge, n_edges, func::Function)
+    x = fill(Float64(0), n_edges)
```
SouthEndMusic:
```suggestion
    x = zeros(n_edges)
```

SouthEndMusic (commented):
I know this isn't ready for review yet, but I had a look and there were some notes I'd like to make regarding code quality and performance.

Also, I want to encourage getting into the habit of annotating the units of computations like in https://github.com/Deltares/Wflow.jl/pull/735.

Wflow/src/routing/surface/floodplain.jl
vers-w: I am more in favor of describing this (instead of renaming indices which is not very generic), made an attempt in https://github.com/Deltares/Wflow.jl/pull/870/commits/9c2c9f5ab618048c117cf957a3c92bf7acc6e3e7.

Wflow/src/routing/surface/floodplain.jl
vers-w: Yep, merged #904. That would also be my preference, make these changes while merging and not keeping track of all possible changes in other not ready to merge PR's ;-)

Wflow/src/routing/surface/floodplain.jl
vers-w: See my previous comment about this.

Wflow/src/routing/surface/floodplain.jl
vers-w: This code is not added as part of this PR (only moved to another file), so let's wait on the finalization of #886 :-).

Wflow/src/routing/surface/floodplain.jl
vers-w: See https://github.com/Deltares/Wflow.jl/pull/900#issuecomment-4437754318.

Wflow/src/routing/surface/surface_kinwave.jl
vers-w: Yes, #735 has been merged.

Wflow/src/routing/surface/surface_staggered_scheme.jl
vers-w: Merged #904

Wflow/src/routing/surface/floodplain.jl
vers-w: Good suggestion, made the following global constant: `LDD_PIT = 5`.

Wflow/src/routing/surface/surface_kinwave.jl
vers-w: Removed `Float64`.

Wflow/src/routing/surface/surface_staggered_scheme.jl
vers-w: Good point, removed `pow` from the computation.

Wflow/src/routing/surface/surface_staggered_scheme.jl
vers-w: Making use of `BETA_KINWAVE` now.

Wflow/src/routing/surface/surface_staggered_scheme.jl
vers-w: Fair point, removed these.

Wflow/src/routing/surface/floodplain.jl
```diff
@@ -0,0 +1,425 @@
...
+
+"Compute flood depth by interpolating flood storage `flood_storage` using flood depth intervals."
+function compute_flood_depth(
+    profile::FloodPlainProfile,
+    flood_storage::Float64,
+    flow_length::Float64,
+    i::Int,
+)
+    i1, i2 = interpolation_indices(flood_storage, @view profile.storage[:, i])
+    ΔA = (flood_storage - profile.storage[i1, i]) / flow_length
```
JoostBuitink: Is there a specific reason to use `Δ` here, and not `d` (like `dh`)? Maybe more a style choice, but would it perhaps be better to write it as `delta_h` and `delta_A`?

Wflow/src/routing/surface/surface_kinwave.jl
```diff
@@ -1,5 +1,8 @@
-"Struct for storing (shared) variables for river and overland flow models"
-@with_kw struct FlowVariables
+# constant in Manning's equation [-]
+const BETA_KINWAVE = 0.6
```
JoostBuitink, line 2: I notice that some constants are defined in Wflow.jl, while this one is defined here. Not sure about best practices, but do we want a single location where constants are stored?

docs/model_docs/routing/staggered-grid.qmd
```diff
@@ -1,8 +1,11 @@
 ---
-title: Local inertial
+title: Staggered grid
```
JoostBuitink, line 2: As also discussed, I am doubting a bit on the name for this page. It is indeed technically correct, just wonder if users get confused by this. I also don't really have a better suggestion, but wanted to note it anyway!

Wflow/src/routing/surface/surface_staggered_scheme.jl
```diff
@@ -379,92 +375,139 @@ function update_river_channel_flow!(
...
+        )
+        hydraulic_radius_at_edge = ifelse(
+            floodplain_v.water_depth_at_edge[i] > profile.depth[i1],
+            hydraulic_radius_at_edge,
+            0.0,
+        )
 
-        floodplain_v.q[i] = if floodplain_v.flow_area[i] > 1.0e-05
+        floodplain_v.q[i] = ifelse(
+            flow_area_at_edge > 1.0e-05,
```
JoostBuitink, line 499: should this 1.0e-05 be defined as a config value or something? it is essentially somewhat similar as the h_thresh, right? Note that it is also further down for the manning_flow option

JoostBuitink (commented):
Thanks for this work! I will still do a run with a different model to check the effects, will update the review in the PR once I have done that (hopefully this week!) Other than that, I only have a couple of small comments, see below

JoostBuitink:
When running a simulation for the Meuse, I noticed two things:
- it looks like the states are not picked up correctly (e.g. the `floodplain_water__instantaneous_volume_flow_rate` and `floodplain_water__depth` when running the local_inertial river routing + 1d floodplains
- I noticed numerical instability when using the `manning_staggered` routing option (see also the model with output in the postbox)

Wflow/src/routing/surface/floodplain.jl
vers-w: No specific reason to use `Δ`  here, good suggestion, changed it to `delta_`.

Wflow/src/routing/surface/surface_kinwave.jl
vers-w: Also some other constants are defined outside Wflow.jl. I think that's fine, as long as the file location of the constant makes sense.

docs/model_docs/routing/staggered-grid.qmd
vers-w: Yes, it is technically correct and also used throughout the code and I also don't have a better suggestion. I would suggest to stay with this term and find out if we get users feedback on this.

Wflow/src/routing/surface/surface_staggered_scheme.jl
vers-w: Please note that this value is not new, it has been part of the code since flood routing was added as a new feature. I would prefer to decide on this as part of a separate issue.

vers-w:
>     * I noticed numerical instability when using the `manning_staggered` routing option (see also the model with output in the postbox)

Thanks for testing this functionality thoroughly! Changing the estimation of water level and depth at the edge in commit https://github.com/Deltares/Wflow.jl/pull/870/commits/5faf1fd6d3048b42fd5b4e8fcc59b0ab27451b84 resolves this.

vers-w:
>     * it looks like the states are not picked up correctly (e.g. the `floodplain_water__instantaneous_volume_flow_rate` and `floodplain_water__depth` when running the local_inertial river routing + 1d floodplains

This has been fixed in commit https://github.com/Deltares/Wflow.jl/pull/870/commits/20ce151f0b9d98730e5a59239bbe6ae2e9cd69e8. There was also a state issue for the kinematic wave floodplain routing, this has been fixed in commit https://github.com/Deltares/Wflow.jl/pull/870/commits/7fe5253b3418082bee682c5feed42ca6d449796a.

JoostBuitink (approved):

PR #873, opened 2026-03-18
fix kvfrac name and update tests

JoostBuitink:
## Issue addressed
Fixes #<issue number>

## Explanation

## Checklist


SouthEndMusic (approved):
LGTM. Just for my understanding:
- This didn't gave errors because the parameter is optional and used the default, which is also why the test results changed slightly?
- If this was also a problem in HydroMT-Wflow, does that mean that a lot of models that users have made had the same problem?

JoostBuitink:
> * This didn't gave errors because the parameter is optional and used the default, which is also why the test results changed slightly?

Correct indeed! Wflow ran without picking up that layer (but using the default instead). Now the layer is properly picked up, and hence the test values needed changes

> * If this was also a problem in HydroMT-Wflow, does that mean that a lot of models that users have made had the same problem?

That is correct, and is now fixed (see https://github.com/Deltares/hydromt_wflow/pull/722). Although there are not a lot of people (as far as I know) that use the water demand feature, but they indeed have this same issue.

PR #890, opened 2026-04-28
Setup AGENTS.md

### LINKED ISSUE #837: Add copilot instruction file
For details see https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions.

We can update this file as we go. Some good things to start with are:
- Instructions how to run tests
- Instructions for code style (e.g. small number of function arguments). This could also be in the developer docs and have the copilot instructions refer to it

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/837


visr:
It is `.github/copilot-instructions.md` right, instead of `.github/github-instructions.md`.
I wonder if it is better to name it [`AGENTS.md`](https://agents.md/) and put it in the repo root, as that seems to be the closest to a standard there is, and would work for others not using Copilot (or GitHub).
Since this gets loaded into every session, adding too much here can also harm performance. I'd probably trim it down to the essential bullets that an agent needs to know to start working, and let it figure out the details by itself on demand. I also want to move in that direction for Ribasim: https://github.com/Deltares/Ribasim/issues/2923.

And then whenever you spot an agent making a repeated mistake, add a line on it.

AGENTS.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Architecture
+- Central type: `Model{R, L, M, T}` — routing, land model, mass balance, model type tag
+- **Land models** (`AbstractLandModel`): vertical per-cell fluxes. `LandHydrologySBM` (hydrology), `SoilLossModel` (sediment)
+- **Routing** (`Routing{O,R,S}`): `overland_flow`, `river_flow`, `subsurface_flow` — each concrete or null (`No*` type)
+- **Model type tags** (dispatch singletons): `SbmModel`, `SbmGwfModel`, `SedimentModel`
+- **Immutable struct updates**: use `@reset` from Accessors.jl
+
+## Simulation loop (`run_timestep!`)
+1. `advance!(clock)` 2. `load_dynamic_input!` 3. `storage_prev!` 4. `update_model!` 5. `compute_mass_balance!` 6. `write_output`
```
visr:
```suggestion
1. `advance!(clock)`
2. `load_dynamic_input!`
3. `storage_prev!`
4. `update_model!`
5. `compute_mass_balance!`
6. `write_output`
```

AGENTS.md
```diff
@@ -0,0 +1,89 @@
...
+- 3D (x,y,time) and 4D (x,y,layer,time) NetCDF variables; layered → SVector
+
+## Forcing
+- `AtmosphericForcing`: precipitation, potential_evaporation, temperature
+- Right-labeling: timestamp marks end of accumulation period
+- `load_dynamic_input!` = `update_forcing!` + `update_cyclic!`
+- Reservoir: precip/evap averaged over coverage cells, zeroed in land model
+
+## Testing
+- `TestItemRunner.jl` with `@testitem` (not `@testset`); unit tests prefixed `"unit: "`
```
visr, line 78: It helps to give concrete commands that it can use, see https://www.julia-vscode.org/docs/stable/userguide/testitems/#Filtering-support-in-TestItemRunner.jl.

visr (approved):
Looks good, some small comments.

PR #897, opened 2026-05-05
Metadata lookup performance improvement

SouthEndMusic:
I noticed that a part of the metadata lookup was slow (specifically the part where lenses are compared to strings), so I refactored it a bit.

### DIFF
--- a/Wflow/src/standard_name/standard_name_utils.jl
+++ b/Wflow/src/standard_name/standard_name_utils.jl
@@ -5,7 +5,7 @@ get_standard_name_map(::Type{<:SoilLossModel}) = sediment_standard_name_map
 get_standard_name_map(::Type{<:Domain}) = domain_standard_name_map
 get_standard_name_map(::Type{<:Routing}) = routing_standard_name_map
 
-const PARAMETER_TYPES = Union{Float64, Int, Bool, Nothing}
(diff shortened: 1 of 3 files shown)


Wflow/src/standard_name/standard_name_utils.jl
vers-w, line 84: Hm, I am wondering if we should support the lookup of metadata for output variables that are given as model path, especially if it is quite expensive. Maybe we could also opt that these variables are always written in SI unit?

SouthEndMusic: The performance problem is actually fixed in https://github.com/Deltares/Wflow.jl/pull/735, where a direct mapping to the output variables is created during initialization;

https://github.com/Deltares/Wflow.jl/blob/cf194931d549fe9c55bd5d18aed9a5d2e77f2b6a/Wflow/src/io.jl#L432-L474

on master this is not done for CSV and NetCDF scalar output, which I thought was a bit inconsistent. So if https://github.com/Deltares/Wflow.jl/pull/735 is merged first, this PR is not needed.

vers-w: OK, didn't know that, suggest to close this PR then.

SouthEndMusic:
Closed as not needed

PR #898, opened 2026-05-05
Kinematic wave performance improvement

### LINKED ISSUE #851: Small performance improvements to kinematic wave solve
Note that

- First note that

$$
q^{\beta - 1} = \frac{q^\beta}{q},
$$

so only $q^\beta$ has to be computed with the expensive power function and then $q^{\beta - 1}$ can be computed by this division.

- It looks like always $\beta = \frac{3}{5}$, so it is probably cheaper to compute this with the exponent being a `Rational`.

- It should be checked whether the initial estimate is already sufficient by swapping the computation and check parts in the iteration loop.

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/851

## Explanation
I went a different route than I suggested in the issue. The equation being solved per cell per timestep is of the form 

$$
aq + bq^\beta = C,
$$

where it is hardcoded that $\beta = \frac{3}{5}$. This means that via the substitution $u = q^\frac{1}{5}$ the above equation can be reformulated as

$$
au^5 + bu^3 = C;
$$

a polynomial equation which is much cheaper to evaluate. The solution is then found with [Halley's method](https://en.wikipedia.org/wiki/Halley%27s_method) instead of Newton's method for faster convergence. The resulting $u$ can then be used to compute both `q` and the `crossarea`.

Performance comparison on the test data:

```
alpha = 2.586
beta = 0.6
dt = 600.0
dx = 1061.375
q_in = 1.104e-6
q_prev = 0.0
q_lat = 1.142e-6


Original average kinematic_wave time: 225.440 ns
New implementation average time:       54.536 ns 
				                       ~4x speedup
```

Performance on the `sbm_piave` model:

```
Before:
- runtime: 19.2 seconds
- % total run time spend on `kinematic_wave` in `kinwave_river_update!`: 36%

After:
- runtime: 12.5 seconds
- % total run time spend on `kinematic_wave` in `kinwave_river_update!`: 18%
```

An open question is whether we still want to support the general $\beta$ case.

### DIFF
--- a/Wflow/src/routing/surface/surface_kinwave.jl
+++ b/Wflow/src/routing/surface/surface_kinwave.jl
@@ -13,7 +13,6 @@ end
 "Struct for storing Manning flow parameters"
 @with_kw struct ManningFlowParameters
     n::Int
-    beta::Float64                 # constant in Manning's equation [-]
     slope::Vector{Float64}        # Slope [m m⁻¹]
     mannings_n::Vector{Float64}   # Manning's roughness [s m⁻⅓]
     alpha_pow::Float64            # Used in the power part of alpha [-]
@@ -26,10 +25,9 @@ function ManningFlowParameters(mannings_n::Vector{Float64}, slope::Vector{Float6
     n = length(slope)
     parameters = ManningFlowParameters(;
         n,
-        beta = Float64(0.6),
         slope,
         mannings_n,
-        alpha_pow = Float64((2.0 / 3.0) * 0.6),
+        alpha_pow=Float64((2.0 / 3.0) * 0.6),
     )
     return parameters
 end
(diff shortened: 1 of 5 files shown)


vers-w:
Nice performance improvement! 
Just wondering, do you have any idea about the performance for very low flows compared to the previous implementation? With the previous implementation kinematic wave routing for very low flows is quite slow.

SouthEndMusic:
@vers-w can you provide some low flow inputs? And do we still want to support other values for $\beta$?

SouthEndMusic:
Low flow example;

```julia
q_in = 0.0          # headwater, no upstream inflow
q_prev = 0.0          # cold start or dry cell
q_lat = 1e-7         # nearly zero lateral inflow [m³/s/m]
alpha = 0.5          # moderate channel geometry
beta = 0.6
dt = 3600.0       # 1 hour [s]
dx = 250.0        # cell length [m]
```

before:
```julia
BenchmarkTools.Trial: 10000 samples with 202 evaluations per sample.
 Range (min … max):  386.634 ns …   7.890 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     435.644 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   460.559 ns ± 122.882 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
```

after:
```julia
BenchmarkTools.Trial: 10000 samples with 965 evaluations per sample.
 Range (min … max):  82.176 ns … 443.316 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     90.674 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   95.967 ns ±  18.767 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%
```

For this example the required number of iterations goes from 7 to 3.

Wflow/src/routing/surface/surface_kinwave.jl
```diff
@@ -518,6 +506,8 @@ function update_river_flow_model!(
     return nothing
 end
 
+const BETA_KINWAVE = 0.6
```
visr, line 510:
```suggestion
# constant in Manning's equation [-]
const BETA_KINWAVE = 0.6
```

visr (approved):
This looks like a very nice performance improvement, math to the rescue!

I think this is a better solution to the power problem than my earlier attempt in https://github.com/Deltares/Wflow.jl/pull/472.

One take-away from that approach for me was that it seems that the convergence criterium might be too tight in relation to the uncertainties in the parameters etc. Possibly that is another avenue for speedup that should be looked into.

PR #900, opened 2026-05-05
Use `DataInterpolations.jl` for reservoir interpolations

### LINKED ISSUE #772: Consider using `DataInterpolations.jl` for `interpolate_linear`
I haven't checked whether this linear interpolation which is only used in the reservoir module contributes significantly to the simulation time, but of so, the highly optimized interpolations in `DataInterpolations.jl` could offer a significant speedup.

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/772


vers-w:
Just a general comment: I think it makes sense to replace the small (and simple) custom interpolation function by `DataInterpolations.jl` if we plan to use other functionality of `DataInterpolations.jl` and/or if it results in a significantly speedup of simulations time. I don't think we plan to use other functionality any time soon, and did you check if there is a significant speedup? (what are your thought on this @JoostBuitink ?)

JoostBuitink:
> Just a general comment: I think it makes sense to replace the small (and simple) custom interpolation function by `DataInterpolations.jl` if we plan to use other functionality of `DataInterpolations.jl` and/or if it results in a significantly speedup of simulations time. I don't think we plan to use other functionality any time soon, and did you check if there is a significant speedup? (what are your thought on this @JoostBuitink ?)

Agree that it is a good idea to check how much (if any) effect this has on runtimes. I expect it will not be noticeable, but I could be wrong. I agree that we should only adopt this is if we see performance improvements and/or we expect we can use this functionality in different parts of the code base.

PR #916, opened 2026-05-20
Fix maximum subsurface flow for sub-daily timestep

vers-w:
This has already been fixed in commit https://github.com/Deltares/Wflow.jl/commit/b02a6a4c625940ffce52409cc63636a20337c0f4 (PR for Master branch).


JoostBuitink (approved):
thanks for applying this to the release branch!

PR #920, opened 2026-05-26
Remove check active floodplain indices `hf_index`

### LINKED ISSUE #921: Check active floodplain indexing `hf_index`
This check is based on the flow threshold of river nodes and should be based on the flow threshold of floodplain nodes (introduced when `LoopVectorization` dependency was removed).

vers-w:
## Issue addressed
Fixes #921


JoostBuitink (approved):
LGTM!

PR #929, opened 2026-06-04
Pin GitHub actions to a hash

visr:
Using `pinact`. For improved security, since e.g. v3 can change from under you and be compromised.

This means that the versions will update more frequently in the code, so to compensate we group them now in one PR, see the dependabot updated config.

This is the same approach as in Ribasim.


evetion (approved):
LGTM, was already considering this. This CI group setting overrides the group actions in the Github settings that I enabled on some repos? (grouped security updates in https://github.com/Deltares/Wflow.jl/settings/security_analysis).

visr:
> CI group setting overrides the group actions in the Github settings

Hmm yes it seems that way. I wasn't aware of that setting, but this is perhaps a bit more clear / explicit anyway.

PR #931, opened 2026-06-08
Formatting via pre-commit

### LINKED ISSUE #765: Format code via pre-commit

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/765

## Additional notes
Note that this only formats when committing; this doesn't affect format-on-save. Thus, it can happen that your format-on-save from the Julia VSCode plugin has different formatting rules than pre-commit, depending on whether there is a JuliaFormatter version difference.

### DIFF
--- a/.github/workflows/pre-commit_check.yml
+++ b/.github/workflows/pre-commit_check.yml
@@ -9,10 +9,10 @@ jobs:
     check:
         runs-on: ubuntu-latest
         steps:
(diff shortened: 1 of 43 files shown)


visr (approved):
Looks good!

PR #935, opened 2026-06-09
Shared module soil moisture

SouthEndMusic:
Linked PR in Ribasim: https://github.com/Deltares/Ribasim/pull/3103


vers-w:
As just briefly mentioned to @SouthEndMusic , we may want to opt for low level sharing of functions (e.g. for each irrigation node). Just a thought, an idea. This way there is no dependence of Wflow structs and there is more flexibility on the side of Ribasim. Not sure how doable this is.

SouthEndMusic:
Closing this as @HendrikKok and I came to the conclusion that for the prototype there are no changes required on the Wflow side. It makes more sense to build the 'shared library' (now more of a wrapper) in the Ribasim repository.

PR #939, opened 2026-06-15
Fix units rainfall erosion `eurosem`

### LINKED ISSUE #922: Check and fix units rainfall erosion sediment
1.  `soil_erosion__eurosem_exponent` has a default value of 2.0. However in the original equation (that uses a default value of 2.0) for soil detachment by raindrop impact this parameter has unit $\mathrm{mm}^{ -1}$ (multiplied by water depth [mm]), and in Wflow the unit of water depth is in [m]. This needs fixing in the `release/v1.0` branch and will be fixed in https://github.com/Deltares/Wflow.jl/pull/735.

2. The multiplication with 0.001 below in the `function rainfall_erosion_eurosem` is not required based on the units:   
    ```
   #Total kinetic energy by rainfall [J/m2]
    ketot = (rddir * kedir + rdleaf * keleaf) * 0.001
    ```
    This also needs fixing in the `release/v1.0` branch and will be fixed in https://github.com/Deltares/Wflow.jl/pull/735.

3. Based on the first point, the soil erosion for EUROSEM is now much higher compared to ANSWERS (`function rainfall_erosion_answers`). Need to check if units for the ANSWERS parameters are correct, especially the `usle_k` parameter that is now provided in metric units.

vers-w:
## Issue addressed
Fixes #922

### DIFF
--- a/Wflow/src/sediment/erosion/erosion_process.jl
+++ b/Wflow/src/sediment/erosion/erosion_process.jl
@@ -56,7 +56,7 @@ function rainfall_erosion_eurosem(
     rddir = max(rdtot - rdleaf - interception, 0.0) #throughfall
 
     #Total kinetic energy by rainfall [J/m2]
-    ketot = (rddir * kedir + rdleaf * keleaf) * 0.001
+    ketot = (rddir * kedir + rdleaf * keleaf)
     # Rainfall / splash erosion [g/m2]
     rainfall_erosion = soil_detachability * ketot * exp(-eurosem_exponent * waterlevel)
     rainfall_erosion *= area * 1e-6 # ton/cell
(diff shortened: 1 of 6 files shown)


JoostBuitink (approved):
Looks good! Since this issue will close #922, are we now also sure on the units of ANSWERS rainfall erosion (given the large difference between the two methods)? Or should we create a separate issue for that?

vers-w:
> Looks good! Since this issue will close #922, are we now also sure on the units of ANSWERS rainfall erosion (given the large difference between the two methods)? Or should we create a separate issue for that?

No, we are completely sure on the ANSWERS parameter units. I did create a separate issue for this #942.

PR #943, opened 2026-06-16
Avoid `NaN`/`Inf` values floodplain hydraulic radius

### LINKED ISSUE #941: Avoid `NaN`/`Inf` values floodplain hydraulic radius
By removing the check for active floodplain edges (based on the water depth threshold `h_thresh`) the hydraulic radius can have a `NaN` or `Inf` value because of a zero wetted perimeter. This needs to be checked/avoided. Additionally, we do (most probably) not need to save these values each sub timestep as these are used only internally (per function) and are not part of model output.

vers-w:
## Issue addressed
Fixes #941 

## Explanation
Floodplain hydraulic radius as part of local inertial routing could have `NaN` or `Inf` values. Additionally, the flow area and hydraulic radius of local inertial routing (river and floodplain) are saved for each sub timestep and are used only internally (per function) and are not part of model output. Removed flow area and hydraulic radius fields of local inertial structs and added check for hydraulic radius floodplain.

### DIFF
--- a/Wflow/src/routing/surface_local_inertial.jl
+++ b/Wflow/src/routing/surface_local_inertial.jl
@@ -281,8 +281,9 @@ function local_inertial_river_update!(
         river_v.zs_max[i] = max(river_v.zs_src[i], river_v.zs_dst[i])
         river_v.hf[i] = (river_v.zs_max[i] - river_p.zb_max[i])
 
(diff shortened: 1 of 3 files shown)


JoostBuitink (approved):
LGTM!

PR #950, opened 2026-06-22
Add wflow version field to TOML

### LINKED ISSUE #947: Add version field to the toml file
### Feature type

None

### Improvement Description

Add support for a field in the toml (at top level) that reads the `wflow_version` (see https://github.com/Deltares/hydromt_wflow/issues/764). I think for now we leave any further checking on this version to a later issue, as we need to come up with a (design) plan

_No response_

### Additional Context

_No response_

vers-w:
## Issue addressed
Fixes #947

### DIFF
--- a/Wflow/test/sbm_config.toml
+++ b/Wflow/test/sbm_config.toml
@@ -5,6 +5,7 @@
 
 dir_input = "data/input"
 dir_output = "data/output"
+wflow_version = "1.1"
 
 [logging]
 loglevel = "info"
--- a/docs/user_guide/toml_file.qmd
+++ b/docs/user_guide/toml_file.qmd
@@ -10,14 +10,18 @@ structured in several sections, which are explained below.
 ## General
 The file paths provided in this file are relative to the location of the TOML file, or to
 `dir_input` and `dir_output` if they are specified. To indicate that a wflow model [runs
-from Delft-FEWS](fews.qmd), the `fews_run__flag` setting needs to be set at `true`.
+from Delft-FEWS](fews.qmd), the `fews_run__flag` setting needs to be set at `true`. The
+`wflow_version` setting represents the Wflow.jl version compatibility of the config .toml
+file. This setting is used by the wflow plugin
+[hydroMT-wflow](https://github.com/Deltares/hydromt_wflow) for incremental Wflow.jl version
+upgrading.
 
 ```toml
 dir_input = "data/input"      # optional, default is the path of the TOML
 dir_output = "data/output"    # optional, default is the path of the TOML
-fews_run__flag = false              # optional, default value is false
+fews_run__flag = false        # optional, default value is false
+wflow_version = "1.1"         # optional, default value is `nothing`
 ```
-
 
 ## Time info
 Time information is optional. When omitted, wflow will perform computations for each
--- a/Wflow/src/config_init.jl
+++ b/Wflow/src/config_init.jl
@@ -13,6 +13,8 @@ function convert_value(::Type{T}, value::Any) where {T}
         T_ = get_something_type(T)
         if T_ <: AbstractConfigSection
             init_config_section(T_, value)
+        elseif T_ == VersionNumber
+            VersionNumber(value)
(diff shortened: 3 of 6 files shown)


docs/user_guide/toml_file.qmd
```diff
@@ -10,15 +10,19 @@ structured in several sections, which are explained below.
...
+file. This setting is used by the wflow plugin
+[hydroMT-wflow](https://github.com/Deltares/hydromt_wflow) for incremental Wflow.jl version
+upgrading.
 
 ```toml
 dir_input = "data/input"      # optional, default is the path of the TOML
 dir_output = "data/output"    # optional, default is the path of the TOML
-fews_run__flag = false              # optional, default value is false
+fews_run__flag = false        # optional, default value is false
+wflow_version = "1.1"         # optional, default value is `nothing`
```
HendrikKok, line 23: Better use the expected version number format like 1.1.0 right?

Wflow/test/sbm_config.toml
```diff
@@ -5,6 +5,7 @@
 
 dir_input = "data/input"
 dir_output = "data/output"
+wflow_version = "1.1"
```
HendrikKok, line 8: 1.1.0?

HendrikKok (changes requested):

HendrikKok (changes requested):
we should rerun the failed CI job

docs/user_guide/toml_file.qmd
vers-w: Yes, the use of `1.1` is based on the [hydromt_wflow](https://github.com/Deltares/hydromt_wflow/blob/feat/update-config-params/hydromt_wflow/data/wflow_sbm/wflow_sbm.toml#L2) implementation.

And for the final version number (and further checking) in Wflow it is not important:

```
julia> VersionNumber("1.1")
v"1.1.0"
```

HendrikKok: Yes, but what does this version represent exactly: the HydroMT-Wflow plugin version, the Wflow.jl version, or the TOML/config schema version?

If it represents the Wflow.jl version, I would suggest enforcing the full semantic version, so there is no ambiguity about which version was and should be used to run the simulation. 

Otherwise, if this is intended to represent the TOML/config schema version, perhaps the field name could be made more explicit?

vers-w: It represents the Wflow.jl version. I would say this is up to the user. In Wflow internally `1.1` (or `1.1.0`) is converted to `v"1.1.0"`. @JoostBuitink : what are your thoughts on this?

JoostBuitink: Maybe you discussed this already, but the two main goals (as I understand it) of this version number is (1) ultimately to be able to flag whether a TOML/model configuration is suitable for the current application (for Wflow.jl) and (2) to be able to understand how older model configurations need to be updated/converted to match the latest version (done in hydromt_wflow). 

With our versioning and release structure, the patch version should not matter and should work with and 1.0.x and 1.1.x release. For minor versions, we already know that some small changes are needed in the model configuration (TOML only I believe), so I think that it why we only need to flag the major and minor versions here. 

I think it will get confusing to include the patch version in this number, because you might give the impression that a v1.0.0 model configuration would not be compatible with v1.0.3. 

In a later PR/stage, we should decide how we flag this within Wflow (warning/error/etc).

HendrikKok: Thanks, that makes sense. My comment was mainly from a reproducibility perspective: in that case, the exact Wflow.jl version used for the simulation should be recorded. However, that does not seem to be the primary purpose of wflow_version here.

HendrikKok (approved):

docs/user_guide/toml_file.qmd
JoostBuitink: That is indeed a valid point. We currently provide the wflow version in the logging that is produced with the run, do you think that would be sufficient?

PR #970, opened 2026-07-02
Remove redundant input parameter `specific_yield`

### LINKED ISSUE #967: subsurface_water__specific_yield seems to be redundant
### Wflow version checks

- [x] I have checked that the issue still exists on the latest versions of the docs on `master` [here](https://github.com/Deltares/Wflow.jl)

### Kind of issue

Docs are wrong

### Location of the documentation

Dear Wflow developper,

In the [parameters_routing](https://deltares.github.io/Wflow.jl/dev/model_docs/parameters_routing.html) the parameter `subsurface_water__specific_yield` seems to be no longer in use as it is computed in Wflow model and does not offer any more the option to provide your own parameter.

```julia
specific_yield = @. lower_bound_drainable_porosity(soil.parameters.theta_s, soil.parameters.theta_fc)
```

With kindness,
Joseph A.P. Pollacco

### Documentation problem

If I am correct than `subsurface_water__specific_yield` could be removed from  [parameters_routing](https://deltares.github.io/Wflow.jl/dev/model_docs/parameters_routing.html)

May I sugget in the excellent documentation  that in the **Model and parameter section** there is a seperate section for **Groundwater**.

### Suggested fix for documentation

Same as above

vers-w:
## Issue addressed
Fixes #967

### DIFF
--- a/Wflow/src/standard_name/standard_name_routing.jl
+++ b/Wflow/src/standard_name/standard_name_routing.jl
@@ -494,11 +494,6 @@ const routing_standard_name_map = OrderedDict{String, ParameterMetadata}(
             description = "Horizontal conductivity",
             tags = [:groundwater_unconfined_aquifer_input],
         ),
-    "subsurface_water__specific_yield" => ParameterMetadata(;
-        lens = @optic(_.routing.subsurface_flow.parameters.specific_yield),
-        description = "Specific yield",
-        tags = [:groundwater_unconfined_aquifer_input],
-    ),
     "subsurface__horizontal_saturated_hydraulic_conductivity_scale_parameter" =>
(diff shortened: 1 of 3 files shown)


JoostBuitink (changes requested):
I think here is still a reference to the specific_yield as well:

https://github.com/Deltares/Wflow.jl/blob/37b0320b4c0fea6a38d68b6757a5705f37dba789/docs/model_docs/routing/gwf.qmd#L146

vers-w:
> I think here is still a reference to the specific_yield as well:

Thanks for spotting this! Removed the reference and added a small note about specific yield below the example.

JoostBuitink (approved):

PR #972, opened 2026-07-02
Fix model initialization issue

SouthEndMusic:
This fixes a problem where a model from @JoostBuitink could not initialize


JoostBuitink (approved):
LGTM, thanks a lot for this quick fix! The model does indeed run with these changes now!

PR #978, opened 2026-07-07
Use explicit imports

### LINKED ISSUE #375: Specify which functions
### Feature type

### Improvement Description

Right now, we are using `using NCDatasts` for example in our code. For some functions, it is not fully clear where the functions are imported from (either Wflow.jl functions, or some other imported functions). 

_No response_

### Additional Context

_No response_

visr:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/375
Also removes Roots.jl from the manifest. Not sure why it was there since it wasn't used. I noticed TeamCity failing on this false dependency.

## Explanation
I ran [ExplicitImports](https://juliatesting.github.io/ExplicitImports.jl/stable/). Not all findings from that package are addressed here. For instance use of non-public names and importing from non-owner modules are not fixed here.

### DIFF
--- a/Wflow/src/Wflow.jl
+++ b/Wflow/src/Wflow.jl
@@ -347,7 +347,7 @@ function run!(model::Model; close_files = true)
     # option to support running function twice without re-initializing
     # and thus opening the netCDF files
     if close_files
-        close_files(model; delete_output = false)
+        Wflow.close_files(model; delete_output = false)
     end
 
     # Write config to dir_output, to archive what settings were used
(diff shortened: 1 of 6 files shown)


Wflow/src/Wflow.jl
vers-w, line 349: Maybe cleaner to rename the keyword argument `close_files`?

vers-w (approved):
LGTM, just one small comment.

Wflow/src/Wflow.jl
visr: Agree it would be cleaner, but it would break API for minimal gain, so I don't think it is worth it?

vers-w: Ah yes, good point, I agree that renaming is then not worth it.

Issue #864, opened 2026-03-16
Add test writing output from `SVector` elements (probably broken)


SouthEndMusic:
The code here looks broken (`nc`, `nt` do not exist) but this doesn't fail any current test:

https://github.com/Deltares/Wflow.jl/blob/fe9c2cb26caefbb846e62f95d2597c7fc7652371/Wflow/src/io.jl#L864-L875

vers-w:
This indeed seems broken, needs fixing (first) in v1.0 branch.

PR #983, opened 2026-07-09
Fix unit conversion after reduction

### LINKED ISSUE #974: `from_SI` unit conversion not working when writing aggregated output
### Wflow version checks

### Reproducible Example

When calculating the (e.g.) mean of a certain variable over a subcatchment, using something like the toml settings from below, the units are not correctly converted back to the expected units. In stead, the units remain in the internal SI units. 

```toml
[output.netcdf_scalar]
path = "output_scalar.nc"

[[output.netcdf_scalar.variable]]
name = "swe"
map = "subcatchment_camels_snow"
parameter = "snowpack__leq_depth"
reducer = "mean"

```

### Current behaviour

It is caused by these two functions, where the "fallback" function is dispatched when using a reducer.

https://github.com/Deltares/Wflow.jl/blob/3e751caa8934834187f8b0f689640c82de53c1cc/Wflow/src/units.jl#L303-L313

### Desired behaviour

Correctly convert units from_SI to their expected units.

### Additional Context

I have not fully checked this, but we should ensure that this unit conversion works for all types of output (csv, netcdf_scalar, netcdf_grid), and for all types of output writing (with or without reducer options, direct index/coordinate values, gauges layer, etc)

SouthEndMusic:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/974

## Explanation
Some reducers created `Generator` objects instead of a `Float64` or `Vector{Float64}`, which then dispatched to the fallback method of `from_SI` as @JoostBuitink mentioned. 

@JoostBuitink regarding the testing you mentioned: can you make an issue with an overview of what should be tested and ideally also which test models can be used for it? See also https://github.com/Deltares/Wflow.jl/issues/864.

### DIFF
--- a/Wflow/src/io.jl
+++ b/Wflow/src/io.jl
@@ -840,12 +840,12 @@ function write_netcdf_timestep(model::AbstractModel, writer::NCWriter{<:NCDatase
         elseif elemtype <: SVector
             # check if an extra dimension and index is specified in the TOML file
             if haskey(output_dataset, extra_dim.name)
-                v = from_SI(reducer_(getindex.(A, layer)), unit; dt_val)
+                v = from_SI(reducer_(getindex.(vector, layer)), unit; dt_val)
                 output_dataset[name][:, time_index] .= v
             else
-                nlayer = length(first(A))
+                nlayer = length(first(vector))
                 for i in 1:nlayer
-                    v = from_SI(reducer_(getindex.(A, layer)), unit; dt_val)
+                    v = from_SI(reducer_(getindex.(vector, layer)), unit; dt_val)
                     output_dataset[name][:, i, time_index] .= v
                 end
             end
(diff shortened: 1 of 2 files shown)


JoostBuitink (changes requested):
This did not fix the issue, as the function is still dispatching on https://github.com/Deltares/Wflow.jl/blob/af8225f6e846f4e1cb78d5fed282654c025a195c/Wflow/src/units.jl#L313

This is because of the following:
``` julia
typeof(reducer_(vector)) = Vector{Float64} 
```

while `from_SI` expects an `AbstractFloat`.

I'll create a separate issue to describe how to add this functionality to the tests

SouthEndMusic:
I added an `AbstractArray` method for `from_SI`, based on your comment that should fix it

JoostBuitink:
Thanks that did indeed solve this!

JoostBuitink (approved):

PR #987, opened 2026-07-13
fix negative snow

### LINKED ISSUE #985: Snow can become negative
### Wflow version checks

### Reproducible Example

Most likely caused by a floating point error, the snow_storage can become negative. Since there are no checks in the snow code to account for this, very unrealistic behaviour can occur.

While the code itself has not changed over the last time, everything is now computed at SI units (meaning potentially very small values), meaning we are now much more sensitive to floating point errors

### Current behaviour

I think it is caused by a floating point error in this line: 
https://github.com/Deltares/Wflow.jl/blob/a09c4aed698dc37787326341cd826e74d7ca2b21/Wflow/src/snow/snow_process.jl#L40-L41

Here it is assumed that `snow_storage/dt` should be equal to `snow_melt` in the case, but due to floating point errors, the does not have to be the case. Negative `snow_storage` values can sometimes lead to `-inf` amount of snow, caused by this logic: https://github.com/Deltares/Wflow.jl/blob/a09c4aed698dc37787326341cd826e74d7ca2b21/Wflow/src/snow/snow_process.jl#L56-L60

### Desired behaviour

Limit snow_storage (and perhaps snow_water as well) to no become negative:

```julia
    snow_storage = max(snow_storage, 0.0)
    snow_water = max(snow_water, 0.0)
```

### Additional Context

_No response_

JoostBuitink:
## Issue addressed
Fixes #985 

## Explanation

## Checklist
- [x] Tests & prek hooks pass

### DIFF
--- a/docs/changelog.qmd
+++ b/docs/changelog.qmd
@@ -35,7 +35,6 @@ subsurface_water__instantaneous_volume_flow_rate = "ssf"
 ### Fixed
 - Fixed a bug where the maximum subsurface flow was not adjusted for subdaily timesteps
   (only when using `layered` or `layered_exponential` profiles)
-- Fixed an issue where the snow storage could become negative.
 
 ### Changed
 - All internal computations now use standardized SI units (m, s, kg, J, K) consistently
--- a/docs/changelog.qmd
+++ b/docs/changelog.qmd
@@ -35,6 +35,7 @@ subsurface_water__instantaneous_volume_flow_rate = "ssf"
 ### Fixed
 - Fixed a bug where the maximum subsurface flow was not adjusted for subdaily timesteps
   (only when using `layered` or `layered_exponential` profiles)
+- Fixed an issue where the snow storage could become negative.
 
 ### Changed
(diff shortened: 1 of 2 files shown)


docs/changelog.qmd
```diff
@@ -35,6 +35,7 @@ subsurface_water__instantaneous_volume_flow_rate = "ssf"
 ### Fixed
 - Fixed a bug where the maximum subsurface flow was not adjusted for subdaily timesteps
   (only when using `layered` or `layered_exponential` profiles)
+- Fixed an issue where the snow storage could become negative.
```
vers-w: Should this also be fixed for v1.0? If not, I would suggest to leave this out.

vers-w (commented):
LGTM, just one small comment.

docs/changelog.qmd
JoostBuitink: No in v1.0 the amount of melt (and refreezing) is directly limited by the storages themselves, so without any `dt` corrections. Indeed, then this message can be removed, thanks for spotting this!

vers-w (approved):

Issue #984, opened 2026-07-09
TeamCity installer build fails by missing wix installation


deltamarnix:
It seems that teamcity is failing during the creation of the wix installer. I checked with CoPilot and got the following reponse. I think it makes sense that something is wrong in the import of the WixTargetPath. I think it is a matter of changing the paths in the import to make it work.

Another option to me could be that the pipeline is trying to run the wix build step on the signing agent. And that is not how it's supposed to work. First build the installer on a windows agent, and then send that artifact through the signing agent.

But I don't have enough internal knowledge of the pipeline system now, so I leave it up to the development team.

> The error **MSB4057: The target "Build" does not exist in the project** in a WiX `.wixproj` file almost always comes down to **WiX targets not being imported** — MSBuild has nothing to build because the WiX build logic was never loaded.
>  
> Looking at your `.wixproj`, the targets are imported here:
>  
> ```xml name=Application.Setup.wixproj url=https://github.com/Deltares/Wflow.jl/blob/master/build/wflow_cli/Setup/Installer/Application.Setup.wixproj
> <Import Project="$(WixTargetsPath)" Condition=" '$(WixTargetsPath)' != '' " />
> <Import Project="$(MSBuildExtensionsPath32)\Microsoft\WiX\v3.x\Wix.targets"
>         Condition=" '$(WixTargetsPath)' == '' AND Exists('$(MSBuildExtensionsPath32)\Microsoft\WiX\v3.x\Wix.targets') " />
> ```
>  
> Both imports are **conditional** — if neither condition is met, **no targets are imported at all**, and MSBuild sees an empty project with no `Build` target.
>  
> ---
>  
> ## Why this broke after switching to an Azure signing agent
>  
> On a **standard developer machine**, WiX installs itself into `$(MSBuildExtensionsPath32)\Microsoft\WiX\v3.x\`, so the second `Import` works. On a **fresh Azure/CI agent**, WiX is typically **not installed** into that path, so:
>  
> - `$(WixTargetsPath)` is empty (not set)
> - `$(MSBuildExtensionsPath32)\Microsoft\WiX\v3.x\Wix.targets` doesn't exist
>  
> Both conditions fail → no targets → `MSB4057`.
>  
> ---
>  
> ## Fixes
>  
> ### Option 1 — Install WiX on the agent (simplest)
> Make sure WiX v3 is installed on the Azure agent, either via a pipeline step or a pre-baked agent image:
> ```
> choco install wixtoolset
> ```
[...]

PR #993, opened 2026-07-14
Add manifest for MSIX installer

visr:
Part of https://github.com/Deltares/Wflow.jl/issues/984

Rather than fixing the WiX MSI installer, we can move to MSIX installers like Ribasim (see original PR https://github.com/Deltares/Ribasim/pull/2710). This is also preferred on the tech radar.

This PR adds the manifest for the installer, and the required logos at two sizes, which are just resampled logos from the `docs/_static` folder.

The content of the manifest is the same as that of Ribasim, only updating the names where needed.

After this is merged, I can try out the new TeamCity job that creates the MSIX from the ZIP. When that works, another job needs to sign the MSIX on the signing agent.

This leaves the WiX files in `Wflow/build/wflow_cli/Setup` for now. If all is working well we can update the docs and remove these files and the TeamCity job.


build/create_binaries/AppxManifest.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Package xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+  xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+  xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5"
+  xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities">
+  <Identity
+    Name="Deltares.Wflow"
+    Version="1.1.0.1"
```
visr, line 8: Note this is a new version number that we have to keep updated. It needs to be 4 parts. This can be seen as a kind of build number. There is no separate tag for pre-releases, so in Ribasim we just start from 1 and keep increasing if there are pre-releases. Without that we can keep it at 1.

https://ribasim.org/dev/release.html

> The Identity Version in build/AppxManifest.xml consists of four numbers. The fourth number should be reset to 1 when the 3-part version number changes. When the 3-part version number does not change, increase it by one. That means for the example of the two pre-releases of 2023.1.0 and its release, these should get 1, 2 and 3 respectively. If the fourth number does not increase, users cannot update existing installations, but have to uninstall first.

vers-w: Is is also possible to keep the same version numbering as we already use for Wflow? We have either a release number (`1.1.0`), and in between we use `-dev`.

visr: No unfortunately not, it needs to be 4 integers.

vers-w (approved):

PR #994, opened 2026-07-14
Relocate test data to bucket

### LINKED ISSUE #376: Improve storing and downloading of example models
### Feature type

### Improvement Description

Currenlty, the example models (used for testing and linked in the docs to run example simulations) are stored at this repo: https://github.com/visr/wflow-artifacts. Different releases on this repo contain different versions of the input data. It would be good to improve this, to e.g. link all required input files to the wflow release versions, and perhaps move this to a Deltares repo?

_No response_

### Additional Context

_No response_

### LINKED ISSUE #723: Move the test data to the MinIO bucket
A free Wflow MinIO bucket was created with a capacity of 500 GB. We have some decisions to make for how to use it.

- The test data is now scattered over several releases in a @visr repo: https://github.com/visr/wflow-artifacts/releases. Do we want to retain this structure, making a folder in the bucket for each release? Or do we want to clean it up a bit? I would like to end up with the bucket as the single source for this data, and I think @JoostBuitink and @vers-w mentioned that the older versions should still be available.
- In https://github.com/Deltares/Wflow.jl/pull/706 I created a pixi task to download the test data from the repo, which can be updated to download the data from the bucket (for inspiration see https://github.com/Deltares/Ribasim/blob/main/utils/s3_download.py). We need to make sure that locally the data gets updated by the pixi task if new versions were made, so the pixi task needs to know whether the local data is outdated. However, it would not be nice if the local file path depends on the data version, because then the paths in the tests have to be updated for each new version of the data. @visr suggested adding a file of hashes per file in the test data

rleander73:
## Issue addressed
Fixes https://github.com/Deltares/Wflow.jl/issues/723
Fixes https://github.com/Deltares/Wflow.jl/issues/376

## Explanation

## Checklist
- [ ] Tests & prek hooks pass


Wflow/test/data/input.dvc
visr, line 1: Did you intent to bundle the whole input directory together? I think that means that if one file changes, everything gets re-downloaded. Not too bad right now at 300MB, but this could get annoying. We can also do it per file. Or perhaps per model, but then we should probably bundle the models in subdirectories.

visr (approved):

utils/download_test_data.jl
```diff
@@ -1,63 +0,0 @@
-using Downloads
-
-const source_url = "https://github.com/visr/wflow-artifacts/releases/download"
-const version = v"1.0.0"
```
vers-w, line 4: What about providing the version of the test data (now equal to Wflow version (major and minor))?

Wflow/test/data/input.dvc
vers-w: Right now we bundle all files (version): https://github.com/visr/wflow-artifacts/releases/tag/v1.0.0.

visr: Yes I think it is fine like this. If it starts to get annoying because people are changing different files in different branches, we can still decide to version them separately.
And it doesn't seem to be true that everything gets redownloaded.

visr: Now I see why I was wrong. DVC doesn't track the dir as one blob, it just stores a JSON with pointers to the actual files, see this `eac653f3009e1f9babcae88c07ee56.dir` for instance:

```json
[
   {
      "md5":"76afca075bef6e20f48eb11dd585d7a4",
      "relpath":"hydrotype.dbf"
   },
   {
      "md5":"cb3e648c38fd75188df5ea80c1924386",
      "relpath":"hydrotype.prj"
   },
   {
      "md5":"b5ed91a25adf2925fc573061e948d6ff",
      "relpath":"hydrotype.shp"
   },
   {
      "md5":"2806f374363203bf1e88ce3b3ee238f2",
      "relpath":"hydrotype.shx"
   },
   {
      "md5":"399c079845032dfa1749335c619725e3",
      "relpath":"vdGaast_water_depth.csv"
   }
]
```

visr:
Was it considered that anonymous access should be possible?

If I understand correctly, as it is right now people can only get the test data if they have access keys. That means that external people cannot get the data unless they have such a key, and since secrets are or shared with PRs from forks, CI will also not run from forks. That is different from the current situation, where anybody can get the test data without authentication.

We should be able to support anonymous read access with [`allow_anonymous_login = true`](https://doc.dvc.org/user-guide/data-management/remote-storage/amazon-s3) on the DVC side, and similar configuration in MinIO. A separate DVC remote with write access via access keys could be used for maintainers to update the test data when needed.

visr (changes requested):
I discussed the anonymous access (https://github.com/Deltares/Wflow.jl/pull/994#issuecomment-5089495723) with @JoostBuitink, and we think it is important to address this within this PR as well. That probably should've come up earlier, sorry for adding that requirement now.

utils/download_test_data.jl
visr: Not sure exactly what you mean. In a way it is now versioned itself, with a hash tracked in git.

visr (approved):

PR #996, opened 2026-07-14
Update docs for MSIX, remove WiX installer

### LINKED ISSUE #984: TeamCity installer build fails by missing wix installation

visr:
## Issue addressed
Fixes #984

## Explanation
Continues from https://github.com/Deltares/Wflow.jl/pull/993. Since the MSIX can be made and signed in TeamCity now, we can remove the WiX files from the repository, and update the documentation in the second commit.


JoostBuitink:
I downloaded the msix installer from TeamCity and the install works smoothly. 

However, it is completely unclear to me (and I think impossible to change) where the wflow_cli gets installed. Also, in the previous installer, there was a Deltares T&C that people had to accept before installing, which is no longer part of the new installer right? Not sure if that is a change in Deltares policy, but I wanted to share and check this anyway.

Ah I found the path of the CLI: "C:\Users\USERNAME\AppData\Local\Microsoft\WindowsApps\wflow_cli.exe"

JoostBuitink:
In addition, I think we are no longer sharing the licenses of all dependencies (see below), that we used to do and store at the same location as the CLI

image

visr:
Already discussed in person, but to be clear. Indeed you cannot chose the installation location anymore, it is installed in a hidden directory. It will put `wflow_cli` in the path, so you don't need the full path. Only one version can be installed at the same time via MSIX, but you can switch. You can find it and the `dep_licenses` etc. will all still be there. The T&C are no longer part of the installer. We need to link them at the download location ("by downloading you agree to").

Whilst MSIX is convenient for the average user, we can consider also offering a ZIP download. That makes more sense e.g. if you have multiple versions that you use at the same time.

JoostBuitink:
Thanks for also writing it here! 

Just one addition (not sure if I properly mention it), but for the QGIS plugin that is developed together with Witteveen&Bos, they recently added a check on the version of the wflow_cli.exe, to see if that version is compatible with the plugin. I believe they currently get that information from the Manifest.toml file that is packaged in the CLI as well. Perhaps adding support for something like `wflow_cli.exe --version` would be important to have then as well, and I hope that they can perform a similar check that way.

visr:
Ah yes I agree, that reminds me I was drafting an issue about libwflow / CLI.

docs/getting_started/install.qmd
```diff
@@ -108,14 +108,7 @@ start a Julia session.
...
 [download.deltares.nl](https://download.deltares.nl/en/wflow/), and are currently available
-for Windows. Download and install the `.msi` file. After installation, you will see two
-folders in the installation directory. Only the `bin/wflow_cli` is used. The `artifacts`
-folder contains binary dependencies such as netCDF.
-
-```bash
-artifacts\
-bin\wflow_cli
-```
+for Windows. Download and install the `.msix` file. This will put `wflow_cli` on your path.
```
visr:
```suggestion
for Windows. Download and install the `.msix` file.
```

vers-w (approved):

PR #1013, opened 2026-07-23
Set Docker Julia version to 1.12.6 also

visr:
To be in line with the rest of the repo.


visr:
Already backported, otherwise we cannot build Docker (it was still on julia 1.9 there)

PR #1017, opened 2026-07-27
Put Julia directly in the Pixi environment

visr:
We had juliaup from conda-forge in our Pixi workspace. That did not directly install the julia version we needed in the environment, it just installed juliaup, and we had to remember to run `pixi run install-julia` which installed the right julia version and set the Wflow directory as a juliaup override. This is easy to forget and feels a bit two-staged.

https://prefix.dev/ has the julia-forge channel, which contains the same julia binaries that juliaup serves: https://prefix.dev/channels/julia-forge/packages/julia. If we use this, we directly have the right julia in our pixi environment. The Julia dependency is explicitly pinned to the `julia-forge` channel, so future lockfile updates cannot select Julia from another configured channel.

So I did essentially this:

```
pixi rm juliaup
pixi workspace channel add https://prefix.dev/julia-forge
pixi add "https://prefix.dev/julia-forge::julia==1.12.6"
```

Then I could remove `pixi run install-julia` from the CI jobs, since it is no longer needed. At first I thought I'd remove `install-julia` completely, but kept it for now. Because if developers are used to running `julia` instead of `pixi run julia`, or activating their pixi environment, then they may get the wrong version. We can always remove it later on as well.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,8 +33,6 @@ jobs:
       - uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
         with:
           pixi-version: "latest"
-      - name: Prepare pixi
-        run: pixi run install-julia
       - name: Test Wflow
         run: |
           pixi run test-wflow-cov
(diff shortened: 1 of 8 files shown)


SouthEndMusic (approved):
LGTM!

PR #1030, opened 2026-07-30
build: Use recursion instead of explicit stack in license collection

visr:
## Summary

Follow-up to #1025. Replaces the explicit uuid stack/worklist in `collect_dependency_licenses` with plain recursion, per review feedback that recursion reads more naturally for this dependency-graph walk.

- `collect_dependency_license` recurses into each dependency's own deps before checking/copying its license
- `visited_uuids` still guards against revisiting shared transitive dependencies (the manifest dependency graph is not a tree)
- behavior is unchanged: only dependencies reachable from the `Wflow` subproject are scanned; stdlib packages (no tree hash) are skipped for license lookup but still traversed

## Validation

- ran the license collector against the `Wflow` project on top of latest `main`; collected 86 licenses
- asserted `NCDatasets` is included and workspace-only `JSON`/`PackageCompiler` are excluded
- pre-commit formatting and hygiene hooks passed

### DIFF
--- a/build/create_binaries/add_metadata.jl
+++ b/build/create_binaries/add_metadata.jl
@@ -9,8 +9,17 @@ Add the following metadata files to the newly created build:
 - dep_licenses/
 """
 
-function collect_dependency_license(ctx, license_dir, visited_uuids, package_uuid)
-    package_uuid in visited_uuids && return
+"""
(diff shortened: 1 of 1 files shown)


rleander73 (approved):
Looks ok

PR #1047, opened 2026-08-16
Update julia to 1.12.7

visr:
Now that Julia comes from julia-forge, I did `pixi upgrade julia`. After that I did `Pkg.resolve()`, and replaced the final instances of 1.12.6.

https://discourse.julialang.org/t/julia-v1-12-7-has-been-released/138832


SouthEndMusic (approved):
LGTM

PR #1053, opened 2026-08-24
Host versioned example models in public MinIO storage

### LINKED ISSUE #1004: Host example models for the docs
In https://github.com/Deltares/Wflow.jl/issues/723 test data moved to DVC in a MinIO bucket.

The docs currently rely on the test data being accessible to users, since they can download it directly in https://github.com/Deltares/Wflow.jl/blob/a8955dc25326e1d09ff11cb921501e1b1ee0a510/docs/getting_started/download_example_models.qmd.

We need to support this docs use case. I don't think we can expect users to use DVC to get example models.

Options:
- Add them to the release assets like Ribasim's `generated_testmodels.zip`.
- Make a separate public folder in the bucket with one directory per tag with the test data.

Either needs to become part of the release process, so it connects to https://github.com/Deltares/Wflow.jl/issues/826.

visr:
Fixes https://github.com/Deltares/Wflow.jl/issues/1004.

https://s3-console.deltares.nl/browser/wflow/examples%2F

Makes a separate public folder (examples) in the bucket with one ZIP per test model. For the `release/v1.y` branch we can push to other folders. See the added dev docs. CI ensures consistency between the bucket and DVC.


docs/developments/guide.qmd
```diff
@@ -21,6 +21,52 @@ below.
...
+Their input data under `Wflow/test/data/input` is versioned with [DVC](https://dvc.org/): Git tracks
+the directory through `Wflow/test/data/input.dvc`, while the files themselves are stored in the
+public `wflow/testdata` folder on the Deltares MinIO server. Download the input data matching the
+current Git revision with:
+
+```sh
+pixi run download-test-data
+```
+
+Maintainers can update the data using the `wflow-testdata-rw` DVC remote. Configure MinIO
```
vers-w: Might be useful to describe this (`wflow-testdata-rw`) in bit more detail (for persons that are not very familiar with DVC)? It was not clear to me that this points to a configuration section of DVC. 

Did notice here is also `wflow-testdata-ro`, pointing to the same remote storage, what is the difference? Maybe useful to add a comment(s) to dvc.config about this if possible.

docs/developments/guide.qmd
```diff
@@ -21,6 +21,52 @@ below.
...
+credentials locally or through the standard AWS environment variables; never commit credentials.
+After changing files under `Wflow/test/data/input`, update and publish the DVC state with:
+
+```sh
+pixi run dvc add Wflow/test/data/input
+pixi run dvc push --remote wflow-testdata-rw
```
vers-w: Would it be useful to define 1 pixi task for this?

docs/developments/guide.qmd
```diff
@@ -21,6 +21,52 @@ below.
...
+published under `wflow/examples/dev`; release branches use their corresponding
+`wflow/examples/vX.Y` folder. After updating test data, example configurations, the selection, or
+the publishing utility, publish the archives and commit the updated lockfile:
+
+```sh
+pixi run dvc repro publish-example-models
+git add Wflow/test/data/input.dvc dvc.lock
+```
+
+When creating the `release/v1.1` branch, replace `examples/dev` with `examples/v1.1` in the stage
```
vers-w: Maybe better to describe this more generally (not for `v1.1` specific, but for major and minor releases)?

docs/developments/guide.qmd
```diff
@@ -21,6 +21,52 @@ below.
...
+`utils/example_models.toml` defines which Git-tracked configuration and which DVC-tracked input
+files together constitute each example model used by the documentation. A DVC pipeline packages
+these selections into three public archives as non-cached, persistent outputs. On `main` these are
+published under `wflow/examples/dev`; release branches use their corresponding
+`wflow/examples/vX.Y` folder. After updating test data, example configurations, the selection, or
+the publishing utility, publish the archives and commit the updated lockfile:
+
+```sh
+pixi run dvc repro publish-example-models
+git add Wflow/test/data/input.dvc dvc.lock
```
vers-w, line 54: One pixi task?

docs/developments/guide.qmd
```diff
@@ -21,6 +21,52 @@ below.
...
+```
+
+When creating the `release/v1.1` branch, replace `examples/dev` with `examples/v1.1` in the stage
+command and all outputs in `dvc.yaml`. Make the same replacement in the download links in
+`docs/getting_started/download_example_models.qmd`, then publish and commit the release-specific
+pipeline state:
+
+```sh
+pixi run dvc repro publish-example-models
+git add dvc.yaml dvc.lock docs/getting_started/download_example_models.qmd
```
vers-w, line 64: One pixi task?

vers-w (changes requested):
LGTM!
 
Some suggestions, mainly doc improvements and to consider to make use of pixi tasks (pixi.toml) to update the test data etc. if useful/possible.

docs/developments/guide.qmd
visr: Yeah that was still missing. I added comments in 658d425721a4789fbb3c67c8c3c981b98d99f069 and renamed the remotes to `read-testdata` and `write-testdata`.

docs/developments/guide.qmd
visr: I considered it, but since this is (currently) an infrequent task, and you probably want to verify the results and logging of `dvc add` before proceeding to `dvc push`, I lean towards not abstracting it away in a Pixi task.

docs/developments/guide.qmd
visr: Done

docs/developments/guide.qmd
visr: I'd rather not have Pixi doing `git add` for me, I prefer to do that manually when verifying the changes in VSCode.

docs/developments/guide.qmd
visr: See above. Also the rename in `download_example_models.qmd` is not automated.

Ideally we'd have some separate dev docs on how to create and maintain release branches, then this can move there, because it is easy to forget.

For now I don't envision this going onto the 1.0 branch.

vers-w (approved):

PR #1057, opened 2026-08-25
Migrate CompatHelper to Dependabot

visr:
Like https://github.com/Deltares/Ribasim/pull/3217


.github/dependabot.yml
```diff
@@ -4,8 +4,16 @@ updates:
...
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
     groups:
       all-actions:
         patterns:
           - "*"
+  - package-ecosystem: "julia"
+    directory: "/"
```
HendrikKok, line 13: CompatHelper previously ran with subdirs=["Wflow"]. Shouldn't the Dependabot configuration use the same?

visr: I think we want the compat entries everywhere in the workspace. Let's give this a try, if we don't like it, we could perhaps limit it like that.

PR #1066, opened 2026-09-01
Switch from JuliaFormatter to Runic

visr:
https://github.com/fredrikekre/Runic.jl is newer, well-made, and I like its formatting better. Ribasim has been using it for a while. Now Julia-VSCode also supports it (currently requires the pre-release), so it is a good moment to consider switching.

I made some changes to both Wflow and Ribasim to handle formatting in a comparable manner:

- https://github.com/Deltares/Ribasim/pull/3238
- https://github.com/Deltares/Ribasim/pull/3241

For reviewing it is good to separate the implementation in bd0651333b5393fd46f6a140a201c85bb4f77b99 from the formatting in 05cd638eb4f9538303b99157db8cf18127af2c14.
I am also adding the `pixi run check` task to run prek on all files in https://github.com/Deltares/Wflow.jl/pull/1066/changes/14f4e91fd5d47bce7aa69aab2ffcfb9b9a82273a.
The `fix-check` task is run for instance when updating the manifest. If that updates our formatter we want to directly switch to the new formatting, not fail that it now wrongly formatted.


vers-w (approved):
LGTM! 

I focused mostly on the formatting and did fix explicit returns for mutating functions and changed the explicit returns that were in front of try and if blocks (noticed there is also an issue about this https://github.com/fredrikekre/Runic.jl/issues/52).

I guess this works now better with VSCode than JuliaFormatter (version)? In the end I stopped using it in VSCode.

visr:
> I guess this works now better with VSCode than JuliaFormatter (version)? In the end I stopped using it in VSCode.

Both should now work equally well in the current VSCode Julia pre-release. Part of the pain with JuliaFormatter I think was that our env was on v2 and the plugin on v1, causing edit wars. To some degree this can still happen as the version of Runic can differ between our env and the version bundled with the plugin.

PR #1090, opened 2026-09-16
Fix BMI `get_var_units` and `get_var_location` for layered soil variables

### LINKED ISSUE #1064: Wflow server: BMI `get_var_units` not working for layered variables
Issue encountered with OpenDA:
```
julia> request((fn = "get_var_units", name = "soil_layer_1_water_unsaturated_zone__depth"))
[ Info: Received request to run function `get_var_units`...
┌ Error: Wflow function `get_var_units` failed
│   exception =
│    FieldError: type Nothing has no field `unit`; Nothing has no fields at all.
```

Probably some other BMI functions have the same issue.

vers-w:
## Issue addressed
Fixes #1064


Wflow/src/bmi.jl
```diff
@@ -425,3 +424,11 @@ function grid_element_type(model::Model, lens::ComposedFunction)
     end
     return element_type
 end
+
+function get_standard_name(name::String)
```
visr:
```suggestion
function get_standard_name(name::String)::String
```

For non-generic functions I think this makes it easier to read.

visr (approved):

